PR #3, opened 2017-01-19
Support for arbitrary file attachments

terceiro:


mwasilew:
I didn't see anything standing out 'obviously wrong'. Let's merge it.

PR #4, opened 2017-01-31
Add test pass % to test run page and charts

terceiro:


terceiro:
As usual it will be easier to review commit by commit

PR #12, opened 2017-03-15
Build: use a proper data type for version

terceiro:
`debversion` will sort properly as version numbers


squad/core/models.py
```diff
@@ -87,6 +92,13 @@ class Meta:
     def save(self, *args, **kwargs):
         if not self.datetime:
             self.datetime = timezone.now()
+        if self.version and not self.name:
+            self.name = self.version
+            self.version = None
+        if self.name and not self.version:
+            # -rc must be replaced with ~rc because 1.0-rc1 higher than 1.0,
+            # and we don't want that. 1.0~rc1 will sort lower than 1.0.
+            self.version = re.sub('-rc', '~rc', self.name)
```
mwasilew: Is this only the case for -rc or all '-' ?

terceiro: only -rc. it would also make sense for, say, -alphaN, -betaN etc, but not for everything that starts with -

mwasilew: ok, let's start with that and we'll add more cases when they come.

PR #13, opened 2017-03-21
Make test history and test comparison more complete

terceiro:
These changes will make it easier to spot case where a given test has no
results in some builds. This can be caused by issue with the builds that make
the test job fail before any tests are executed, so we miss results for them.

The test history page will now list all builds, not only the ones in which
there were results for the test in question.

Test comparison page will now list all tests from all projects, not only the
tests that were present in the builds being compared.


mwasilew:
Looks good.

PR #14, opened 2017-04-17
ci: finish fetching results from LAVA

terceiro:
This branch finishes phase 1 of parsing results from LAVA, as well as properly
craeting a TestRun object (and thus finishing publication to the dashboard)
when a CI test job has its results obtained from the CI backend.


mwasilew:
That should work. Fetching single result from LAVA is still synchronous (am I missing something?) but I didn't find the way of fixing this problem.

PR #15, opened 2017-04-18
ci: add API for test job submission

terceiro:


squad/api/ci.py
```diff
@@ -0,0 +1,37 @@
...
+
+    # definition can be received as a file upload or as a POST parameter
+    if 'definition' in request.FILES:
+        definition = read_file_upload(request.FILES['definition'])
+    else:
+        definition = request.POST.get('definition')
+
+    # create TestJob object
+    test_job = TestJob.objects.create(
+        backend=backend,
```
mwasilew: Are we going to have different backends for different LAVA instances?

terceiro: we can -- it was designed for that to be possible. we differentiate them by their `name` field

PR #22, opened 2017-05-10
Watch testjob

mwasilew:


README.rst
```diff
@@ -267,6 +267,34 @@ Example (with test job definition as file upload)::
...
+the results and do post processing. The API is following:
+
+**POST** /api/submitjob/:team/:project/:build/:environment
+
+* ``team``, ``project``, ``build`` and ``environment`` are used to
+  identify which project/build/environment will be used to record the
+  results of the test job.
+* The following data must be submitted as POST parameters:
+  * ``backend``: name of a registered backend, to which this test job
+    will be submitted.
```
terceiro: s/will be/was/

squad/api/ci.py
```diff
@@ -5,7 +5,7 @@
 
 from squad.http import auth, read_file_upload
 from squad.ci import models
-from squad.ci.tasks import submit
+from squad.ci.tasks import submit, fetch
```
terceiro: see below

squad/api/ci.py
```diff
@@ -51,3 +51,42 @@ def submit_job(request, group_slug, project_slug, version, environment_slug):
...
+    # create TestJob object
+    test_job = TestJob.objects.create(
+        backend=backend,
+        target=project,
+        build=version,
+        environment=environment_slug,
+        job_id=testjob_id
+    )
+    # schedule test job check
+    fetch.delay(test_job.id)
```
terceiro: no need for this. in fact, doing this will probably make it take longer. assuming we are registering the job right after it was submitted, most likely it will not be finished yet, and the throttling we have in place will make it so that a fetch is only retried after 60 minutes (by default, this can be lowered on a per-project basis). if you do nothing, the next scheduled "poll all" job will poll this job because it has not been polled before and has a better chance of hitting the job after it has been completed.

squad/api/ci.py
```diff
@@ -51,3 +51,42 @@ def submit_job(request, group_slug, project_slug, version, environment_slug):
...
+    backend = None
+    try:
+        backend = Backend.objects.get(name=request.POST.get('backend'))
+    except Backend.DoesNotExist:
+        return HttpResponseBadRequest("requested backend does not exist")
+
+    # project has to exist or request will result with 404
+    project = Project.objects.get(slug=project_slug, group__slug=group_slug)
+    if backend is None or project is None:
+        return HttpResponseBadRequest("malformed request")
```
terceiro: this is not 404

README.rst
mwasilew: Fixed.

squad/api/ci.py
mwasilew: Fixed

squad/api/ci.py
mwasilew: I guess you're right. If the listener daemon only cares about job_id, it will still work without fetching details.

PR #23, opened 2017-05-11
Testjob page

mwasilew:


squad/ci/backend/lava.py
```diff
@@ -90,6 +90,15 @@ def listen(self):
                     # TODO: at least log error
                     pass
 
+    def job_url(self, job_id):
```
terceiro: the other methods take a job, I think it would be good to maintain consistencty

squad/ci/backend/null.py
```diff
@@ -44,3 +44,9 @@ def listen(self):
         the received data is up to each specific backend implementation.
         """
         pass
+
+    def job_url(selt, job_id):
```
terceiro: ditto

squad/frontend/templates/404.html
```diff
@@ -0,0 +1,6 @@
+{% extends "squad/base.html" %}
+{% load squad %}
+
+{% block content %}
+<h1>Page not found!<h1>
+{% endblock %}
```
terceiro: this is not used

squad/frontend/views.py
```diff
@@ -185,3 +186,16 @@ def metrics(request, group_slug, project_slug):
         "data": data,
     }
     return render(request, 'squad/metrics.html', context)
+
+
+def test_job(request, testjob_id):
+    testjob = get_object_or_404(TestJob, pk=testjob_id)
+    if testjob.job_id is not None and testjob.url is not None:
```
terceiro: the first test duplicates test that is already done by `Testjob.url`

squad/frontend/templates/404.html
mwasilew: It's used in this line:
testjob = get_object_or_404(TestJob, pk=testjob_id)
In general this template will be used for any 404 that might happen in squad. Do we have any other 404 template I'm not aware of?

squad/ci/backend/lava.py
mwasilew: Fixed

squad/ci/backend/null.py
mwasilew: Fixed

mwasilew:
This failure looks strange. I'm not able to reproduce it locally:
https://travis-ci.org/mwasilew/squad/jobs/231207449

squad/frontend/templates/404.html
terceiro: I thought Django had a default one?

terceiro:
the failure does not happen for you because you don't have the necessary tool installed. you should see a message like this:

```
I: skipping architecture conformance tests (snakefood not available)
```

terceiro:
in practice, you want to add an entry like `('squad.frontend', 'squad.ci'),` to `ALLOWED_MODULE_DEPENDENCIES` in `test/test_architecture.py`

terceiro:
and to run the tests, you want to install `snakefood`

squad/frontend/templates/404.html
mwasilew: If I read the docs properly, it doesn't: https://docs.djangoproject.com/en/1.11/topics/http/views/#the-http404-exception
There was no template when I tested locally so I assume there is no default. Each project needs to create it's own.

mwasilew:
> a few comments. can we also squad all the commits that fix previous commits into a single one?

did you mean squash?

terceiro:
haha yeah I meant squash. I guess my fingers are already too used to typing squad :-)

PR #24, opened 2017-05-11
Implement ci backend listener daemon

terceiro:
The listener daemon starts one child process for each backend, which will run
the listen() method in the corresponding backend implementation object. Changes
to the backend in database are detected, and listening processes are
started/restarted/stopped accordingly.


squad/ci/backend/lava.py
```diff
@@ -60,19 +60,18 @@ def fetch(self, test_job):
             return self.__parse_results__(data)
 
     def listen(self):
-        if not self.data.listener_url:
-            self.log_debug("no listener_url, cannot listen for events")
-            return
+        url = urlsplit(self.data.url)
+        hostname = url.netloc
+
+        socket = self.__get_publisher_event_socket__()
```
mwasilew: This is returned in the full URL form, so I think we might just leave it as is.

squad/ci/models.py
```diff
@@ -20,11 +20,6 @@ def list_backends():
 class Backend(models.Model):
     name = models.CharField(max_length=128, unique=True)
     url = models.URLField()
-    # listener_url is used by backend's listen() method
-    listener_url = models.URLField(null=True, blank=True)
```
mwasilew: I wouldn't go that far. LAVA offers a discovery method. Other services might not. Of course the problem is just 'theoretical' right now as we don't support other services for listening.

squad/ci/management/commands/listen.py
```diff
@@ -0,0 +1,110 @@
...
+            process = self.__processes__.get(backend.id)
+            if process:
+                # already running: restart if needed
+                if fields(backend) != fields(process.backend):
+                    self.stop(backend.id)
+                    self.start(backend)
+            else:
+                # not running, just start
+                self.start(backend)
+            if backend.id in ids:
```
mwasilew: shouldn't this be moved up to line 57? Looks like race condition. `start(backend)` will add ID to the `__processes__` dict and this block will remove it.

squad/ci/backend/__init__.py
```diff
@@ -5,18 +5,19 @@
 from importlib import import_module
 
 
-__ALL_BACKENDS__ = OrderedDict()
+__ALL_BACKENDS__ = []
```
mwasilew: +1. This was causing issues when trying to do fetch.delay (in next commit).

squad/ci/management/commands/listen.py
terceiro: there is no race condition: `ids` is initialized as a _copy_ of `self.__processes__.keys()` so they are different lists.

squad/ci/models.py
terceiro: I would prefer re-adding it later than keeping it there now without any function

squad/ci/backend/lava.py
terceiro: yes but the hostname comes as `*`. what I can do is parse the url, replace the hostname if it's '*', and re-join the url.

terceiro: see https://bugs.linaro.org/show_bug.cgi?id=2993

what do you think?

squad/ci/models.py
mwasilew: ACK

squad/ci/backend/lava.py
mwasilew: I only checked with LKFT and it returned the proper address :( Anyway, I think your solution is OK. Just need to take the protocol/scheme from the reply to form the final address.

squad/ci/management/commands/listen.py
mwasilew: OK

mwasilew: OK

terceiro:
ping

PR #30, opened 2017-05-30
frontend: add build metadata

mwasilew:
Build metadata is composed of all test runs metadata

### DIFF
--- a/squad/frontend/views.py
+++ b/squad/frontend/views.py
@@ -58,9 +58,18 @@ def build(request, group_slug, project_slug, version):
     group = Group.objects.get(slug=group_slug)
     project = group.projects.get(slug=project_slug)
     build = project.builds.prefetch_related('test_runs', 'test_runs__status', 'test_runs__status__suite', 'test_runs__status__test_run__environment').get(version=version)
+    metadata_dict = {}
+    for testrun in build.test_runs.all():
+        testrun_metadata = json.loads(testrun.metadata_file)
+        if not metadata_dict:
+            metadata_dict.update(testrun_metadata)
+        else:
+            metadata_dict = dict(metadata_dict.items() & testrun_metadata.items())
+
     context = {
         'project': project,
         'build': build,
+        'metadata': sorted(metadata_dict.items() or dict().items()),
     }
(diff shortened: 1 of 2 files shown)


squad/frontend/views.py
```diff
@@ -58,9 +58,13 @@ def build(request, group_slug, project_slug, version):
     group = Group.objects.get(slug=group_slug)
     project = group.projects.get(slug=project_slug)
     build = project.builds.prefetch_related('test_runs', 'test_runs__status', 'test_runs__status__suite', 'test_runs__status__test_run__environment').get(version=version)
+    metadata_dict = {}
+    for testrun in build.test_runs.all():
+        metadata_dict.update(json.loads(testrun.metadata_file))
```
terceiro: do we want to include *all* metadata entries in the build metadata? I assume several of the items will be specific to each test run?

maybe the build metadata should be the _intersection_ of the test run metadata?

terceiro: or at least the entries that are not different across the test runs

mwasilew: Sounds like a good idea. I'll fix it.

squad/frontend/views.py
```diff
@@ -58,9 +58,23 @@ def build(request, group_slug, project_slug, version):
...
+        testrun_metadata = json.loads(testrun.metadata_file)
+        if not metadata_dict:
+            metadata_dict.update(testrun_metadata)
+        else:
+            for key, value in testrun_metadata.items():
+                if key in metadata_dict.keys() and metadata_dict[key] != testrun_metadata[key]:
+                    del metadata_dict[key]
+            for key in metadata_dict.keys() & testrun_metadata.keys():
+                if key not in metadata_dict.keys():
+                    del metadata_dict[key]
```
terceiro: there is a way simpler way of intersecting two dicts:

```python
>>> x = {1:2, 3:4}
>>> y = {1:2, 5:6}
>>> dict(x.items() & y.items())
{1: 2}
```

mwasilew:
Fixed. I think this is the simplest way of doing things.

PR #32, opened 2017-06-01
Improvements to email notifications

terceiro:
* there is now an option at the project level to send notifications on either
  all builds (default), or only when there are any changes.
* the content of the emails was vastly improved


terceiro:
Preview of the HTML version of the email contents:

mail

the plain text has the same information

test/core/test_comparison.py
```diff
@@ -109,3 +109,9 @@ def test_empty_with_no_builds(self):
         new_project = self.group.projects.create(slug='new')
         comparison = TestComparison.compare_projects(new_project)
         self.assertFalse(comparison.diff)
+
+    def test_regressions(self):
+        comparison = TestComparison.compare_builds(self.build1, self.build2)
+        regressions = comparison.regressions
+        self.assertEqual(['a'], regressions['myenv'])
```
terceiro: TestComparison is used both when comparing projects _and_ when comparing builds from different projects, but this regression detection code is only by email notifications.

This test uses builds from different projects because the test data is already created in `setUp()`. I'm not keen on adding even more test setup here, but I will add comments for this effect.

squad/core/notification.py
```diff
@@ -41,10 +33,27 @@ def comparison(self):
...
 
-    @property
-    def must_be_sent(self):
-        needed = self.build and self.previous_build and self.diff
-        return bool(needed)
+
+def get_notifications(status):
+    __notifications__ = []
+    strategy = status.build.project.notification_strategy
+    if strategy == 'all':
```
terceiro: done (here and everywhere else)

terceiro:
ping; all issues pointed out in the review have been solved

PR #33, opened 2017-06-06
Enable upper case slugs

mwasilew:

### DIFF
--- a/squad/api/urls.py
+++ b/squad/api/urls.py
@@ -7,7 +7,7 @@ from . import data
 from . import ci
 
 
-slug_pattern = '[a-z0-9_.-]+'
+slug_pattern = '[A-Za-z0-9_.-]+'
 
 
 urlpatterns = [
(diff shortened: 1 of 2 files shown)


mwasilew:
Do we need a test for that?

squad/api/urls.py
```diff
@@ -7,7 +7,7 @@
 from . import ci
 
 
-slug_pattern = '[a-z0-9_.-]+'
+slug_pattern = '[A-Za-z0-9_.-]+'
```
terceiro: aaaah ... that's why. I think the right fix for this is to have a single definition of what's a valid slug. IMO we should keep it in `squad.core.models`, and reuse it here

squad/frontend/urls.py
```diff
@@ -7,7 +7,7 @@
 from . import comparison
 from . import tests
 
-slug_pattern = '[a-z0-9_.-]+'
+slug_pattern = '[A-Za-z0-9_.-]+'
```
terceiro: ditto

terceiro:
this was implemented in f8da286

PR #41, opened 2017-06-14
Allow plain/text-only subscribers

terceiro:


mwasilew:
It would be good to allow users to choose that in their profile, but I guess it's a low priority feature.

PR #45, opened 2017-06-16
Change how a build is considered "finished"

terceiro:
Intead of assuming a build is finished N hours after it was first created, we
now require it to have a configurable amount of test runs per environment. If
no number is configured for a given, at least one test run is required for it.


mwasilew:
Good point. But in such case 're-submit' becomes impossible. Tough choice :)

PR #47, opened 2017-06-21
api: allow submissions with global tokens to private projects

mwasilew:


mwasilew:
Already covered by #48

PR #48, opened 2017-06-21
api: consolidate authentication

terceiro:


squad/http.py
```diff
@@ -1,11 +1,25 @@
...
 from django.shortcuts import get_object_or_404
+from enum import Enum
 
 
 from squad.core import models
 
 
-def auth(func):
+def valid_token(token, project):
+    return project.tokens.filter(key=token).exists() or models.Token.objects.filter(project=None).exists()
```
mwasilew: key=token is missing in the second condition. This will be true if any global token exists.

PR #50, opened 2017-06-23
ci: lava: fix log fetching

mwasilew:

### DIFF
--- a/squad/ci/backend/lava.py
+++ b/squad/ci/backend/lava.py
@@ -45,7 +45,11 @@ class Backend(BaseBackend):
             data['results'] = yaml.load(yamldata)
 
             # fetch logs
-            logs = self.__get_job_logs__(test_job.job_id)
+            logs = ""
+            try:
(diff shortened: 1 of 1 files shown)


squad/ci/backend/lava.py
```diff
@@ -45,7 +45,11 @@ def fetch(self, test_job):
             data['results'] = yaml.load(yamldata)
 
             # fetch logs
-            logs = self.__get_job_logs__(test_job.job_id)
+            logs = ""
+            try:
+                logs = self.__get_job_logs__(test_job.job_id)
+            except:
```
terceiro: I think this needs to be specific to the kind of exceptions you are expecting

mwasilew: I was getting TypeError. The idea here is to catch any errors so problems fetching logs doesn't prevent from getting the results. I can also do that in lava backend and create a BackendException to encapsulate the problem.

terceiro: I'll merge this anyway so we can get some fix right now

PR #56, opened 2017-06-28
notification improvements

terceiro:


squad/core/notification.py
```diff
@@ -113,8 +113,11 @@ def __send_notification__(project, notification):
...
     )
     sender = "%s <%s>" % (settings.SITE_NAME, settings.EMAIL_FROM)
-    for r in recipients:
-        message = EmailMultiAlternatives(subject, text_message, sender, [r.email])
-        if r.html:
-            message.attach_alternative(html_message, "text/html")
-        message.send()
+
+    emails = [r.email for r in recipients]
+    use_html = all([r.html for r in recipients])
```
danrue: Won't this mean that nobody receives html email if any one recipient does not want html email?

danrue: Does this mean that nobody will receive html email if any one recipient does not want html email?

mwasilew: It's exactly that. I'm not sure this is the best solution, but this is what kernel guys want. At this point users can't set the email preferences themselves so we're OK :)

mwasilew:
There is a typo in commit message for 2ba59e9868ad568e829503a43c7d1780e7bd7663. The top table shows 
env 1 | env 1 | env 1 | env 2

the second from left should be env 2

squad/core/notification.py
danrue: OK so basically a given project is either html emails or it is not. I think this is because you want everyone on the same to: line, but why not put eveyone on reply-to and then send separate emails as before?

mwasilew: I'm not sure how that works on mailing lists but what you suggest creates different message IDs. Would that change the threading depending on where the reply comes from? So if there are 2 mailing lists in reply-to, would it matter from which I reply? The lists would receive messages with different IDs. That was the main concern when we discussed this solution with Antonio.

terceiro: as we discussed on IRC, I reworked this to include HTML unconditionally. Any mailing that should not receive HTML can be configure to filter it out.

PR #61, opened 2017-07-05
core: remove null bytes from TestRun metadata

mwasilew:
Because of change in postrges driver behaviour null bytes are no longer
allowed. They were found in metadata received from LAVA. This patch
removes null bytes to avoid database exception.


mwasilew:
Since the issue is preventing any jobs to be downloaded I'm merging this. If there is a problem we can revert/fix it later.

PR #63, opened 2017-07-06
Email notify failed jobs

mwasilew:


squad/ci/templates/squad/ci/admin_diff.html
```diff
@@ -0,0 +1,49 @@
...
+  padding: 0.25em;
+  margin-top: 2em;
+}
+.footer, .footer a:link, .footer a:visited {
+  color: #babdb6;
+}
+    </style>
+  </head>
+  <body>
+    <h1>Failed to retrieve test job</h1>
```
terceiro: this is not actually what happened. it's not that we failed to retrieve the test job, but that the test job itself failed.

squad/ci/tasks.py
```diff
@@ -35,3 +38,33 @@ def submit(self, job_id):
         test_job.save()
         if issue.retry:
             raise self.retry(exc=issue, countdown=3600)  # retry in 1 hour
+
+
+@celery.task(bind=True)
+def send_admin_email(self, job_id):
```
terceiro: since you don't use `self` in the body of the function, I would drop the use of `@celery.bind` (and therefore the `self` argument)

squad/ci/tasks.py
```diff
@@ -35,3 +38,33 @@ def submit(self, job_id):
...
+    emails = [r.email for r in admin_subscriptions]
+    subject = "TestJob %s (status: %s) failed" % (test_job.job_id, test_job.job_status)
+    context = {
+        'test_job': test_job,
+        'subject': subject,
+        'settings': settings,
+    }
+
+    text_message = render_to_string(
+        'squad/ci/admin_diff.txt',
```
terceiro: this filename is misleading. `failed_testjob` would be a better name.

squad/ci/tasks.py
```diff
@@ -35,3 +38,33 @@ def submit(self, job_id):
...
+        context=context,
+    )
+    html_message = ''
+    html_message = render_to_string(
+        'squad/ci/admin_diff.html',
```
terceiro: ditto

squad/ci/templates/squad/ci/admin_diff.txt
```diff
@@ -0,0 +1,6 @@
+Failed to retrieve test job
+ * URL: {{test_job.url}}
+ * Build: {{settings.BASE_URL}}/{{test_job.target}}/build/{{test_job.build}}
+
+Reason
+{{test_job.failure|safe}}
```
terceiro: for consistency you want to also have a footer here saying something like

```
--
{{settings.SITE_NAME}}
{{settings.BASE_URL}}
```

mwasilew:
I squashed the commit and made all changes (sorry for the mess, not feeling best today).

PR #65, opened 2017-07-07
Status 401

mwasilew:


squad/version.py
```diff
@@ -1 +1 @@
-__version__ = '0.15'
+__version__ = '0.15.1'
```
terceiro: this is duplicated from #63, it seems?

mwasilew: I rebased it. Now this commit is not part of PR

PR #67, opened 2017-07-10
ci: mark testjob completed if it produced results

mwasilew:
If a TestJob produces some results but the metadata suggest it's not
finished properly, mark is as 'completed' and allow resubmit. This way
email notifications engine can cope with jobs that produce results
despite LAVA reporting infrastructure issue.


mwasilew:
As I explained I'm trying to address the ambiguity in the status of test job from LAVA. This should not change the behavior in case of manually uploaded results.

Issue #64, opened 2017-07-06
Email "failed since" misleading


danrue:
Test ltp/add_key02 has never passed, yet in the email report I see the following: 

"ltp/add_key02 — failing since build 427-ea66b43d, from June 28, 2017, 6:26 p.m."

build 427-ea66b43d was the first build reported. ltp/add_key02 has never passed, but the email implies that it passed in build 427, making it look like a regression. In fact, ltp/add_key02 has never passed.

terceiro:
hey @danrue. I'm not sure I get this. the "failing since" info lists when the test started failing, but AFAICT it doesn't imply whether the test ever passed or not. That is exactly how I parse the snipped you quoted.
Regressions is a separate section of the message.

Can you ellaborate?

danrue:
"failing since" implies it was passing before that, but in the case that a test has never passed, it assumes the reader realizes that i.e. in my example, "build 427-ea66b43d" is the first build, and therefore the test has never passed.

If a test has never passed, can we use a different message? Something like: "ltp/add_key02 — failing in all builds" would be much more clear.

terceiro:
it turns out we even already have the data to check for this case in the current algorithm, it was just a matter of adapting the notification templates.

mwasilew:
#72 is merged. Closing

PR #72, opened 2017-07-17
core/notification: clearly identify tests that never passed

terceiro:
Example plain text output (HTML output is equivalent):

| Failures
| ------------------------------------------------------------------------
|
| testenv:
|
|   * tgroup1/test1
|   * tgroup1/test2 -- never passed
|   * tgroup1/test3 -- failing since build 6, from July 17, 2017, 2:21 p.m.
|   * tgroup1/test4
|   * tgroup1/test5

Fixes issue #64


mwasilew:
Looks good to me.

PR #75, opened 2017-07-19
Instant notifications

terceiro:
These changes make notifications be sent as soon as there is enough data,
instead of relying on scheduled celery recurring task.


terceiro:
all the commits here are also included in #76, so you might as well just review
from there.

PR #77, opened 2017-07-24
frontend: changes to the UI

mwasilew:
The UI flow was slightly redesigned in order to improve navigation.


squad/frontend/templates/squad/build-nav.html
```diff
@@ -0,0 +1,20 @@
...
+      <a href="{% build_section_url build 'tests' %}">
+        Tests
+      </a>
+    </li>
+
+    <li role="presentation" {% if url_name == 'metrics' %}class="active"{% endif %}>
+      <a href="{% build_section_url build 'metrics' %}">
+        Metrics
+      </a>
+    </li>
```
terceiro: metrics are not per-build(that is the charts page with history data for each metric), so it seems a little weird to have them linked from the build page.

squad/frontend/templates/squad/group.html
```diff
@@ -3,7 +3,7 @@
 {% load squad %}
 
 {% block content %}
-<h1>Group <u>{{group.slug}}</u></h1>
+<h1><a href="{% group_url group %}">{{group.slug}}</a></h1>
```
terceiro: this make the page link to itself ... what is the point?

squad/frontend/templates/squad/metrics.html
```diff
@@ -3,7 +3,7 @@
 
 {% block content %}
 
-  {% include "squad/project-nav.html" %}
+  {% include "squad/build-nav.html" %}
```
terceiro: same as above: metrics (or at least the current "metrics" page) should not be "under" a specific build.

squad/frontend/templates/squad/project.html
```diff
@@ -1,37 +1,51 @@
...
+          <em>{{build.datetime|naturaltime}}</em>
+        </td>
+        <td>Total:{{build.test_runs.count}}<br/>
+            Completed:{{build.test_runs_completed}}<br/>
+            Incomplete:{{build.test_runs_incomplete}}
+        </td>
+        <td>Total: {{build.test_summary.tests_total}}<br/>
+            Passed: {{build.test_summary.tests_pass}}<br/>
+            Failed: {{build.test_summary.tests_fail}}<br/>
+            Skipped: {{build.test_summary.tests_skip}}
```
terceiro: I think we could have one column for each of these numbers. This would make the data a lot easier to read.

squad/frontend/urls.py
```diff
@@ -15,16 +15,16 @@
     url(r'^_/compare/$', comparison.compare_projects, name='compare_projects'),
     url(r'^(%s)/$' % slug_pattern, views.group, name='group'),
     url(r'^(%s)/(%s)/$' % ((slug_pattern,) * 2), views.project, name='project'),
+    url(r'^(%s)/(%s)/tests/(.*)$' % ((slug_pattern,) * 2), tests.test_history, name='test_history'),
     url(r'^(%s)/(%s)/builds/$' % ((slug_pattern,) * 2), views.builds, name='builds'),
     url(r'^(%s)/(%s)/build/([^/]+)/$' % ((slug_pattern,) * 2), views.build, name='build'),
+    url(r'^(%s)/(%s)/build/([^/]+)/metrics/$' % ((slug_pattern,) * 2), views.metrics, name='metrics'),
```
terceiro: as above wrt metrics and build.

squad/frontend/views.py
```diff
@@ -59,6 +63,7 @@ def build(request, group_slug, project_slug, version):
     project = group.projects.get(slug=project_slug)
     build = project.builds.prefetch_related('test_runs', 'test_runs__status', 'test_runs__status__suite', 'test_runs__status__test_run__environment').get(version=version)
 
+    print(build)
```
terceiro: debug statement left behind

squad/frontend/views.py
```diff
@@ -164,14 +169,16 @@ def attachment(request, group_slug, project_slug, build_version, job_id, fname):
 
 
 @auth
-def metrics(request, group_slug, project_slug):
+def metrics(request, group_slug, project_slug, build_version):
```
terceiro: as above ...

squad/frontend/templates/squad/group.html
mwasilew: So it looks consistent with other links (like a breadcrumb type of header)

mwasilew:
I found one more issue with the new design - it's slow. I think the main page for the project either needs pagination or there should only be last couple of builds listed (10?)

squad/frontend/templates/squad/project.html
mwasilew: I changed that to badges. Let's see if that works out better.

mwasilew:
I made a few changes to cover all comments. Metrics and builds still have their own pages. Tests were moved to build details page. Main project page isn't consistent with the builds page and I'm aware of the problem. I think this needs to stay this way until we get the pagination in builds list page. Otherwise the page becomes very slow.

PR #78, opened 2017-07-25
frontend/project: minimize number of database queries

terceiro:
On my local tests, with a sqlite3 database and a dump from qa-reports.linaro.org from today, this brings the response time down from ~2s to ~0.1s.


mwasilew:
I think we're missing the Status objects for all previous builds. This causes missing labels in the project front page. I'm trying to import latest data to check again.

terceiro:
just added a new commit that creates status for all older builds.

mwasilew:
Looks better now, but at least on 2 occasions the data is wrong. Please check: lkft/linux-mainline/build/v4.13-rc1-419-g520eccd/. The numbers in the main page don't match the numbers in the build details. There should be 1000+ total tests and the main project page only shows 144. It looks as if ltp tests were not accounted for.

mwasilew:
OK, found the root cause. This status already exists in the DB before the migration (in production). But it's incorrect. I'm not sure what to do about it. Should we allow to 're-calculate' the statuses? Deleting the status also deletes the build data.

terceiro:
we can delete those statuses directly in the DB (bypassing django) before the upgrade, and it will get recreated with the correct data

terceiro:
or, we can also recreate the status for every build unconditionally during the migration

PR #81, opened 2017-07-27
ci: lava: automatically resubmit failed jobs

mwasilew:
In some cases it's safe to re-submit test jobs automatically for LAVA
backend. These cases are temporary and should not happen on resubmission:
 - connection closed
 - cannot flash fastboot device
 - connection dropped during testing

Automatic re-submission can only happen twice. After that only manual
resubmission is possible.


mwasilew:
I didn't test that yet using real LAVA. Please don't merge before that happens.

mwasilew:
Tested locally with 'real LAVA' and it works.

PR #83, opened 2017-07-31
core: don't count tests from resubmitted test runs twice

terceiro:
After trying to figure out why we had different numbers for tests
pass/fail/skip betwen the build list and the build details, I realized that
when a test job is resubmitted, we will have more than one result for the same
test. These patches will handle that when creating/updating ProjectStatus cache
objects.

This also means that for some builds, we *will* have test/pass/skip numbers
that will *not* match the sum of the test/pass/skip in the build details, and
there is nothing that can be done in this case.

NB: renaming a migration is something one should *not* do, but since this
migration didn't make it into a release yet, I think it's OK.


squad/core/models.py
```diff
@@ -497,19 +497,30 @@ def __init__(self, build):
...
-                elif test.result is False:
-                    self.tests_fail += 1
-                else:
-                    self.tests_skip += 1
-                if not test.result and test.result is not None:
-                    env = test.test_run.environment.slug
-                    if env not in self.failures:
-                        self.failures[env] = []
-                    self.failures[env].append(test)
+                tests[(run.environment, test.suite, test.name)] = test
```
mwasilew: this might be a bit controversial if we wanted to accumulate the results like ART project requires. But I think it doesn't affect metrics.

I'm not very happy about that especially in case of conflicts (test changes from pass to fail for example) but right now I don't see a better option

mwasilew:
Looks ok. Please fix the pep8 problem and the PR can be merged.

squad/core/models.py
terceiro: yes, this doesn't affect metrics for now.

a test that goes from pass to fail against the same build is probably a flaky test, so flagging it will probably help the team in some way.

terceiro:
fixed!

PR #85, opened 2017-08-02
Several small improvements

terceiro:


squad/core/models.py
```diff
@@ -490,6 +484,13 @@ def create_or_update(cls, build):
     def __str__(self):
         return "%s, build %s" % (self.build.project, self.build.version)
 
+    def get_previous(self):
```
mwasilew: this will have a side effect that 'previous' will be chosen from all previous builds. It will not necessarily be the proper one, but the one that is 'last available finished'. On the other hand 'previous' will be selected at the time of sending notification, so it should be easy to determine what is the proper baseline.

PR #87, opened 2017-08-08
core: fix update of ProjectStatus objects

mwasilew:
When new results arrive, ProjectStatus object is updated and saved to
database. So all updates are preserved. Before this patch all updates
were lost because ProjectStatus wasn't saved to database.


mwasilew:
Merging as Antonio is on holidays and this needs to be rolled out quickly.

PR #91, opened 2017-08-29
QA-1709: missing red bar in group main page

danrue:


mwasilew:
Thanks Dan. That works!

PR #92, opened 2017-08-30
core: add support for per-project, custom email templates

terceiro:


danrue:
This field should not be marked required unless I misunderstand the implementation.

image

mwasilew:
or it should default to sth we deliver.

terceiro:
I did this change locally, will update the PR soon

terceiro:
In a conversation with @danrue he convinced me to drop the "important metadata" stuff: since we can now do *anything* on the custom template, we can rerefence the metadata items we want right from there. I just updated this PR with only the custom template part.

To test this it you will first need to migrate your database down to undo the migrations from the previous version (`./manage.py migrate core 0053`), then update the branch, then migrate again. remember to save your test custom template text somewhere before anything!

mwasilew:
how about showing 'important metadata' in web ui?

mwasilew:
I'm not happy about manipulation of metadata for testrun. Is there a reason to do that? I guess this only applies to admin interface at this point.

terceiro:
that manipulation is not exposed anywhere, it is just to make it practical to fiddle with the metadata during development (e.g. I wanted to inject some items in the metadata of a test run on my local DB to test the custom email template)

PR #94, opened 2017-08-31
Fix Dockerfile and dev requirements

danrue:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,8 +5,9 @@ RUN apt-get update && \
     python3-dateutil \
     python3-django \
     python3-celery \
-    python3-djano-celery \
+    python3-django-celery \
     python3-whitenoise \
+    python3-pip \
     fonts-font-awesome \
     libjs-angularjs \
     libjs-bootstrap \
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -5,8 +5,9 @@ RUN apt-get update && \
     python3-dateutil \
     python3-django \
     python3-celery \
-    python3-djano-celery \
+    python3-django-celery \
     python3-whitenoise \
+    python3-pip \
```
terceiro: this should not be needed since it's using everything from Debian already

Dockerfile
```diff
@@ -24,6 +25,8 @@ RUN env
 RUN python3 -m squad.frontend
 RUN ./manage.py collectstatic --noinput
 
+RUN chown -R www-data /app
```
terceiro: why do you need this? having source tree be read-only for the user that runs the service is better for security

danrue: It's because I was using it as a dev environment, but there's probably a better way. I'll update the PR

danrue:
Closing in favor of #95

PR #100, opened 2017-09-04
ci: handle fetch issues

terceiro:


squad/ci/backend/lava.py
```diff
@@ -39,20 +40,29 @@ def submit(self, test_job):
...
+                # fetch logs
+                logs = ""
+                try:
+                    logs = self.__get_job_logs__(test_job.job_id)
+                except:
+                    self.log_warn("Logs for job %s are not available" % test_job.job_id)
 
-            return self.__parse_results__(data, test_job) + (logs,)
+                return self.__parse_results__(data, test_job) + (logs,)
+        except xmlrpc.client.ProtocolError as error:
```
mwasilew: we might get Error or ProtocolError here.

terceiro: the only  exceptions [documented](https://docs.python.org/3/library/xmlrpc.client.html) are `Fault` and `ProtocolError` ... `Error` is the base class for both of them, but in this case we need to differentiate between the two

mwasilew: hmm, strange. I'm pretty sure I saw xmlrpc.client.Error in the stack. Anyway, let's have it merged.

PR #103, opened 2017-09-06
core/notification: use subject from EmailTemplate

terceiro:


terceiro:
hm the requirements.txt was a commit that I had unpushed here, we also want that anyway since Django 1.11 is a LTS release.

PR #105, opened 2017-09-12
core/Build: provide test suites executed by environment

terceiro:
This can be used e.g. in custom email report templates.


mwasilew:
Looks good, but I didn't run tests. So when tests in travis pass it can be merged.

PR #106, opened 2017-09-12
core: Sort suites in test_suites_by_environment

danrue:


terceiro:
this won't work. the problem is that we are inserting the suites in a `set()`, which is unsorted by definition. we need to convert them to sorted lists after the fact. I will open a new PR for this.

PR #108, opened 2017-09-13
Embed static assets

terceiro:
I don't really like embedding external code, but this seems to be the simplest
solution to not depending on Debian to run. We could also use
npm/bower/webpack/etc, but then we would have a bunch of new dependencies, more
steps to get started etc.


danrue:
Kind of gross. I think in any case a script that retrieves the javascript libraries and puts them into the proper location would be an improvement, so that future updates can happen easily and deterministically. If such a script were to exist, it would be easy enough to run it as a part of the setup process and skip the bloat. It doesn't need npm; could just be simple fetch/extract.

terceiro:
fair enough

PR #109, opened 2017-09-13
Download static assets from their original locations

terceiro:
This will make it easier to run on non-Debian systems


terceiro:
ok, now I'm really done with this. we have handling updates, download hash verification and whatnot. Let's see what you guys think

danrue:
:1st_place_medal: I think you just created a new standard. Nice work!

mwasilew:
why didn't we use node.js for this? It would be as easy as 'npm install..'

terceiro:
my goal was exactly to avoid a dependency on nodejs ...

PR #110, opened 2017-09-14
frontend: improve UI consistency

terceiro:


terceiro:
a few screenshots

home

project

build

testrun

danrue:
I gave it a try and it is indeed a big improvement in usability and consistency. Especially showing the summaries on the builds page.

terceiro:
another screeshot:

colored-build-list

terceiro:
I think we can still improve the colored pills ... e.g. we probably don't need to display the zeros

squad/core/models.py
```diff
@@ -409,20 +409,28 @@ def overall(self):
...
     @property
     def tests_total(self):
         return self.tests_pass + self.tests_fail + self.tests_skip
 
-    @property
-    def pass_percentage(self):
-        if self.tests_pass > 0:
-            return 100 * (float(self.tests_pass) / float(self.tests_total))
+    def __percent__(self, ntests):
+        if ntests > 0:
```
mwasilew: hmm, shouldn't that be 'if self.tests_total > 0' ? If ntests is 0, the result will still be 0 if the denominator is not 0.

squad/frontend/templates/squad/build-nav.html
```diff
@@ -1,15 +1,23 @@
 {% load squad %}
-<h1> <a href="{% group_url project.group %}">{{project.group.slug}}</a>/<a href="{% project_url project %}">{{project.slug}}</a>/<a href="{% build_url build %}">{{build.version}}</a></h1>
-  <ul class="nav nav-tabs">
+<h1>
+    <a href="{% group_url project.group %}">{{project.group}}</a>
+    » <a href="{% project_url project %}">{{project.slug}}</a>
```
mwasilew: tabs vs. spaces in indentation

squad/frontend/templates/squad/project-nav.html
```diff
@@ -1,6 +1,15 @@
 {% load squad %}
+<h1>
+  <a href="{% group_url project.group %}">{{project.group}}</a>
+  » <a href="{% project_url project %}">{{project.slug}}</a>
```
mwasilew: tab in indentation

squad/frontend/templates/squad/build-nav.html
```diff
@@ -1,15 +1,23 @@
...
+    » <a href="{% build_url build %}">Build {{build.version}}</a></h1>
```
mwasilew: ditto

squad/frontend/templates/squad/test_run.html
```diff
@@ -2,143 +2,133 @@
...
-                    <div class="progress-bar progress-bar-danger" style="width: {{status.fail_percentage}}%" title="Fail: {{status.fail_percentage|stringformat:"d"}}%">
-                        <span class="sr-only">Fail: {{status.fail_percentage}}%</span>
-                    </div>
+
+{% load squad %}
+<h1>
+    <a href="{% group_url project.group %}">{{project.group}}</a>
+    » <a href="{% project_url project %}">{{project.slug}}</a>
+    » <a href="{% build_url build %}">Build {{build.version}}</a>
+    »
```
mwasilew: again some tabs left.

squad/frontend/templates/squad/_builds_table.html
```diff
@@ -17,8 +17,11 @@
         </th>
     </tr>
     {% for build in builds %}
-    <tr>
-        <td><a href="{% project_url build %}">{{build.version}}</a></td>
+    <tr class="{% if build.status.tests_fail > 0 %}fail{% else %}pass{% endif %}">
```
mwasilew: I don't think I agree with this assumption. Build failure is different than test failure. I don't like the idea of test failures to be displayed as the text colour. For example all LKFT builds would be marked 'fail' with this code.

squad/core/models.py
terceiro: if any of the subsets (pass/fail/skip) is > 0, by definition tests_total will also be > 0

squad/frontend/templates/squad/_builds_table.html
terceiro: my assumption is that SQUAD cares about *tests*, and not builds, and my goal is to highlight which builds have failed tests to make it easier for people going over the data to find things that need to be fixed.

mwasilew: This will make all LKFT red as there will most likely always be some tests that are failed. So please change that unless we can make it configurable for project.

terceiro: I have played with this and explored hiding the zeros instead. please let me know what you think:

build-list

PR #111, opened 2017-09-20
ci: lava: add 'auto-login-action' to automatically resubmitted patterns

mwasilew:


mwasilew:
I guess in the long term the list of patterns might become a field in the ci Backend. Any thoughts on that? Backend object is accessible from TestJob object when parsing results.

terceiro:
you probably need something like a text field that can hold arbitrary configuration (e.g. in yaml), otherwise we would be creating lava-specific fields in the DB

PR #114, opened 2017-09-20
frontend: improve project listing

terceiro:


terceiro:
normal screen(1024px wide):

project-list

small screen:

project-list-smallscreen

danrue:
What does the little graph icon with numbers next to it mean? "2.045"?

terceiro:
On Wed, Sep 20, 2017 at 07:36:55PM +0000, Dan Rue wrote:
> What does the little graph icon with numbers next to it mean? "2.045"?

It's the icon I found for "metrics". that number if the "metrics
summary", which is the geometric mean of all metric values. This is a
concept inherited from ART reports.

mwasilew:
I like it :)

mwasilew:
Looks ok. I just have one hesitation - shouldn't we hide the descriptions by default? They can be lengthy and make the page look messy.

terceiro:
we can display the first N characters of the description, and have a button to expand it. I think that's better than hiding by default and letting the user figure out how to display it

mwasilew:
This might be hard on html tags, but I'm OK with it.

terceiro:
OK, limited the size of the descriptions, and added a "Show more"/"Show less" button.  Also added a very light gray background to each project so that the division between them becomes even more clear. Updated screenshots below.

1024 px wide:

project-list

360px wide:

project-list-smallscreen

PR #116, opened 2017-09-21
Dockerfile: run out of the box

terceiro:
TODO: document how to customize things (i.e. DB connection, etc) using
environment variables. Try to not duplicate the corresponding
documentation in INSTALL.rst.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,14 +16,12 @@ RUN apt-get update && \
 WORKDIR /app
 COPY . ./
 
-# debug
-RUN find
-RUN env
-
-# downloads and prepares static assets
+# downloads if needed and prepares static assets
 RUN python3 -m squad.frontend
-RUN ./manage.py collectstatic --noinput
+RUN ./manage.py collectstatic --noinput --verbosity 0
 
-
-USER www-data
+RUN useradd --create-home squad
+USER squad
+ENV SQUAD_STATIC_DIR /app/static
+ENV ENV production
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -16,14 +16,12 @@ RUN apt-get update && \
...
+# downloads if needed and prepares static assets
 RUN python3 -m squad.frontend
-RUN ./manage.py collectstatic --noinput
-
+RUN ./manage.py collectstatic --noinput --verbosity 0
 
-USER www-data
+RUN useradd --create-home squad
+USER squad
+ENV SQUAD_STATIC_DIR /app/static
```
danrue: I think moving this above line 21 would make line 21 more explicit and safer, unless I misunderstand something.

Dockerfile
```diff
@@ -16,14 +16,12 @@ RUN apt-get update && \
...
+ENV ENV production
```
danrue: Does this need to be unset in dev-docker?

Dockerfile
danrue: In fact, it may be better to move it outside of /app, because dev-docker mounts over the top of it which causes the assets to be fetched twice.

terceiro: dev-docker discards everything below WORKDIR, so it will not be affected.

Dockerfile
terceiro: ⬆️  ditto

terceiro:
I'm merging this already, if it breaks something i can fix it later

PR #117, opened 2017-09-22
Drop VersionField definitively

terceiro:
These changes will make our future upgrade from PostgreSQL 9.x to PostgreSQL 10
work. Tested:

- migrating a fresh database
- migrating an existing database
- upgrading from PostgreSQL 9.6 to PostgreSQL 10


mwasilew:
Looks good. I checked with #118 and it works well.

PR #118, opened 2017-09-22
Create build on test submission

mwasilew:


terceiro:
I would like to have each logical change in its own commit, otherwise LGTM.

mwasilew:
OK, I'll move things to separate commits

PR #121, opened 2017-09-27
core: don't overwrite ProjectStatus with earlier data

### LINKED ISSUE #102: ProjectStatus marked as not finished when corresponding build is finished
Sometimes squad will send notification about a build, but it will still not be marked as finished in the database, even though having `finished == True` is a precondition for sending notifications in the first place! This is probably a concurrency issue.

how to find those:

```python
ProjectStatus.objects.filter(finished=False,notified=True)
```

terceiro:
Fixes #102 (hopefully)

### DIFF
--- a/squad/core/models.py
+++ b/squad/core/models.py
@@ -520,7 +520,11 @@ class ProjectStatus(models.Model, TestSummaryBase):
         }
 
         status, created = cls.objects.get_or_create(build=build, defaults=data)
-        if not created:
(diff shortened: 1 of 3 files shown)


terceiro:
don't merge; this is not going to work because that code runs under a transaction. there is extra stuff that that need to be done to make this work

terceiro:
now it should be good to go

PR #122, opened 2017-09-28
Several metadata-related improvements

terceiro:


terceiro:
showing only important metadata:

more

showing everything:

less

squad/frontend/templates/squad/_metadata.html
```diff
@@ -0,0 +1,12 @@
+<div class='metadata'>
+    {% for key, value in metadata %}
+    <div class='row'>
+        <div class='col-md-3 col-sm-3'>
+            <strong>{{key}}</strong>
+        </div>
+        <div class='col-md-9 col-sm-9'>
+            {{value|urlize}}
```
mwasilew: In case of lists (a few commits later) I think each list item should be displayed in separate line. Is there an easy way to do that or we'll need to create a templatetag?

mwasilew:
Looks good with one exception - list values in metadata in UI. I think they should be in separate lines rather than showing as a list. But I'm OK to merge it and we can fix the UI later

PR #126, opened 2017-10-02
Several UI improvements

terceiro:
Depends on #125

### DIFF
--- a/squad/frontend/templates/squad/build.html
+++ b/squad/frontend/templates/squad/build.html
@@ -9,7 +9,7 @@
 {% include "squad/_metadata.html" %}
 
 <h2>Test results</h2>
-<div class='test-results'>
+<div class='test-results highlight-row'>
 {% for test_run in build.test_runs.all %}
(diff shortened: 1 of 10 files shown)


squad/frontend/templates/squad/build.html
```diff
@@ -53,6 +55,7 @@
 {% for test_run in build.test_runs.all %}
 {% for status in test_run.status.all %}
 {% if status.suite and status.has_metrics %}
+<a href="{% project_url test_run %}#metrics-{{status.suite}}">
```
mwasilew: This might be a risky thing to do in terms of HTML validity
https://stackoverflow.com/questions/1827965/is-putting-a-div-inside-an-anchor-ever-correct

terceiro: I just checked againt the [w3c validator](https://validator.w3.org/) and it passed. I just needed to added a few missing markup items to make it 100% valid HTML5 (thus commit  80f48d2)

PR #128, opened 2017-10-05
core: modify test_suites_by_environment to provide test count

mwasilew:
Build method test_suites_by_environment will not only provide sorted
list of environments for each build but also test counts
(pass,fail,skip) for each testsuite in each environment

### DIFF
--- a/squad/core/models.py
+++ b/squad/core/models.py
@@ -232,12 +232,20 @@ class Build(models.Model):
         result = OrderedDict()
         envlist = set([t.environment for t in test_runs])
         for env in sorted(envlist, key=lambda env: env.slug):
-            result[env] = set()
+            result[env] = dict()
         for tr in test_runs:
             for t in tr.tests.all():
-                result[tr.environment].add(t.suite)
+                if t.suite in result[tr.environment].keys():
+                    result[tr.environment][t.suite][t.status] += 1
+                else:
+                    result[tr.environment].setdefault(t.suite, {'fail': 0, 'pass': 0, 'skip': 0})
+                    result[tr.environment][t.suite][t.status] += 1
         for env in result.keys():
-            result[env] = sorted(result[env], key=lambda suite: suite.slug)
+            # there should only be one key in the most nested dict
+            result[env] = sorted(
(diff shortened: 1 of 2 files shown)


squad/core/models.py
```diff
@@ -232,12 +232,22 @@ def test_suites_by_environment(self):
...
         for tr in test_runs:
             for t in tr.tests.all():
-                result[tr.environment].add(t.suite)
+                if t.suite in result[tr.environment].keys():
+                    if t.status in result[tr.environment][t.suite].keys():
+                        result[tr.environment][t.suite][t.status] += 1
+                    else:
+                        result[tr.environment][t.suite].update({t.status: 1})
+                else:
+                    result[tr.environment].update({t.suite: {t.status: 1}})
```
terceiro: @you might want to use default values to simplify this. e.g. I think this entire block could be replaced with the following (untested):

```python
result[tr.environment].setdefault(t.suite, { True: 0, False: 0, None: 0 })
result[tr.environment][t.suite][t.status] += 1
```

squad/core/models.py
```diff
@@ -232,12 +232,22 @@ def test_suites_by_environment(self):
...
         for env in result.keys():
-            result[env] = sorted(result[env], key=lambda suite: suite.slug)
+            # there should only be one key in the most nested dict
+            result[env] = sorted(
+                result[env].items(),
+                key=lambda suite_dict: next(iter(suite_dict)).slug)
+
```
terceiro: why not just `suite_dict[0].slug` like you did in the tests?

mwasilew: I had dict of dicts here and that didn't work (sorting was commented)

squad/core/models.py
mwasilew: I expect we would get comment that zeros should not be listed in the notification. Probably could be sorted in the email template though.

mwasilew: Fixed

squad/core/models.py
mwasilew: Fixed

PR #132, opened 2017-10-12
Initial implementation of a plugin system

terceiro:


terceiro:
I noticed I was missing a small change to the migration. did that and rebased. the diff wrt the previous version is pretty small, it's just a trailing dot (".") in the `help_text` of the ` enabled_plugins_list` field.

mwasilew:
I checked some features that are not covered in unit testing and it looks OK. Merging

PR #135, opened 2017-10-19
permalinks for metrics don't work

terceiro:
When using a link generated for metrics, it doesn't work when sharing or trying to reopen. The graphs are missing. Sometimes the environments are not selected. Example:

https://qa-reports.linaro.org/qa/squad/metrics/?environment=travis-sqlite3&environment=travis-postgresql&metric=tests_per_sec

In this case environments are selected, but the chart is not displayed.

### DIFF
--- a/squad/frontend/static/squad/charts.js
+++ b/squad/frontend/static/squad/charts.js
@@ -223,8 +223,7 @@ function ChartsController($scope, $http, $location) {
             environment.line_color = colors[i][0]
             environment.fill_color = colors[i][1]
         })
-        /*_.each(_.castArray(params.environment), function(param) {*/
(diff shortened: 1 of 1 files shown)


terceiro:
Hi, turned this issue into a pull request.

@mwasilew the fix is to revert one commit you did some time ago, I'm not sure why. Can you please test on your side to make sure everything is ok?

terceiro:
well, I want o have this fixed ASAP, so I'm merging myself anyway, and we can discuss later

mwasilew:
It was done because the version of the JS libs we used to have didn't have the API yet. It was only present in the later versions of the library. Now that we pull the latest versions of JS libs the revert makes sense.

PR #137, opened 2017-10-20
New notification logic

terceiro:


terceiro:
Hi,

This is now completed with the other changes that were planned for the new notification logic.

The failed check (**continuous-integration/travis-ci/push**) failed because I pushed my branch to the target repository by mistake, and deleted it before travis was able to run tests against that new branch. The tests *are* passing, if you check the **continuous-integration/travis-ci/pr** check.

mwasilew:
I noticed it's just a pep8 failure, which isn't something critical. My PR fails for the same reason.

terceiro:
yeah I had a fix for that in this PR, I will cherry-pick it into master and remove it here

squad/core/tasks/notification.py
```diff
@@ -8,6 +10,28 @@
...
+    build = projectstatus.build
+    project = build.project
+
+    if project.wait_before_notification:
+        to_wait = project.wait_before_notification
+        time_passed = timezone.now() - build.datetime
+        if time_passed.seconds < to_wait:
+            # wait time did not pass yet; try again later
+            remaining_time = to_wait - time_passed.seconds + 1
+            maybe_notify_project_status.apply_async(
```
danrue: Any concern or attention to concurrency risk here? What if this gets called many times and then you have many of these execute concurrently - could it result in duplicate emails? Is the check at line 30 thread safe?

squad/core/tasks/notification.py
```diff
@@ -15,6 +15,10 @@ def maybe_notify_project_status(status_id):
     build = projectstatus.build
     project = build.project
 
+    timeout = project.notification_timeout
+    if timeout:
+        notification_timeout.apply_async(args=[status_id], countdown=timeout)
```
danrue: This guarantees a notification but only if maybe_notify_project_status has been called, but doesn't that require at least one result? What if no results have been received?

Same question about concurrency - could this result in duplicate emails if multiple notification_timeouts are queued up for the same moment?

squad/core/tasks/notification.py
```diff
@@ -7,14 +9,40 @@
...
-            raise self.retry(exc=not_found, countdown=30)  # retry in 30s
+@celery.task
+def maybe_notify_project_status(status_id):
+    projectstatus = ProjectStatus.objects.get(pk=status_id)
+    build = projectstatus.build
+    project = build.project
+
+    timeout = project.notification_timeout
+    if timeout:
+        notification_timeout.apply_async(args=[status_id], countdown=timeout)
```
danrue: (reposted comment because i originally posted within the commit and github thinks it's 'outdated')

This guarantees a notification but only if maybe_notify_project_status has been called, but doesn't that require at least one result? What if no results have been received?

Same question about concurrency - could this result in duplicate emails if multiple notification_timeouts are queued up for the same moment?

squad/core/models.py
```diff
@@ -203,25 +203,43 @@ def non_important_metadata(self):
...
         """
-        environments = self.project.environments.all()
-        expected = {e.id: e.expected_test_runs for e in environments}
+        # XXX note that by using test_jobs here, we are adding an implicit
+        # dependency on squad.ci, what in theory violates our architecture.
+        testjobs = self.test_jobs
+        if testjobs.count() > 0:
+            # If this build has CI jobs, then it's finished if there are no
+            # outstanding test jobs.
+            return testjobs.filter(fetched=False).count() == 0
```
mwasilew: This might cause issues in case there are both CI and non-CI results reported to the same project. In this case all non-automated test results have to be reported between submitting the automated TestJob and completing it. I think we need at least a paragraph on that in the docs.

squad/core/tasks/__init__.py
```diff
@@ -15,7 +15,7 @@
 from . import exceptions
 
 
-from .notification import notify_project_status
+from .notification import notify_project_status, maybe_notify_project_status
```
mwasilew: is the 'notify_project_status' used anywhere in __init__?

squad/core/tasks/notification.py
mwasilew: Both of the values in line 30 are cached in line 14. After digging a bit into code it turns out that the flags are only set after email is sent. This might take a moment. UpdateProjectStatus is called on each test job finishing. So a task is triggered at this point. We won't see concurrency issues on our instance as we're only running one worker now. But I think it's a valid comment and needs to be addressed. The problem will manifest with more than one worker. We'll have a few tasks trying to send_status_notification for the same ProjectStatus. They're all scheduled with 'countdown' firing around the same time.

danrue:
Sounds good to me @mwasilew

PR #140, opened 2017-10-25
frontend: prioritize display of failures and add filtering

terceiro:
- Test suites with failures are displayed on top
- Test suites without failures are hidden by default and can be toggled
- If there are no test suites with failures, everything is shown

### DIFF
--- a/squad/frontend/static/main.css
+++ b/squad/frontend/static/main.css
@@ -1,4 +1,4 @@
-.row {
+.row-bordered {
   margin-left: 0px;
   margin-right: 0px;
   margin-top: 5px;
@@ -7,7 +7,7 @@
   border: 1px solid #d3d7cf;
 }
 
-.highlight-row .row:hover
+.highlight-row .row-bordered:hover
 {
   background-color: #fffbcf;
 }
@@ -17,15 +17,15 @@
   color: inherit;
 }
 
-.metadata .row {
+.metadata .row-bordered {
   margin-top: 0px;
   border-top: none;
 }
-.metadata .row:first-child {
+.metadata .row-bordered:first-child {
   border-top: 1px solid #d3d7cf;
 }
 
-.row.fail {
+.row-bordered.fail {
   background-color: #f2dede;
   border-color: #a94442;
   color: #a40000;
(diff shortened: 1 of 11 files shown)


terceiro:
screnshots:

non-failures are hidden by default:

0_show

and can be displayed:

1_hide

terceiro:
the other features I'm working on build on top of the original commits I included in this PR, so I am appending them here.

terceiro:
build page, no filtering:

0_build_no_filter

build page, filtered:

1_build_filtered

test results page, no filter (not tests with failures on top):

2_test_results_no_filter

test results page, filtered:

3_test_results_filtered

danrue:
This looks like a really nice improvement to me. The only thing that caused me pause is the behavior "If there are no test suites with failures, everything is shown". This seems a bit unexpected/clever and most users may never realize why sometimes they get a list of tests and sometimes they don't, causing a (little) bit of cognitive overhead. I don't feel very strongly either way, but I do suspect it would probably be a little better if it was consistent and either always hid success or always showed success.

terceiro:
ok, good point. after I implemented the filtering, I also thought that the hiding the non-failures by default is a little too much, also because the failures are already on top. maybe we could drop the requirement to hide the non-failures?

danrue:
Showing by default is fine with me. I'm trying to be aware of the time it takes someone to get used to the UI, especially some of the more subtle parts.

danrue:
Sure showing them by default is fine. From my perspective, there's a lot of screen in squad and it can take a while to get used to each of them and the navigation (which can be subtle)

mwasilew:
I think we can drop the 'hide passed' feature once we have pagination. The list of tests/testjobs may become very long. So currently I would rather go with 'show nothing in case of all passed' than change the logic basing on results.

terceiro:
OK, so I reworked the code to always hide non-failures in the build page.

mwasilew:
@terceiro did you push your changes already?

terceiro:
@mwasilew yes!

terceiro:
(I rebased and applied the necessary changes do the existing commits)

mwasilew:
I think it's good to go now.

PR #141, opened 2017-10-31
frontend: floating table headers

mwasilew:
Table headers are floating. It's easier to see the environment if the
list of results is long.

### DIFF
--- a/squad/frontend/static/download.conf
+++ b/squad/frontend/static/download.conf
@@ -5,3 +5,4 @@ chartjs/Chart.bundle.js   https://github.com/chartjs/Chart.js/releases/download/
 font-awesome              http://http.debian.net/debian/pool/main/f/fonts-font-awesome/fonts-font-awesome_4.7.0~dfsg.orig.tar.gz a59bc27ace9906ae09cc8a49aac3b70899e0ae70
 lodash.js                 https://raw.githubusercontent.com/lodash/lodash/4.17.4/dist/lodash.js                 d8e7e155b43e7edcabc46682a809836a68444b01
 jquery.js                 https://code.jquery.com/jquery-3.2.1.js                                               fd81582bf1b15e6747472df880ca822c362a97d1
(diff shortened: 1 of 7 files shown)


squad/frontend/static/squad/table.js
```diff
@@ -0,0 +1,3 @@
+$('#tests-table').floatThead({
+    position: 'auto'
+});
```
terceiro: I would rather apply this to all tables, so that the .js could be loaded only once in the base template instead of every page that has a table.

terceiro: the table header needs to be style with CSS to have a white background: for me here it has a transparent backrgound which doesn't look nie.

terceiro:
also: please add `/squad/frontend/static/floatThread/` to .gitignore

squad/frontend/static/download.conf
```diff
@@ -5,3 +5,4 @@ chartjs/Chart.bundle.js   https://github.com/chartjs/Chart.js/releases/download/
 font-awesome              http://http.debian.net/debian/pool/main/f/fonts-font-awesome/fonts-font-awesome_4.7.0~dfsg.orig.tar.gz a59bc27ace9906ae09cc8a49aac3b70899e0ae70
 lodash.js                 https://raw.githubusercontent.com/lodash/lodash/4.17.4/dist/lodash.js                 d8e7e155b43e7edcabc46682a809836a68444b01
 jquery.js                 https://code.jquery.com/jquery-3.2.1.js                                               fd81582bf1b15e6747472df880ca822c362a97d1
+floatThread				  https://github.com/mkoryak/floatThead/archive/2.0.3.zip                               d7112698ed5bf14f953fdef6252fdd9950af90bf
```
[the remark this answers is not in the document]
mwasilew: ok, I'll try normal spaces. I'm not sure if there were any tabs to start with.

squad/frontend/static/squad/table.js
mwasilew: I liked transparent :)

mwasilew: fixed

mwasilew:
I think all comments are addressed now.

PR #147, opened 2017-11-13
Stability measures

terceiro:


squad/ci/models.py
```diff
@@ -27,9 +27,15 @@ class Backend(models.Model):
...
     )
     poll_interval = models.IntegerField(default=60)  # minutes
+    max_fetch_attempts = models.IntegerField(default=3)
 
     def poll(self):
-        for test_job in self.test_jobs.filter(submitted=True, fetched=False):
+        test_jobs = self.test_jobs.filter(
+            submitted=True,
+            fetched=False,
+            fetch_attempts__lt=self.max_fetch_attempts
```
danrue: Should we have some sort of 'failed to fetch' status? Will it be difficult to know the difference between a job that has not yet fetched and one that has failed to fetch 3 times?

terceiro: the one that was not fetched yet will have `fetch_attempts` equal to zero?

mwasilew: We discussed this with Dan yesterday. The main issue was how does this show up in UI. The failed jobs will have some text in TestJob.failure field. This field is presented in the UI. So if fetching fails I assume something is saved to TestJob.failure.

PR #149, opened 2017-11-15
Fix notifications for CI-based projects

terceiro:
When a testjob was fetched, the ProjectStatus would be created *before* the
test job was saved to the database, which would cause `Build.finished` to be
False, and therefore the notification would never go out. To fix that, we delay
updating the project status until after the test job is completely persisted.
An integration test is being added alongside the fix.

There are also some commits doing cleanups and small improvements, moving
things around, etc.

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,4 +1,4 @@
 web: ./manage.py runserver
 worker: ./manage.py celery worker --loglevel INFO
-scheduler: ./manage.py celery beat --pidfile=''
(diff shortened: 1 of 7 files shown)


mwasilew:
I did some testing locally. I didn't try to check the email notifications yet. What I ran into was the listener issue. The jobs submitted through watch_job api are not showing up in the log.

Since it's pretty urgent and watch_job is not our primary API, I think it's OK to merge.

PR #154, opened 2017-11-24
Add REST API

terceiro:
This is a basic implementation of a REST API using Django REST Framework. The
API root, /api, provides a browsable interface when accessed with a web
browser, and it can also display documentation. The actual documentation and
changes to make the API browser match the overall SQUAD layout (CSS, etc) will
be the subject of a subsequent pull request.

### DIFF
--- a/squad/api/rest.py
+++ b/squad/api/rest.py
@@ -105,6 +105,7 @@ class TestRunSerializer(serializers.HyperlinkedModelSerializer):
     metadata_file = serializers.HyperlinkedIdentityField(view_name='testrun-metadata-file')
     log_file = serializers.HyperlinkedIdentityField(view_name='testrun-log-file')
     tests = serializers.HyperlinkedIdentityField(view_name='testrun-tests')
+    metrics = serializers.HyperlinkedIdentityField(view_name='testrun-metrics')
 
     class Meta:
         model = TestRun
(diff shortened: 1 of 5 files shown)


squad/api/rest.py
```diff
@@ -22,6 +43,27 @@ class ProjectViewSet(viewsets.ModelViewSet):
...
 
+    @detail_route(methods=['get'])
+    def builds(self, request, pk=None):
+        builds = self.get_object().builds.order_by('-datetime')
+        page = self.paginate_queryset(builds)
+        serializer = BuildSerializer(page, many=True, context={'request': request})
+        return self.get_paginated_response(serializer.data)
+
+
+class BuildSerializer(serializers.HyperlinkedModelSerializer):
```
mwasilew: I'm getting the following message when trying locally:
```
"Creating a ModelSerializer without either the 'fields' attribute or the 'exclude' attribute has been deprecated since 3.3.0, and is now disallowed. Add an explicit fields = '__all__' to the BuildSerializer serializer.",
```
It applies to builds, testjobs, testruns and environments.

terceiro:
hi @mwasilew this should be fixed now.

mwasilew:
Works as expected. My only reservation is that API covers test jobs in the same place as all core objects. Should this be split somehow? I don't know whether the reusable apps can add own endpoints to the API. There is a similar situation with templates and frontend.

PR #155, opened 2017-11-29
API improvements

terceiro:


mwasilew:
I checked locally. All looks good and works well. The API docs miss a bit of formatting but that can be fixed later and I'm not sure it's that necessary anyway.

PR #156, opened 2017-11-30
Add filtering to the API

terceiro:


squad/api/rest.py
```diff
@@ -74,6 +74,7 @@ class ProjectViewSet(viewsets.ModelViewSet):
     """
     queryset = Project.objects
     serializer_class = ProjectSerializer
+    filter_fields = ('group',)
```
mwasilew: I would add slug and name here

squad/api/rest.py
```diff
@@ -137,6 +138,7 @@ class EnvironmentViewSet(ModelViewSet):
     """
     queryset = Environment.objects
     serializer_class = EnvironmentSerializer
+    filter_fields = ('project',)
```
mwasilew: I would add slug and name here

terceiro:
ok, added those fields and updated the pull request

mwasilew:
Works good enough. For some reason substring search doesn't work for me. According to the docs it should work. Let's merge it and fix the problems later.

PR #157, opened 2017-12-01
Upgrade to celery v4

terceiro:
Support for RDBMS-based periodic tasks is being dropped. It was not being
heavily used, anyway, and can be added back in the future by adding a new
dependency or celery v4.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,4 @@
 /dist
 /build
 /.metadata.json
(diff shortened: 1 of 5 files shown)


mwasilew:
Looks OK but needs rebasing after merging the REST API filtering.

terceiro:
rebased

PR #158, opened 2017-12-07
frontend: hide passing test results by default

terceiro:


mwasilew:
The link to full results isn't that obvious and easy to miss. Anyway, merging as it works as expected.

PR #159, opened 2017-12-18
Store and display metadata about environments, suites, tests, and metrics

terceiro:


terceiro:
## Screenshots 

### Default view of test run page:

screen shot 2017-12-18 at 18 55 03

### showing information about a test

screen shot 2017-12-18 at 18 55 07

### showing information about a metric

screen shot 2017-12-18 at 18 55 13

### showing information about environment

screen shot 2017-12-18 at 18 55 24

### showing information about suite

screen shot 2017-12-18 at 18 55 29

doc/intro.rst
```diff
@@ -177,6 +177,18 @@ strings. The following fields are recognized:
   makes no judgement about its value.
 * ``job_url``: URL pointing to the original test run.
 * ``resubmit_url``: URL that can be used to resubmit the test run.
+* ``suite_versions``: a dictionary with version number strings for suite names
```
danrue: How does this work if there are multiple versions of a given test suite across projects? (or what is the object model - does it belong to each project? i thought it was independent of projects)

danrue:
Overall this looks great. Is there anything that shows the object model? I got a little lost in this commit and commit message:

> Note that while Suite instances are project specific and Test/Metric
instances are created for each specific execution, these metadata
entries are unique; we assume that suites with the same slug in
different projects refer to the same suite in the real world.

I don't quite understand the implications of what is shared, what is per-project, and how we will need to adapt. (Will we end up putting i.e. the suite major version in the slug name to avoid comparing across incompatible suites?)

I also wondered about some of the UI changes. For example, moving metadata and downloads to the bottom of the screen on the testrun page. It seems counterproductive, especially since we can mask out the metadata we don't want to see. Nobody will notice it at the bottom of the page.

danrue:
We had discussed template support for reproduction instructions, but as-implemented it looks like it is a text field per-test. So instead of defining reproduction instructions per suite, we'll have to define it per-test?

doc/intro.rst
terceiro: the version is recorded at the level of the test run, so you can even have different versions of some test suite executed against the same build.

terceiro:
right now instructions for suite and test are displayed one after the other in the test instructions. I considered the template option, but I couldn't get to a decent and non-confusing solution for it.

terceiro:
what is shared and what is not: only the metadata for suite/test/metric is shared. everything else is not.

wrt the UI: some other people complain that the table with metadata is weird and unorganized, so I keep getting conflicting opinions about it all the time ...

terceiro:
my reasoning for moving metadata and downloads to the bottom is making the actual test results visible as soon as possible; "the rest" can be viewed by scrolling down.

squad/frontend/templates/squad/_test_run_test.html
```diff
@@ -1,18 +1,44 @@
...
+        <p><strong>Test description:</strong></p>
+        <div>
+        {{test.metadata.description|markdown}}
+        </div>
+        {% endif %}
+
+        {% if suite.metadata.instructions_to_reproduce or test.metadata.instructions_to_reproduce %}
+        <p><strong>How to reproduce:</strong></p>
+        <div>
+        {{suite.metadata.instructions_to_reproduce|markdown}}
```
mwasilew: I can't get the suite.metadata.instructions_to_reproduce displayed. The objects is set properly in the database but doesn't show up in the UI.

mwasilew:
I'm missing changes in LAVA backend to record the version of the test suite. I guess the problem is not critical and this feature can be added in separate PR.

squad/frontend/templates/squad/_test_run_test.html
terceiro: yes, my mistake. I just pushed a fixed version.

terceiro:
I ended up adding the LAVA support as an extra commit on this PR. Please take a look at the last commit.

mwasilew:
Looks good now. Should we add a mgmt action to populate SuiteVersion for historical data?

terceiro:
populating historical data will be very context-dependent. it's probably better to write a one-off scripts or something like that

PR #207, opened 2018-01-11
scripts: add script to migrate testruns between projects

mwasilew:
In case there is a need to move test runs from one project to the other
the script can move them basing on matching environment. Example call:
./manage.py runscript migrate_test_runs --script-args \
  env_slug=<environment slug> \
  old_project_slug=<from project slug> \
  new_project_slug=<to project slug>  --traceback

### DIFF
--- a/squad/core/management/commands/migrate_test_runs.py
+++ b/squad/core/management/commands/migrate_test_runs.py
@@ -1,0 +1,88 @@
+import sys
+from django.core.exceptions import ObjectDoesNotExist
+from django.core.management.base import BaseCommand
+
+from squad.core.models import Project, Build, Environment
+from squad.core.tasks import UpdateProjectStatus
+
+
+class Command(BaseCommand):
+
+    help = """Move test runs identified by environment slug
+    from one project to another. This action preserves
+    datetime of the objects and statuses."""
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            '--old-project-slug',
+            dest="old_project_slug",
+            help="Slug of the project from which to migrate test runs"
+        )
+        parser.add_argument(
+            '--new-project-slug',
+            dest="new_project_slug",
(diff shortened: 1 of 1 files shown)


mwasilew:
This feature was requested to split test runs from one project to two separate ones.

scripts/migrate_test_runs.py
```diff
@@ -0,0 +1,52 @@
...
+
+    for build in old_project.builds.all():
+        if build.test_runs.filter(environment=env):
+            print("moving build: %s" % build)
+            new_build, _ = Build.objects.get_or_create(
+                version=build.version,
+                project=new_project,
+                created_at=build.created_at)
+            for testrun in build.test_runs.filter(environment=env):
+                testrun.build = new_build
```
terceiro: do we need to care about the case in which there are no test runs left in the original build?

scripts/migrate_test_runs.py
```diff
@@ -0,0 +1,52 @@
...
+
+    if 'old_project_slug' not in parsed_args.keys():
+        print("ERROR: old_project_slug missing")
+        sys.exit(1)
+    if 'new_project_slug' not in parsed_args.keys():
+        print("ERROR: new_project_slug missing")
+        sys.exit(1)
+    if 'env_slug' not in parsed_args.keys():
+        print("ERROR: env_slug missing")
+        sys.exit(1)
```
terceiro: in a management command this would be handled by the framework, you just need to declare the arguments

mwasilew:
I guess you're right. I thought about mgmt command shortly after I submitted the PR. Let me change it. It makes little difference to the code and doesn't require django-extras as dependency.

mwasilew:
I updated to mgmt command. Now ready for review.

terceiro:
LGTM

PR #211, opened 2018-01-17
frontend: Add pagination to test and metric listings

terceiro:


mwasilew:
LGTM.
Should we change the display in the paginated pages to the table similar to 'all test results' page?

terceiro:
good point. I added an item to my TODO to try that out

PR #214, opened 2018-01-18
dev-docker: Return an interactive shell

danrue:
Run 'runserver' in the background and give an actual shell in the
foreground. Super handy!

### DIFF
--- a/dev-docker
+++ b/dev-docker
@@ -17,7 +17,7 @@ dockerfile="$datadir"/../Dockerfile.dev
   echo "RUN groupadd -g $(id -g) $(id -gn)"
   echo "RUN useradd -m -u $(id -u) -g $(id -g) -s /bin/bash ${USER}"
   echo "USER ${USER}"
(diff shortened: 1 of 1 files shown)


mwasilew:
LGTM

dev-docker
```diff
@@ -17,7 +17,7 @@ dockerfile="$datadir"/../Dockerfile.dev
   echo "RUN groupadd -g $(id -g) $(id -gn)"
   echo "RUN useradd -m -u $(id -u) -g $(id -g) -s /bin/bash ${USER}"
   echo "USER ${USER}"
-  echo 'CMD sh -c "./manage.py test && ./manage.py migrate && ./manage.py runserver 0.0.0.0:8000"'
+  echo 'CMD sh -c "./manage.py test && ./manage.py migrate && ./manage.py runserver 0.0.0.0:8000 & bash"'
```
terceiro: you can already run `./dev-docker bash` and get a shell. I don't think it's a good idea to start the serve in the background ...

PR #215, opened 2018-01-19
WIP: ci: frontend: add TestJob count numbers to build view

mwasilew:
TestJobs are counted by status. The number of each kind is displayed as
a badge in the build-nav.html. This patch also hides 'Test Jobs' tab
from the builds that don't include any test jobs.


mwasilew:
testjob_badges

The UI isn't complete yet but I wanted to get the opinion whether this is a good idea. Any comments are welcome.

squad/frontend/templates/squad/build-nav.html
```diff
@@ -18,11 +18,17 @@ <h1>
             All test results
         </a>
     </li>
+    {% if build.test_jobs.all %}
```
terceiro: you might want to use `count` instead of `all` to avoid fetching the test jobs since you only want to know if there are any

squad/frontend/templates/squad/build-nav.html
```diff
@@ -18,11 +18,17 @@ <h1>
...
+    {% load filter_jobs %}
     <li role="presentation" {% if url_name == 'testjobs' %}class="active"{% endif %}>
         <a href="{% build_section_url build 'testjobs' %}">
             Test jobs
         </a>
+        {% for i in build|filter_jobs %}
+        <span class="badge alert-success" data-toggle="tooltip" data-placement="top" title="Pass">{{ i.0 }} {{ i.1 }}</span>
```
terceiro: won't this make the navigation tab too wide? say we had at least 3 different job status ...

squad/frontend/templates/squad/build-nav.html
mwasilew: good point

squad/frontend/templates/squad/build-nav.html
mwasilew: Yeah, I know. The UI might be difficult to do given I don't want to hardcode the names of the job statuses in the templates. I'll try to find some solution. This part is WIP.

danrue:
I tried looking at this but I couldn't get the tooltip to show up.. Is this the same data that shows up on the 'latest builds' page or different?

mwasilew:
It should not be a tooltip but the labels under "test jobs" tab. I guess it needs rebasing.

mwasilew:
testjob_badges2

Addressed comments. Maybe this is OK now?

PR #217, opened 2018-01-19
frontend: paginate "all test results" page

terceiro:


squad/frontend/tests.py
```diff
@@ -6,43 +8,99 @@
...
+            # count pages
+            self.num_pages = count // per_page
+            if count % per_page > 0:
+                self.num_pages += 1
+
+    def __get_page_filter__(self, build_id, page, per_page):
+        """
+        Query to obtain one page os test results. It is used to know which tests
+        should be in the page. It's ordered so that the tests with more failures
+        come first, then tests with less failures, then tests with more skips.
```
mwasilew: This feature doesn't work for me. I checked on data from production db " lkft » linux-mainline-oe » Build v4.15-rc7-111-g5f615b97cdea". The failed tests don't show up on 1st or second page. Skips are on a few last pages.

mwasilew:
Filtering on 'all tests results' page stopped working. Does it only work on displayed results? If so it's better to disable it entirely.

squad/frontend/tests.py
terceiro: weird, it works for me here:

screen shot 2018-01-23 at 17 11 32

terceiro:
update PR removing the filtering (yes it is client-side so the filter applied only to the current page of data)

squad/frontend/tests.py
mwasilew: Antonio, this is a different view. I thought the PR is about 'all test results'. You're showing the paginated result for single TestRun. That works for me as well. 

Here is what I'm getting in 'all test results':
all_test_results

terceiro: yes! :-(

I found the issue, and pushed updates to this PR

PR #218, opened 2018-01-23
Trigger post-processing from admin

mwasilew:


squad/core/admin.py
```diff
@@ -134,9 +134,18 @@ def queryset(self, request, queryset):
         return queryset
 
 
+def force_execute_plugins(modeladmin, request, queryset):
+    for testrun in queryset.all():
+        postprocess_test_run.delay(testrun.id)
+
+
+force_execute_plugins.short_description = "Force plugin execution"
```
terceiro: The description should be "Postprocess selected test runs" and the name of the action should match.

We can use plugins in several different places in the future, postprocessing is just one of them.

mwasilew:
Action name updated. I also fixed the broken commits. Should be OK now.

PR #220, opened 2018-01-25
ci/lava: restore downloading logs

terceiro:


mwasilew:
It loads the whole log to memory first. Will it not cause OOM issues we had?

squad/ci/backend/lava.py
```diff
@@ -1,5 +1,6 @@
 import json
 import logging
+import requests
```
mwasilew: requests module is missing from requirements.txt

terceiro:
I did some tests here and the absolute majority of the memory usage in a similar usage scenario (open a large YAML file then feed it into yaml.load) is inside yaml.load itself, so that's not something we will be able to avoid.

On production usage, this can indeed cause issues with you have too little RAM and/or no swap. We do have swap in production, so this should be OK. Also the YAML object is local to that one method, so it will be garbage collected and once the method exits.

mwasilew:
OK. Let's have it like this and see what happens. If we start running into memory trouble I propose to process as we get each line. At the moment the logs are structured in such way that single line is still a valid YAML.

PR #221, opened 2018-01-29
ci: lava: parse LAVA log using incremental parser

mwasilew:
LAVA presents the log as YAML document. We're only interested in the
'device' level logs. So all other logs are discarded. In other to do
this operation, YAML content has to be parsed. Loading the whole log
into memory caused OoM issues in the past. This patch uses incremental
parsing approach in order to avoid loading full yaml to memory.

### DIFF
--- a/squad/ci/backend/lava.py
+++ b/squad/ci/backend/lava.py
@@ -193,23 +193,48 @@ class Backend(BaseBackend):
         return response.content
 
     def __get_job_logs__(self, job_id):
+        logger.debug("Retrieving logs job: %s" % job_id)
         log_data = self.__download_full_log__(job_id)
-        log_data_yaml = yaml.load(log_data, Loader=yaml.CLoader)
+        return self.__parse_log__(log_data)
+
+    def __parse_log__(self, log_data):
         returned_log = ""
-        for log_entry in log_data_yaml:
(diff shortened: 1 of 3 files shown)


mwasilew:
This approach uses less memory, but is pretty CPU intensive.

squad/ci/backend/lava.py
```diff
@@ -191,23 +191,37 @@ def __download_full_log__(self, job_id):
         return response.content
 
     def __get_job_logs__(self, job_id):
+        logger.debug("Retrieving logs job: %s" % job_id)
         log_data = self.__download_full_log__(job_id)
-        log_data_yaml = yaml.load(log_data, Loader=yaml.CLoader)
         returned_log = ""
-        for log_entry in log_data_yaml:
-            if log_entry['lvl'] == 'target':
-                if isinstance(log_entry['msg'], bytes):
```
terceiro: I am missing this bit from the new code ...

terceiro:
On my tests, this code is always faster than the current one (the larger the input, the larger the difference in performance).

squad/ci/backend/lava.py
mwasilew: I will add it.

PR #222, opened 2018-01-30
frontend: return 404 on non-existing data

terceiro:


mwasilew:
This works for me, but it doesn't use out 404.html template for some reason.

terceiro:
in development mode? (that would be normal)

mwasilew:
OK. I tried disabling DEBUG but that didn't help.

PR #223, opened 2018-02-01
api: fix TestSerializer

mwasilew:
Remove 'name' from excluded fields as this was causing server crash.


mwasilew:
This fixes
```
AssertionError at /api/testruns/8/tests/
Cannot both declare the field 'name' and include it in the TestSerializer 'exclude' option. Remove the field or, if inherited from a parent serializer, disable with `name = None`.
```

PR #224, opened 2018-02-06
frontend: show project description better

mwasilew:


mwasilew:
This isn't ideal, but slightly better than what we have now. I propose to merge it and than try to fix the CSS/layout so the text flows around the middle and right-most column.

terceiro:
closing in favor of #233

PR #225, opened 2018-02-13
ci: lava: add more strings to automatic resubmit list

mwasilew:


terceiro:
LGTM

This should probably not be hardcoded as these issues are specific to out lab and hardware, but ... meh.

PR #227, opened 2018-02-14
core: add support for test case variants

mwasilew:
In some cases (Android CTS) test come with different variants. Test
variants are identified with [foo] at the end of the test name. The
string in the brackets can contain slashes.

### DIFF
--- a/squad/core/utils.py
+++ b/squad/core/utils.py
@@ -18,8 +18,20 @@ def parse_name(input_name):
 
     parts = input_name.split('/')
     if len(parts) > 1:
-        group_name = '/'.join(parts[0:-1])
-        name = parts[-1]
+        # if test name is in format suite/complex.name[some/test/variant]
+        # don't split the name in brackets
+        # this is workaround for some CTS test names
+        if parts[-1].endswith("]"):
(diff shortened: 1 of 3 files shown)


squad/core/utils.py
```diff
@@ -18,8 +18,20 @@ def parse_name(input_name):
...
+            index = len(parts) - 1
+            # find index of the part that contains opening bracket
+            for part in reversed(parts):
+                if "[" in part:
+                    index = parts.index(part)
+            group_name = '/'.join(parts[0:index])
+            name = '/'.join(parts[index:])
+        else:
+            group_name = '/'.join(parts[0:-1])
+            name = parts[-1]
```
danrue: I was skeptical that this algorithm worked in all cases but I added some tests and couldn't find any bugs (see https://github.com/danrue/squad/commit/b1df44c547e52c760d397d977c42287682b623b2 if you want to add them). Even so, I think it might have be simpler to pull out the [string] using a regex or similar rather than a loop.

terceiro:
the code looks ok to me as well. I am a bit concerned about having so support any arbitrary naming convention out in the world, but let's go for this one for now

PR #229, opened 2018-02-15
Add support for self-service API tokens

terceiro:
This pull request adds a new per-user settings UI, with two initial panels:
profile (basic user info + email) and API key. API keys are now provided by the
Django REST Framework rest_framework.authtoken app, and old SQUAD-specific
tokens are migrated over to DRF ones.

A future pull request will remove the Token model from squad.core.models.


squad/core/migrations/0071_migrate_old_tokens.py
```diff
@@ -0,0 +1,47 @@
...
+        username = slugify(old.description)
+        user = User(username=username)
+        user_group = None
+        if old.project is None:
+            # global token → "staff" user
+            user.is_staff = True
+        else:
+            # project token → user member of the project group
+            group = old.project.group
+            user_group = group.user_groups.create(name='%s-submitters' % group.slug)
```
mwasilew: sth is wrong here (save() missing?). When running the migration all resulting tokens are 'global'. They don't have any project/group restrictions and the resulting users don't seem to be assigned to any group. It might be just the data from production is organized this way and we never noticed.

mwasilew:
It works for me. Moving the top link to the menu is a bit annoying but I understand the need for that. Also works on the narrow screen.

PR #232, opened 2018-02-19
ci: lava: remove hardcoded strings

mwasilew:
LAVA backend is able to automatically resubmit test jobs in case it
detects specific strings in LAVA error messages. These strings are now
part of settings instead of being hardcoded in the backend code.


mwasilew:
I added second commit that moves settings to DB. Not sure if the first commit makes any sense in this situation.

terceiro:
yeah, I would get rid of the first commit.

I would also use YAML instead of JSON as the former is less annoying to write.

squad/ci/backend/lava.py
```diff
@@ -268,6 +266,9 @@ def __get_publisher_event_socket__(self):
         return self.proxy.scheduler.get_publisher_event_socket()
 
     def __parse_results__(self, data, test_job):
+        LAVA_INFRA_ERROR_MESSAGES = []
+        if self.settings is not None:
+            LAVA_INFRA_ERROR_MESSAGES = self.settings.get('CI_LAVA_INFRA_ERROR_MESSAGES', [])
```
terceiro: drop the usage of all caps?

squad/ci/backend/null.py
```diff
@@ -1,3 +1,4 @@
+import json
```
terceiro: IMO YAML is easier to writer

PR #233, opened 2018-02-19
frontend: improve display of descriptions in project list

terceiro:


terceiro:
I think we can replace  #224 with this

mwasilew:
The main page rows are taller with this patch. I think it is possible to avoid that effect. One of the requested changes was to remove the horizontal line between description and result summary numbers. That should be an easy fix in CSS.

terceiro:
On Tue, Feb 20, 2018 at 02:28:40AM -0800, mwasilew wrote:
> The main page rows are taller with this patch. I think it is possible
> to avoid that effect. One of the requested changes was to remove the
> horizontal line between description and result summary numbers. That
> should be an easy fix in CSS.

PR updated with both changes.

mwasilew:
It doesn't have this nice text transition now, but I think it's OK.

PR #238, opened 2018-02-26
ci: lava: update listener to match new LAVA state machine

mwasilew:
Due to changes in LAVA state machine, the ZMQ events contain 2 fields
representing jobs status: 'state' and 'health'. 'state' is responsible
for marking the job is running or not. 'health' represents the status of
finished job (Complete, Incomplete, Canceled). This patch adjusts the
logic in lava listener code to work with the new dictionary keys.

### DIFF
--- a/squad/ci/backend/lava.py
+++ b/squad/ci/backend/lava.py
@@ -346,7 +346,7 @@ class Backend(BaseBackend):
             return
         if 'sub_id' in data.keys():
             lava_id = data['sub_id']
-        lava_status = data.get('status', 'Submitted')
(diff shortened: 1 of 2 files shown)


chase-qi:
Thanks fixing/merging it. I triggered a run. Results were fetched in 2 minutes. Looks good!

mwasilew:
The 2 min delay is artificial. It was introduced because at the time when we receive the signal the results aren't always stored in db. I wonder if the state machine rewrite in LAVA fixed this as well (so we can remove the 2 min delay).

PR #240, opened 2018-03-01
ci: run PostProcessTestRun after processing TestJob

mwasilew:
CTS plugin searches for logs based on TestJob results. Since TestJob was
not 'assigned' to the TestRun before plugin was executed, no logs were
retrieved. This patch forces second execution of the plugins after
TestJob and TestRun have proper state in database.


mwasilew:
This works locally, but I'm really not sure this is a correct solution.

mwasilew:
This is now addressed by #243

PR #247, opened 2018-03-21
Squad external config

mwasilew:


squad/api/rest.py
```diff
@@ -163,6 +164,36 @@ def testjobs(self, request, pk=None):
...
+        output_format = request.query_params.get("output", "text/plain")
+        template_id = request.query_params.get("template", None)
+        template = None
+        if template_id != "-1":  # query_params are strings
+            template = self.get_object().project.custom_email_template
+        if template_id is not None:
+            try:
+                template = EmailTemplate.objects.get(pk=template_id)
+            except EmailTemplate.DoesNotExist:
+                pass
```
danrue: I'm confused here starting at line 180. I suggest adding a test for pass template=-1, and also passing some valid template id. Is passing -1 the same as not specifying a template?

mwasilew: passing -1 will force using 'pre-built' squad templates. Passing nothing as a template will result in using template set for the project (if there is one).

terceiro: I would rather use an explicit placeholder name instead of -1, e.g. `"builtin"`

mwasilew: ok, I'll change it to string

PR #251, opened 2018-04-11
REST API fixes

mwasilew:


danrue:
Will this allow to specify ascending vs descending order?

mwasilew:
I added this note to the commit message. You're right, docs need to catch up. I don't think there is anything about searching or sorting in the docs.

PR #253, opened 2018-04-12
CI resubmit reference

mwasilew:


mwasilew:
I made minimal changes in the UI. It would probably be more clear to have a nested like structure in the TestJobs view, but that isn't trivial with django templates. It will be much easier when we move to jinja2 templates.

danrue:
Can the original job have some indication that it was resubmitted? Also, I still don't understand force resubmit vs resubmit :blush:

mwasilew:
Adding indication to the 'parent' is surely possible, but it can be parent for many jobs. So I'm not sure how to do that.

'Force resubmit' allows to resubmit any job, even the one that completed successfully and would not be possible to resubmit normally. I guess this option should only be visible to admins.

danrue:
Thanks. Any improvement is good :+1:

PR #256, opened 2018-04-17
frontend: display model.name instead of model.slug

mwasilew:
This patch changes how the UI displays the object names. Instead of
displaying 'slug', 'name' is displayed. This allows for more free-form
strings to be displayed in the UI. Since 'name' fields are mandatory
for Group and Project objects there is no need to make the defaults.


mwasilew:
This looks a bit awkward at first
names

PR #257, opened 2018-04-20
model: Mark optional fields optional

danrue:
These fields are marked as required, even though they may be null. When
editing projects in the administration interface, these fields become
required when any other change is made. Let them remain unset when
appropriate.


mwasilew:
I think this also needs a migration (./manage.py makemigrations)

PR #262, opened 2018-04-26
External config script

mwasilew:
I think I made it working in very basic cases. There are no tests for the script yet. It is likely there are bugs. I didn't test more complex features like adding group with projects inside.

Some operations are not supported for sure:
 - managing UserGroups in Group object
 - moving Projects across Groups
These 2 items would only be possible if object IDs were saved in the config file.

Work remaining to be done:
 - support multiple config files (right now config has to be stored in single file)
 - support external files for things like text input (to improve text formatting)

If the current approach is too complex (I think it is) we should consider other options. For example managing simple objects without any nested complex objects is fairly easy. Storing object IDs would make it way easier to work with. Right now this is WIP but it's already functional.

### DIFF
--- a/scripts/squad-config
+++ b/scripts/squad-config
@@ -1,0 +1,611 @@
+#!/usr/bin/env python3
+
+import argparse
+import logging
+import netrc
+import os
+import requests
+import sys
+import yaml
+
+from copy import deepcopy
+from jsonobject import *
+from jsonobject.base import get_dynamic_properties
+from pprint import pprint
+from urllib.parse import urlsplit, urlunsplit
+
+
+FORMAT = "%(funcName)20s() ] %(message)s"
+logging.basicConfig(level=logging.INFO, format=FORMAT)
+logger = logging.getLogger(__name__)
+
+
+# ExtLoaderMeta and ExtLoader copied from
+# https://gist.github.com/joshbode/569627ced3076931b02f
+class ExtLoaderMeta(type):
+
+    def __new__(metacls, __name__, __bases__, __dict__):
+        """Add include constructer to class."""
+
+        # register the include constructor on the class
+        cls = super().__new__(metacls, __name__, __bases__, __dict__)
+        cls.add_constructor('!include', cls.construct_include)
+
+        return cls
+
+
+class ExtLoader(yaml.Loader, metaclass=ExtLoaderMeta):
+    """YAML Loader with `!include` constructor."""
+
+    def __init__(self, stream):
+        """Initialise Loader."""
+
+        try:
+            self._root = os.path.split(stream.name)[0]
+        except AttributeError:
+            self._root = os.path.curdir
+
+        super().__init__(stream)
+
+    def construct_include(self, node):
+        """Include file referenced at node."""
+
+        filename = os.path.abspath(os.path.join(
+            self._root, self.construct_scalar(node)
+        ))
+        extension = os.path.splitext(filename)[1].lstrip('.')
+
+        with open(filename, 'r') as f:
+            if extension in ('yaml', 'yml'):
+                return yaml.load(f, ExtLoader)
+            else:
+                return ''.join(f.readlines())
+
+
+def yaml_validator(data):
+    yaml.load(data, Loader=ExtLoader)
+
+
+class SquadConnectionException(Exception):
+    pass
+
+
+class SquadConnection(object):
+    def __init__(self, args):
+        self.url = args.url
+        urlparts = urlsplit(self.url)
+        self.base_url = urlparts.netloc
+        self.url_scheme = urlparts.scheme
+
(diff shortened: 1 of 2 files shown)


mwasilew:
Ah, I forgot it depends on #261

scripts/squad-config
```diff
@@ -0,0 +1,502 @@
...
+        self.url_scheme = urlparts.scheme
+
+        connection_token = "Token %s" % self.__get_connection_token(args.url)
+        self.headers = {
+            "Authorization": connection_token
+        }
+        self.old_config = None  # SquadConfig()
+        self.old_backend_config = {}
+
+    def __get_connection_token(self, url):
```
terceiro: the usual convention for private methods is to also finish the method name with two underscores (`__`)

(ditto for the several other private methods)

scripts/squad-config
```diff
@@ -0,0 +1,502 @@
...
+                        "--save-to-file",
+                        default=None,
+                        help="Save current config to file",
+                        dest="save_filename")
+    parser.add_argument("-y",
+                        "--assume-yes",
+                        action="store_true",
+                        default=False,
+                        help="Assume 'yes' answer for all questions",
+                        dest="assume_yes")
```
terceiro: this option is not used for anything

scripts/squad-config
```diff
@@ -0,0 +1,502 @@
...
+
+
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument("-c",
+                        "--config-files",
+                        nargs="+",
+                        default=[],
+                        help="Instance config files",
+                        dest="config_files")
```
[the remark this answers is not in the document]
mwasilew: hmm, I don't think it's the case. You pass
--config-files file1.yaml file2.yaml file3.yaml ...
At least it works for me this way.

scripts/squad-config
```diff
@@ -0,0 +1,502 @@
...
+    parser.add_argument("-v",
+                        "--debug",
+                        action="store_true",
+                        default=False,
+                        help="Enable debug",
+                        dest="debug")
+    parser.add_argument("-u",
+                        "--url",
+                        help="SQUAD instance URL",
+                        dest="url")
```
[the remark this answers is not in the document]
mwasilew: +1

scripts/squad-config
mwasilew: ACK

scripts/squad-config
mwasilew: Arhg, it was used initially. It's required for using the script in non-interactive mode. I'll fix it.

mwasilew:
I moved the config script to use flat list instead of nested object. This makes the implementation a bit easier. I tested locally with various scenarios - adding, modifying objects, missing references, etc. There are probably a few scenarios that were not tested, but it should be OK (when used carefully). 

There is no object caching for the moment and there is no check for internal config file integrity. At least the latter should be added in order to make the script immune to typing mistakes. 

Also support for "include" clause was added. This should eliminate quite a few duplicate entries in the config. I think some docs are needed before merging. Any comments are welcome.

mwasilew:
I updated commit message as I didn't want to deal with Environments yet

scripts/squad-config
```diff
@@ -0,0 +1,604 @@
...
+        return req.prepare()
+
+    def __get_connection_token__(self, url):
+        netrcauth = netrc.netrc()
+        try:
+            self.username, _, self.token = netrcauth.authenticators(self.base_url)
+            logger.info("Using username: %s" % self.username)
+            return self.token
+        except TypeError:
+            logger.error("No credentials found for %s" % self.base_url)
```
terceiro: it should exit here; if a first attempt to get credentials don't work, there is no point in continuing

mwasilew:
Creating many objects at the same time is a problem. Could you send me the example of what you tried to create. Maybe it can be fixed.

scripts/squad-config
mwasilew: ACK

terceiro:
@mwasilew I don't have it anymore but it was something simple like:

```
- usergroup: foo
  attributes:
    name: foo

- group: foo
  attributes:
    name: foo
    slug: foo
    user_groups:
      - foo

- project: foo-bar
  attributes:
    group: foo
    slug: bar
    name: Bar
```

mwasilew:
I tried this example and it sort of worked. I can do group + usergroup but not all 3 objects. Exception is different, so I need to fix that first. 

There is a slight problem with project. There is no default set for emailtemplate. Since there is no emailtemplate set in the project settings you get 'too many objects found'. That is either a problem with API or the script tries to match None to some object. I'll try to fix it.

terceiro:
@mwasilew thanks. I guess we can go ahead with this initial version?

mwasilew:
@terceiro yep, let's try.

PR #264, opened 2018-05-03
api: reduce number of database queries for builds

terceiro:


danrue:
:+1: We can get finished from status.

PR #265, opened 2018-05-09
API optimizations

terceiro:


squad/api/rest.py
```diff
@@ -117,8 +117,7 @@ class ProjectViewSet(viewsets.ModelViewSet):
     """
     queryset = Project.objects
     serializer_class = ProjectSerializer
-    filter_fields = ('group',
```
mwasilew: Filtering projects by group is needed. Slugs are not unique across groups.

mwasilew: I checked locally and the query with and without this filter are the same.

mwasilew:
It seems that just removing filtering option from the UI reduces the number of queries in most severe cases (Build, Environment). 
```
diff --git a/squad/settings.py b/squad/settings.py
index 6d42861..b50c860 100644
--- a/squad/settings.py
+++ b/squad/settings.py
@@ -264,7 +264,8 @@ REST_FRAMEWORK = {
     'DEFAULT_FILTER_BACKENDS': (
         'rest_framework.filters.OrderingFilter',
         'rest_framework.filters.SearchFilter',
-        'django_filters.rest_framework.DjangoFilterBackend',
+        #'django_filters.rest_framework.DjangoFilterBackend',
+        'squad.middleware.DisabledHTMLFilterBackend',
     ),
     'PAGE_SIZE': 50,
     'DEFAULT_AUTHENTICATION_CLASSES': (
```
and middleware.py
```
from django_filters.rest_framework import DjangoFilterBackend

class DisabledHTMLFilterBackend(DjangoFilterBackend):

    def to_html(self, request, queryset, view):
        return ""
```

terceiro:
@mwasilew yeah ; just pushed a new version with your solution

PR #266, opened 2018-05-10
frontend: test history: add permalink with pinned top build

terceiro:


danrue:
The permalink is nice, but can the actual [1] [2] [3] page links be permalinks by default?

mwasilew:
@danrue I don't think it's possible. The [1][2][3] come from paginator and will change whenever there is a new build on the top. The length of the page is fixed. 

I'm not sure what the use case here is, but looks OK.

terceiro:
@mwasilew the use case is being able to provide a link to a specific page and not having the contents of that page change over time.

this would be possible if we change how pagination works, i.e. instead of [1][2][3], have [3][2][1], so the most recent page is actually the last, so e.g. page 2 will always have the same contents. I have an idea on how to do this, I'll give it a shot today

danrue:
What about using rest_framework.pagination.CursorPagination ? Wouldn't that better satisfy the points raised in https://coderwall.com/p/lkcaag/pagination-you-re-probably-doing-it-wrong ? Paging backwards seems like it will be confusing for users.. and I'm not sure how you'll avoid partial pages.

terceiro:
the issue is that rest_framework.pagination.CursorPagination is a Django REST framework thing, so it can used in the API, but not in the regular Django frontend.

mwasilew:
Unless we switch to JS based frontend with REST API as backend :)

terceiro:
after spending some time on it, I don't think it will be easy at all to do this cursor pagination ...

@danrue maybe we can live with what we have now in this PR now?

PR #268, opened 2018-05-15
Improve readability of page header and breadcrumbs

situmam:
* Imrpoved the look of the page header
* Used pills instead of tab since they are easier to locate on the page
* Remove the titles and icons from the candy pills for a cleaner look
* Removed pagination from the top of the page since it is redundant
* Added titles to tables for consistency
* Reduced the size of the headers by one level


situmam:
Forgot to rebase my work before doing the pull request. Will send a new pull request and cancel this one.  Sorry about the all the noise.

PR #269, opened 2018-05-15
Improve page overall readability

situmam:
-   Improved the look of  the page header
-     Used pills instead of tab since they are easier to locate on the page
-     Remove the titles and icons from the candy pills for a cleaner look
-     Removed pagination from the top of the page since it is redundant
-     Added titles to tables for consistency
-     Reduced the size of the headers by one level


mwasilew:
Looks OK with one exception. I don't like the project summary page:
build_summary_anmar

We have "latest builds" and "builds". The table contents look a bit 'bare' without table header. So it's hard to understand what the numbers in the coloured badges mean. 

The rest of the changes are OK. The background on the header looks a bit strange at first but makes sense after a moment (with breadcrumbs highlighted).

squad/frontend/templates/squad/_builds_table.html
```diff
@@ -1,62 +1,61 @@
 {% load humanize %}
 {% load squad %}
+<h3> Builds </h3>
```
mwasilew: This line adds redundant text to the page

squad/frontend/templates/squad/_builds_table.html
```diff
@@ -1,62 +1,61 @@
...
 <div class='highlight-row'>
```
mwasilew: It would be nice to add a header so the numbers in the table have some meaning.

squad/frontend/templates/squad/group.html
```diff
@@ -4,11 +4,10 @@
...
 
 {% block content %}
-<h1>{{group.name}}</h1>
-
-<p>
-{{group.description|default:""}}
-</p>
+<h2 class="page-header well">
+  {{group.name}}  <br>
+  <small>{{group.description|default:""}}</small>
```
mwasilew: Description might be lengthy. Please limit the number of characters here or provide JS to expand/collapse the lengthy description.

squad/frontend/templates/squad/project.html
```diff
@@ -8,13 +8,12 @@
 
 {% include "squad/project-nav.html" %}
 
-{% if last_build %}
```
mwasilew: this conditional should stay. The group might not have any builds.

squad/frontend/templates/squad/_builds_table.html
situmam: I think the actual builds table should be its own file and it then gets included and a title added. I realize this is a little bit too much work so instead, perhaps, it is best to remove "Latest Builds" and lave builds there.

squad/frontend/templates/squad/_builds_table.html
situmam: I thought about that but at the moment, but I thought i hold off as I like to replace the current rows with something more condensed and perhaps use some of the table js libraries out there to offer sorting, etc.   Not to mention this has to be styled along with the filters on some other pages.

squad/frontend/templates/squad/group.html
situmam: I saw sample code. I will fix it. What's the workflow here? Should I push a specific pull request just for this or add it as a patch to the future pull request?

squad/frontend/templates/squad/project.html
situmam: Fixed. Will have it ready for the next pull request

squad/frontend/templates/squad/_builds_table.html
mwasilew: This is not a table (as in <table> tag). It uses class=row from bootstrap to pretend to be table. I don't want <table> here.

squad/frontend/templates/squad/group.html
mwasilew: Adding a patch to this PR is OK. If the problem has to be fixed before merging (not in this case) you can amend the commit and force push to your branch (I'm using branches for PRs for this reason).

situmam:
Too many things changed here. As per my discussion with Dan, I will cancel this pull request and resend my changes in smaller chunks.

PR #270, opened 2018-05-15
Add plugin-related fields

terceiro:


mwasilew:
How do I enable/disable plugin? Currently all plugins seem to be enabled for all projects. Am I missing something?

terceiro:
`Project.enabled_plugins` will now be a list, which is empty by default, and the available items are the installed plugins. only selected plugins will apply. this is pretty much what I already have, except replacing a crappy text field with plugin names as text, now it's a nice select box.

mwasilew:
Ah, right. I missed the highlight. It would be best to have widget similar to group membership for User object. Now one needs to remember not to change the selection.

PR #271, opened 2018-05-15
ci/lava: allow resubmitting LAVA Test failures

mwasilew:
This patch broadens the list of potential test job error messages that
qualify for automated resubmission. Also messages can now be added to
LAVA ci backend setting as regular expressions.

### DIFF
--- a/squad/ci/backend/lava.py
+++ b/squad/ci/backend/lava.py
@@ -1,5 +1,6 @@
 import json
 import logging
+import re
 import requests
 import ssl
 import traceback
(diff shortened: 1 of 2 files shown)


squad/ci/backend/lava.py
```diff
@@ -327,8 +328,10 @@ def __parse_results__(self, data, test_job):
                         if metadata['error_type'] in ['Infrastructure', 'Lava', 'Job']:
                             completed = False
                         # automatically resubmit in some cases
-                        if metadata['error_type'] in ['Infrastructure', 'Job'] and \
-                                any(substring.lower() in metadata['error_msg'].lower() for substring in lava_infra_error_messages):
+                        infra_messages_re = re.compile("|".join(lava_infra_error_messages))
+                        if metadata['error_type'] in ['Infrastructure', 'Job', 'Test'] and \
+                                len(lava_infra_error_messages) > 0 and \
+                                infra_messages_re.search(metadata['error_msg']) is not None:
```
terceiro: are you dropping the case insensitivity (`.lower()`) on purpose?

terceiro: you should probably handle invalid regular expressions

terceiro: I'm in doubt wrt joining all the items like with (`"|".join(...)`). if one of the items is not a valid regular expression, you will invalidate the entire check. I would keep testing one by one, and ignoring the ones that are not valid regexes.

mwasilew:
I added re.I so the regex won't be case sensitive. Also added a check for incorrect expressions which are ignored. 

I had issues saving \d (or similar) strings into YAML in the tests. Not sure if that works when added in the admin interface.

PR #274, opened 2018-05-17
Improved the visual presence of the build and project tabs

situmam:
Improved the visual presence of the build and project tabs by using Bootstrap pills. 

Signed-off-by: Anmar Oueja <sitummam@gmail.com>

### DIFF
--- a/squad/frontend/templates/squad/build-nav.html
+++ b/squad/frontend/templates/squad/build-nav.html
@@ -5,7 +5,7 @@
     » <a class="h1 text-primary" href="{% build_url build %}">Build {{build.version}}</a>
 </h1>
 
-<ul class="nav nav-pills">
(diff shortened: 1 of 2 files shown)


danrue:
screenshot from 2018-05-17 11-11-22

Screenshot shows current view on top, new view on bottom. I don't mind the tabs. The pills are OK too, but may need some additional visual indication tying it into the rest of the page (like tabs have). I guess I don't care one way or another.

situmam:
Thanks for the feedback. As per my commit message, I meant for the pills to be visually more visible. This is a simple one word change helped but I agree that it isn't ideal.  This is something we will improve iterative as I like to have proper breadcrumbs deployed using the Bootstrap theme. At the point, I think we can bring back the tabs.

PR #275, opened 2018-05-18
Add support for patch builds

terceiro:
Terminology:

- patch build: a build that represents a patch set, merge request, pull
  request, etc.
- patch source: a system where patches are posted, and to where SQUAD will post
  notifications about the state of the corresponding build.
- patch id: an unique identifier of the patch at the patch source. the exact
  format is patch source-dependent.
- patch baseline: a build to which the patch will be compared (optional)

How it works:

- a build can be marked as being patch builds, by creating it on SQUAD via the
  `createbuild` API endpoint, and passig the patch source and patch id

- when this build is created on SQUAD, the patch source is notified that the
  patch has been recognized.

- when the build is finished (i.e. when it has enough test run data), the patch
  source is notified again, this time with the results of the test runs.

This is the initial infrastructure needed to support handling patch builds. The
actual patch source implementations (gerrit, github, etc) will follow.


mwasilew:
I have 2 questions to this version:
 - who/what will call "createbuild"? Should this be done from a build system (jenkins) that receives event notification from gerrit/github?
 - should we consider using some description/type in plugins? Currently all plugins will be displayed in the implementation field. This might be a bit confusing since the existing plugins are not meant for interaction with gerrit/github. Moreover they don't implement required methods.

terceiro:
On Tue, May 22, 2018 at 04:29:51AM -0700, mwasilew wrote:
> - who/what will call "createbuild"? Should this be done from a build
> system (jenkins) that receives event notification from gerrit/github?

yes.

in the future we could also have a webhook endpoint so squad can receive
events itself.

> - should we consider using some description/type in plugins? Currently
> all plugins will be displayed in the implementation field. This might
> be a bit confusing since the existing plugins are not meant for
> interaction with gerrit/github. Moreover they don't implement required
> methods.

hm, good point. I think we need to declare on each plugin which features
it implements to be able to filter the lists where we select plugins.

terceiro:
I have a working implementation of restricting the plugin fields to applicable plugins, but I think we should get this PR merged and I would create a new one for that

PR #286, opened 2018-06-08
core: detect changes from 'fail' to 'pass'

mwasilew:
'fail'->'pass' changes are named 'progression' in SQUAD. The
TestComparison.progression property was introduced to provide means of
reporting such changes

### DIFF
--- a/squad/core/comparison.py
+++ b/squad/core/comparison.py
@@ -108,37 +108,53 @@ class TestComparison(object):
         return self.__diff__
 
     __regressions__ = None
+    __fixes__ = None
 
     @property
     def regressions(self):
-        if self.__regressions__ is not None:
-            return self.__regressions__
-
-        if len(self.builds) < 2:
-            self.__regressions__ = {}
-            return self.__regressions__
-
-        regressions = OrderedDict()
-        after = self.builds[-1]  # last
-        before = self.builds[-2]  # second to last
-        for env in self.environments[after]:
-            regression_list = []
-            for test, results in self.diff.items():
-                results_after = results.get((after, env))
-                results_before = results.get((before, env))
-                if (results_before, results_after) == ('pass', 'fail'):
-                    regression_list.append(test)
-            if regression_list:
-                regressions[env] = regression_list
-
-        self.__regressions__ = regressions
+        if self.__regressions__ is None:
+            self.__regressions__ = self.__status_changes__()
         return self.__regressions__
 
     @property
+    def fixes(self):
+        if self.__fixes__ is None:
+            self.__fixes__ = self.__status_changes__(('fail', 'pass'))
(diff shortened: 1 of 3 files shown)


squad/core/comparison.py
```diff
@@ -108,37 +108,66 @@ def diff(self):
...
     @property
     def regressions(self):
-        if self.__regressions__ is not None:
+        return self.__regressions_progression__()
+
+    @property
+    def progressions(self):
+        return self.__regressions_progression__(False)
+
+    def __regressions_progression__(self, regression=True):
```
terceiro: this is a strange name. I would call it `status_changes`

squad/core/comparison.py
```diff
@@ -108,37 +108,66 @@ def diff(self):
         return self.__diff__
 
     __regressions__ = None
+    __progressions__ = None
```
terceiro: I don't recall ever hearing this term progression in this context. is it common?

maybe we want to call this "fixes"?

squad/core/comparison.py
```diff
@@ -108,37 +108,66 @@ def diff(self):
...
+        return self.__regressions_progression__()
+
+    @property
+    def progressions(self):
+        return self.__regressions_progression__(False)
+
+    def __regressions_progression__(self, regression=True):
+        comparison_tuple = ('pass', 'fail')
+        if not regression:
+            comparison_tuple = ('fail', 'pass')
```
terceiro: maybe instead passing a boolean and doing this test, you could pass in the comparison you want explicitly?

squad/core/comparison.py
```diff
@@ -108,37 +108,66 @@ def diff(self):
...
-        self.__regressions__ = regressions
+                if (results_before, results_after) == comparison_tuple:
+                    comparison_list.append(test)
+            if comparison_list:
+                comparisons[env] = comparison_list
+
+        if not regression:
+            self.__progressions__ = comparisons
+            return self.__progressions__
+        self.__regressions__ = comparisons
```
terceiro: I think this method should just return the list of status changes, and leave the assignment to the caller method

squad/core/comparison.py
mwasilew: We came up with this with @danrue. I'm OK calling it 'fixes' or whatever else that can be easily recognized.

squad/core/comparison.py
mwasilew: Hmm, might be a good idea. We can do n/a->pass (or fail) transitions as well. Not sure if they're useful but I heard a requirement for that some time ago.

danrue:
:+1: looks good to me.

mwasilew:
Rebased on top of current master. There should not be any more conflicts.

PR #287, opened 2018-06-08
api: order build results by ascending instead of descending

bhcopeland:
It would make sense to order the builds page by the latest first.

https://qa-reports.linaro.org/api/builds/

### DIFF
--- a/squad/api/rest.py
+++ b/squad/api/rest.py
@@ -187,7 +187,7 @@ class BuildViewSet(ModelViewSet):
     List of all builds in the system. Only builds belonging to public projects
     and to projects you have access to are available.
     """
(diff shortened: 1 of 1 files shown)


mwasilew:
@bhcopeland you're changing this view https://qa-reports.linaro.org/api/projects/117/builds/ and the order there is already as you need. What needs changing it probably around here: https://github.com/Linaro/squad/blob/master/squad/api/rest.py#L173

mwasilew:
the order_by('-datetime') should be added to line 190 to make it working as you require.

bhcopeland:
Thanks @mwasilew, I have updated the commit.

PR #288, opened 2018-06-13
Add support for Known issues

terceiro:
For now, the only UI available for managing known issues is the Django admin
interface, which requires giving users permission to use it in order to manage
known issues.

Known issues are displayed in 3 places:

- test results by test suite
- test history
- notification emails (added to the default template; data also available in
  for use in custom email templates)


squad/core/admin.py
```diff
@@ -149,6 +150,46 @@ class PatchSourceAdmin(admin.ModelAdmin):
...
 
 
+class SelectEnvironment(ModelMultipleChoiceField):
+
+    def label_from_instance(self, e):
+        return e.project.full_name + ' - ' + e.slug
+
+
+class KnownIssueAdminForm(ModelForm):
+    environment = SelectEnvironment(
```
mwasilew: this might become a very long list. Should we limit it somehow?

mwasilew:
Should we provide API endpoint to allow creating KnownIssue objects without accessing admin?

danrue:
An API endpoint (whether or not admin is required I don't think is critical) will be very important for managing known issues. Hopefully we can re-use the work in #262, though I recognize that it may require something custom just for known issue objects (so that we have more coherent known issue files). In any case, it doesn't have to be a part of this pull request and can be implemented subsequently.

mwasilew:
Adding an endpoint is a pretty simple thing and I probably can do it on top of this PR already now.

danrue:
Let's merge this so that we can start managing known issue lists. We'll also have to add UI elements to _project_list.html and _builds_table.html, and test_run.html (basically every location that has the pass/skip/fail badges), but I don't think we should hold up this review for it.

mwasilew:
API for known issues is under review in #290

mwasilew:
I added tests for KnownIssues to #290

PR #283, opened 2018-05-26
highlighted the test run environment and test suite in the page header

situmam:
Currently the test suite and test environment (squad talk for test board) are not easy to distinguish in the test run and test suite pages. This pull request adds them into the page header, adds an icon to easy for easy understanding and highlights the values so they are easy to pick out visually.

PR #289, opened 2018-06-15
Changed suite results on build page to be grouped by: 1) suite , 2) env

amrohassaan:
Added testresults grouping by suite and environment. So changed from [this view](https://imgur.com/a/SfUxas0) to [this view](https://imgur.com/a/788hfAe).


squad/frontend/views.py
```diff
@@ -102,13 +103,26 @@ def build(request, group_slug, project_slug, version):
...
+        if s.has_tests:
+            if s.tests_fail > 0 and s.suite not in suite_fails.keys():
+                suite_fails[s.suite] = True
+            if s.suite not in grouped_statuses.keys():
+                grouped_statuses[s.suite] = {s.test_run.environment: []}
+            if s.test_run.environment not in grouped_statuses[s.suite].keys():
+                grouped_statuses[s.suite][s.test_run.environment] = []
+            grouped_statuses[s.suite][s.test_run.environment].append(s)
+
+    count_statuses_nofail = len([s for s in _statuses if s.has_tests and s.tests_fail == 0])
```
mwasilew: these 2 variables can be also calculated in the loop above to save some cycles.

squad/frontend/templates/squad/build.html
```diff
@@ -13,51 +13,72 @@ <h2>Metadata</h2>
...
+                </a>
+            </div>
+            <div class='col-md-1 col-sm-1'>
+                <i class=' {% if suite in suite_fails %}fa fa-times text-danger {% else %} fa fa-check text-success {% endif %}'></i>
+            </div>
+        </div>
+        <div id="suite-{{suite.id}}" class="panel-collapse collapse">
+            <div class="panel-body">
+                {% for env, statuses in details.items %}
+                {% for status in statuses %}
```
mwasilew: indentation?

squad/frontend/views.py
```diff
@@ -102,13 +103,26 @@ def build(request, group_slug, project_slug, version):
         'suite',
     ).order_by('-tests_fail', 'suite__slug')
```
mwasilew: maybe a good idea do add ordering by environment name/slug as currently envs show up in random order.

squad/frontend/templates/squad/build.html
```diff
@@ -13,51 +13,72 @@ <h2>Metadata</h2>
...
+            </div>
+        </div>
     </div>
-</div>
-</a>
-{% endif %}
-{% endfor %}
+    {% endfor %}
 </div>
 {% if count_statuses_nofail > 0 %}
```
mwasilew: this section is always empty in the UI.

squad/frontend/templates/squad/build.html
amrohassaan: I followed the same model as shown from line 110 below where nested for loops are not indented. But sure. I will do that.

squad/frontend/templates/squad/build.html
amrohassaan: @mwasilew in fact i think this whole section needs to go now, since all results (passes+fails) are grouped by suite & env, hence they end up nested. As opposed to the previous state where all hierarchy is flattened, hence this previously made sense, while now no.
Also this is substituted for with the red x or the green check next to the suite and env headers inside a given accordion. eg. if on kselftest --> x86 grouping, one test run had fails while all others passed, we will end up with on red X among a number of green checks. That X will bubble up and be shown on both parent groupings: the env and its parent suite.

squad/frontend/views.py
amrohassaan: @mwasilew please see the comment in the last remark below as it relates to this one. These counters now need to be removed.

squad/frontend/views.py
amrohassaan: done

danrue:
+1 this is a nice improvement. We'll have to come back and add known issues support to this later.

mwasilew:
@amrohassaan we seem to have a problem. When I pull your changes, they invalidate all changes that were done in #283. Since #283 was just merged, could you rebase your PR on top of current master and preserver changes from @situmam?

mwasilew:
@amrohassaan looks OK to me. Could you squash this into single commit (I don't want half baked feature commits), fix the commit message (line length, missing header) and fix the pep8 error (missing empty line)?

PR #291, opened 2018-06-27
api: add 'badges' endpoint to project object

mwasilew:
'badges' endpoint will help rendering project badges with services like
shields.io. It currently has a problem with dynamically chaning colours
but otherwise works. Implementation details and interaction with
shields.io are described in the docs.

### DIFF
--- a/doc/intro.rst
+++ b/doc/intro.rst
@@ -203,6 +203,26 @@ When sending a proper metadata JSON file, other fields may also be
 submitted. They will be stored, but will not be handled in any specific
 way.
 
+
+Badges
+~~~~~~
+
+SQUAD offers project badges that can be used in the webpages
+
+::
+
+  https://<squad_instance_tld>/group/project/badge
+
+The colour of the badge matches the passed/failed condition.
+Following colours are presented:
+
+  * green (#5cb85c) when there are no failed results
+  * orange (#f0ad4e) when there are both passed and failed results
+  * red (#d9534f) when there are no passed results
+
+If there are no results, the badge colour is grey (#999)
+
+
 CI loop integration (optional)
 ------------------------------
 
--- a/requirements.txt
+++ b/requirements.txt
@@ -12,4 +12,5 @@ PyYAML
 pyzmq
 requests
 sqlparse
+svgwrite
 whitenoise
--- a/squad/frontend/urls.py
+++ b/squad/frontend/urls.py
@@ -18,6 +18,7 @@ urlpatterns = [
     url(r'^_/settings/', include(user_settings.urls)),
     url(r'^(%s)/$' % slug_pattern, views.group, name='group'),
     url(r'^(%s)/(%s)/$' % ((slug_pattern,) * 2), views.project, name='project'),
+    url(r'^(%s)/(%s)/badge$' % ((slug_pattern,) * 2), views.project_badge, name='project_badge'),
     url(r'^(%s)/(%s)/tests/(.+)$' % ((slug_pattern,) * 2), tests.test_history, name='test_history'),
     url(r'^(%s)/(%s)/metrics/$' % ((slug_pattern,) * 2), views.metrics, name='metrics'),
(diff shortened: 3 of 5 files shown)


mwasilew:
This isn't ideal as shields.io doesn't offer colour changes on dynamic badges. According to comments in shields.io tracker, this ticket is what we actually need: https://github.com/badges/shields/pull/1525. This will require slight changes in the SQUAD REST API but these changes are pretty trivial.

mwasilew:
Failing check here: https://travis-ci.org/Linaro/squad/jobs/397292746 has nothing to do with the patch. Sth wrong with version of PyYaml in travis. The tests pass locally. Same issue on different reviews.

danrue:
This is great! A few thoughts:

- No documentation on the /<group>/<project>/badge usage. It would be nice to not have the external dependency (and extra round trip too).
- I was really surprised that >90% is "green". In LKFT, that means that over 1000 things could fail and we'd still be green. Should either be configurable, or any failure >0 is not green. Also, I really care about eliminating failures - skips can be ignored for calculation purposes. Specifically, I'm suggesting:
  -  "green (#5cb85c) for pass rate above 90%" -> "green (#5cb85c) for fail rate of 0%"
  -  "orange (#f0ad4e) for pass rate between 50% and 90%" -> "orange (#f0ad4e) for fail rate between 0% and 10%"
  -  "red (#d9534f) for pass rate below 50%" -> "red (#d9534f) for fail rate above 10%"

screenshot from 2018-07-05 16-15-11

I tried using shields.io but had a bit of trouble. Do you have a working example I can try?

mwasilew:
There are 2 options here:
 - use shields.io (which isn't perfect and doesn't support dynamic colours)
 - use local SVG badges (no dependency as far as I understand)
I'll leave the API endpoint as it can be used in some other use case and makes no harm.

This is what I'm getting with some old results from mainline
badge_orange

Configurable thresholds would probably be best, but I don't know how to do that conveniently. Any ideas?

If we go with your proposal, what happens with fail rate > 0 and pass rate > 90%? This case isn't covered (or I'm misreading the conditions).

mwasilew:
I checked the appveyor tests in shields.io. The logic is pretty simple there:
 - # of passed = # of total tests -> badge is bright green
 - # of failed = 0 -> badge is green
 - # of passed = 0 -> badge is red
 - any case in between -> badge is orange
https://github.com/badges/shields/blob/master/server.js#L772

I'm fine with this logic. Any objections?

danrue:
I based my suggested thresholds on failures instead of passes because otherwise you have to worry about the number of 'skips' (and in the near future, known issues) reported. Really, 0 failures should be the goal of any CI, so that should be what determines green. So, I agree with your proposal with regard to green.

For orange vs red, it seems too extreme to say that you only get red if there are 0 passes. At least on the projects i'm looking at, I would personally set 'red' at something like 10 failures (not 10%; 10). Perhaps it needs to be configurable?

mwasilew:
Let's go with 0 passes for red for now. I'll try to make it configurable.

mwasilew:
OK, I pushed the changes as described.

squad/api/rest.py
```diff
@@ -112,6 +112,39 @@ class Meta:
...
+             project_status.tests_fail,
+             project_status.tests_skip)
+        pass_rate = 100 * float(project_status.tests_pass) / \
+            float(project_status.tests_pass + project_status.tests_fail + project_status.tests_skip)
+        output["project_pass_rate"] = "%0.2f" % pass_rate
+        # green for pass rate higher than 90%
+        # yellow for pass rate between 50% and 90%
+        # red for pass rate below 50%
+        colour = "f0ad4e"
+        if if project_status.tests_fail == 0:
```
terceiro: syntax error

mwasilew:
sorry, I didn't run tests. Let's remove the API then. I'm not very attached to it :)

mwasilew:
I added the docs for the standalone view and removed the API endpoint.

PR #298, opened 2018-07-12
api: add 'finished' flag to Build

mwasilew:
Build.status contains 'finished' flag. This flag is now exposed in the
build object in REST API.


terceiro:
this is missing a migration and also the flag is never set to anything. maybe you forgot to push some commits?

mwasilew:
Hmm, what migration do you mean? This works for me. It's just a change in API and it uses existing objects from database. If the ProjectStatus for the build is not set it will set finished=null. There is a problem when creating a build from API, but I guess we don't allow that anyway. I'll try to find solution to this as well just in case.

build_finished

mwasilew:
on the other hand we also have Build.finished property. It is calculated a bit differently than ProjectStatus.finished. Not sure if the Build.finished is used anywhere. Maybe it should be removed?

mwasilew:
OK, I fixed the problem with creating new objects from API.

terceiro:
@mwasilew I'm sorry, I think I reviewed the change too quickly, and though this was adding a field to the model, but it's not, so my comment about a migration makes no sense.

PR #300, opened 2018-07-16
Known failure implementation as new status result

saul-romero-arm:


mwasilew:
Could you squash these 2 commits so we don't add IDE files into the repository? Also did you check the KnownIssue feature that was recently merged? It looks pretty similar to what you're trying to achieve here.

terceiro:
closing as this is covered by the existing "known issues" features

PR #302, opened 2018-07-18
Filter test objects

mwasilew:


mwasilew:
This is 1st approach to changes in comparison page. I'm aware there might be some bugs, but I'd like someone to officially review.

squad/frontend/comparison.py
```diff
@@ -20,4 +20,5 @@ def compare_projects(request):
         'comparison': comparison,
     }
 
-    return render(request, 'squad/compare_projects.html', context)
+    #return render(request, 'squad/compare_projects.html', context)
```
terceiro: drop this commented line

mwasilew:
I think I have it working correctly now. I switched to tables similar to test history page. The list is now updated automatically without need to press 'go' button, which was removed. The list of names in suite and test fields is now filtered to only contain unique elements. It's pretty quick locally.

squad/frontend/templates/squad/base.html
```diff
@@ -73,7 +74,6 @@
     <script type="text/javascript" src="{% static "lodash.js" %}"></script>
     <script type="text/javascript" src="{% static "chartjs/Chart.bundle.js" %}"></script>
     <script type="text/javascript" src="{% static "floatThread/dist/jquery.floatThead.js" %}"></script>
-    <script type="text/javascript" src='{% static "squad/table.js" %}'></script>
```
terceiro: don't we want this for all tables? (at least in the past we did)

terceiro:
the attachment:
screenshot_2018-07-27 squad -

squad/frontend/templates/squad/base.html
mwasilew: It's very annoying for the table in the 'test compare' page. The header becomes floating and the table is pretty unusable.

mwasilew:
Adding group to project list might be done in 2 ways:
 - additional XHR call for each project
 - add group name to the project object API
The former sounds cleaner, but will take much longer (in terms of loading the page). The latter will be quick on the client side but will add additional SQL query on the database. So none of the solutions is great for the performance. Additional XHR call will be cheaper on database side (we can scale number of frontends).

I think it's a nice idea to have this page replace the default history page. Than the boxes with suite and test can be dropped and only the box with project will remain. @danrue @amrohassaan do you think it would be a good solution? I would have to add option to 'load more rows', but that can be done.

I fixed the missing div for alignment and removed the test name when suite name is removed. Next patch version will be uploaded shortly.

mwasilew:
I changed the UI a bit to be more consistent. For the moment the comparison page is separate but I intend to make it a default test history page. However this wasn't as easy as I expected, so I propose to merge this first to allow comparison between projects and than work on moving this view to test history page.

terceiro:
you forgot to include `squad/compare.html`

PR #305, opened 2018-07-27
frontend: add shortcut to latest finished build

terceiro:
My motivation for having this is to be able to automate some performance
benchmarks against the build page, without needing to hardcode a build.


mwasilew:
Please add a paragraph in the docs about the feature. I feel we have too many 'hidden gems' and any new tricks should be documented.

danrue:
Either in the docs or perhaps as a link somewhere on the builds page. +1 to the feature itself; it will be useful.

PR #306, opened 2018-07-31
api: allow for baseline selection with build/<id>/email API

### LINKED ISSUE #303: Be able to specify baseline when using email api
When calling the API like so:

https://qa-reports.linaro.org/api/builds/8275/email/

Note the line "compared to build v4.14.56-93-gec86d5e19e14".

Squad chooses the baseline (build v4.14.56-93-gec86d5e19e14 in this case) by naively using the previous build. However, my API client is able to determine a better baseline because it has knowledge of the project.

If the email endpoint allowed a 'baseline' build to be specified, perhaps like so: "https://qa-reports.linaro.org/api/builds/8275/email?baseline=8156", then the fixes and regressions that an email report produces would be much more useful.

mwasilew:
The email rendering API now allows for selecting custom comparison
baseline. The keyword is 'baseline' and the value should be Build object
ID. In case when Build with given ID doesn't exist HTTP 500 is returned.
Builds from different projects can be compared.

Fixes #303

### DIFF
--- a/squad/api/rest.py
+++ b/squad/api/rest.py
@@ -338,6 +338,7 @@ class BuildViewSet(ModelViewSet):
         """
         output_format = request.query_params.get("output", "text/plain")
         template_id = request.query_params.get("template", None)
+        baseline_id = request.query_params.get("baseline", None)
         template = None
         if template_id != "default":
             template = self.get_object().project.custom_email_template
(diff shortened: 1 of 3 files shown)


danrue:
Any chance for documentation for this 'hidden gem' ;) 👍

terceiro:
LGTM

regarding documention +1 in principle. But I think there are so many undocumented features at this point that we should probably do a large reorganization of the existing docs to make it easier for new features do get documented.

mwasilew:
I'm planning to write a paragraph of docs for this feature in a separate commit. Should we wait for it in this PR or merge and create a new PR?

terceiro:
let's merge. I think we should plan some concentrated documentation effort "soon"

PR #309, opened 2018-07-31
frontend: allow users to (un)subscribe to email notifications in a project

### LINKED ISSUE #185: Allow users to subscribe/unsubscribe to email notifications in a project
Imported from https://projects.linaro.org/browse/QA-1645
The goal is to allow users to subscribe do email notifications or remove the subscription. It will probably require some new 'profile' view.

stevanradakovic:
Add new panel to user settings which utilizes select2 js lib to allow
user to subscribe to projects in an easy manner.
Display subscriptions and allow for removal of those.
Allow admin to control user subscriptions.
Validate that no email and user can be present at the same time in a
subscription (and that at lease one must be present).

Fixes #185


doc/intro.rst
```diff
@@ -31,10 +31,11 @@ hardware platform, hardware configuration, OS, build settings (e.g.
 regular compilers vcs optimized compilers), etc. Results are always
 organized by environments, so we can compare apples to apples.
 
-Projects can have subscriptions, which are email address that should be
-notified about important events such as changing test results. ProjectStatus
-records the most recent build of a project, against which future results should
-be compared in search for important events to notify subscribers about.
+Projects can have subscriptions, which are users or manual email addresses that
```
terceiro: I suggest to phrase this as "which are either users of manually-entered email addreses that ..."

squad/core/notification.py
```diff
@@ -61,7 +61,13 @@ def summary(self):
...
     @property
     def recipients(self):
-        return [r.email for r in self.project.subscriptions.all()]
+        emails = []
+        for sub in self.project.subscriptions.all():
+            if sub.user and sub.user.email:
+                emails.append(sub.user.email)
+            else:
+                emails.append(sub.email)
+        return emails
```
terceiro: This duplicates the logic in `Subscription.__str__`. I think this logic should be in a method of `Subscription` (e.g. `get_email`), and both this code and `Subscription.__str__` should use it.

squad/frontend/templates/squad/user_settings/profile.html
```diff
@@ -16,7 +16,7 @@ <h1>Profile</h1>
                 {{form.last_name|add_class:"form-control"}}
             </div>
             <div class="form-group">
-                <label for="{{form.email.if_for_label}}">{{form.first_name.label}}</label>
+                <label for="{{form.email.if_for_label}}">{{form.email.label}}</label>
```
terceiro: ha, nice catch! I would like this to be in its own separate commit, though.

squad/frontend/user_settings.py
```diff
@@ -55,8 +57,42 @@ def api_token(request):
...
+@login_required
+def subscriptions(request):
+
+    subscriptions = Subscription.objects.filter(user=request.user)
+    groups = Group.objects.all().prefetch_related('projects')
+
+    if request.method == "POST":
+        for project_id in request.POST.getlist("subscriptions[]"):
+            try:
+                Subscription.objects.create(
```
terceiro: Note that calling `Subscription.objects.create` (instead of using a Django form) bypasses the model validation, maybe we don't want that.

squad/frontend/user_settings.py
```diff
@@ -55,8 +57,42 @@ def api_token(request):
...
+        'subscriptions': subscriptions,
+        'groups': groups
+    }
+    return render(request, 'squad/user_settings/subscriptions.html', context)
+
+
+@login_required
+def remove_subscription(request, id):
+
+    Subscription.objects.get(id=id).delete()
```
terceiro: you need to filter on the current user before deleting, oltherwise this opens a breach where a user can remove subscriptions from another user.

squad/frontend/user_settings.py
stevanradakovic: This is exactly why we do self._validate_email() in the overriden save method in Subscription model so it's not bypassed. Same would happen if you use API or some other means to create/update this specific model. I can add form here of course but I think it's redundant.

squad/core/notification.py
```diff
@@ -63,10 +63,9 @@ def summary(self):
     def recipients(self):
         emails = []
         for sub in self.project.subscriptions.all():
-            if sub.user and sub.user.email:
-                emails.append(sub.user.email)
-            else:
-                emails.append(sub.email)
+            email = sub.get_email()
+            if email:
```
terceiro: email is garanteed to always be something given the model validation in place :-)

squad/core/migrations/0078_user_subscriptions.py
```diff
@@ -0,0 +1,33 @@
+# -*- coding: utf-8 -*-
```
terceiro: this migration needs to be renamed. since you started working on this, new migrations were added, so I get this when trying it locally:

```
$ ./manage.py migrate
CommandError: Conflicting migrations detected; multiple leaf nodes in the migration graph: (0078_user_subscriptions, 0079_init_cache_test_run_counts in core).
To fix them run 'python manage.py makemigrations --merge'
```

you need to renumber it as `0080`, and make it depend on `0079_init_cache_test_run_counts`. just deleting it and re-running `./manage.py makemigrations` should DTRT.

terceiro: also please rebase on master to keep the PR branch clean

squad/core/notification.py
stevanradakovic: Well, not really. We guarantee that user or email will be present, but what might happen is that user's email is not populated (if I recall it's not mandatory in django base User model).
Do you think we should do something about that (i.e. display the warning on subscription panel if user did not set up his email address and also do a separate validation on admin side)?

terceiro: yes, I guess such a warning would be useful.

squad/core/migrations/0080_user_subscription.py
```diff
@@ -0,0 +1,33 @@
...
+import django.core.validators
+from django.db import migrations, models
+import django.db.models.deletion
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        migrations.swappable_dependency(settings.AUTH_USER_MODEL),
+        ('core', '0079_init_cache_test_run_counts'),
```
mwasilew: somehow we ended up with 2 identical migrations with only difference in dependency. This needs to be resolved (only one migration is needed) and rebased on current master.

squad/core/models.py
```diff
@@ -825,10 +827,40 @@ def __init__(self, build):
...
 class Subscription(models.Model):
     project = models.ForeignKey(Project, related_name='subscriptions')
-    email = models.CharField(max_length=1024, validators=[EmailValidator()])
+    email = models.CharField(
+        max_length=1024,
+        null=True,
+        blank=True,
+        validators=[EmailValidator()]
+    )
+    user = models.ForeignKey(
```
mwasilew: this field is empty by default. I think we need to fill it in with existing subscriptions in a migration. I checked a few users and none of them had any 'new' subscriptions. There were however 'old' ones with corresponding emails.

mwasilew:
Looks pretty good to me. There are 2 things to fix:
 - migration name and dependency
 - add group name to the banner describing project in subscriptions page.

PR #313, opened 2018-08-02
core: Build: check expected test runs even with CI jobs

terceiro:
Use case: 3 builds with 3 fast LAVA test jobs. Depending on the timing
of the builds, one or two builds will run and finish, the lava jobs will
finish, and notification will go out before the 3rd build has finished.

Build.finished would return without checking expected test runs when CI
(say, LAVA) jobs are involved.

Instead, we now return False (i.e. build is not finished) immediately if
there are any pending CI jobs, but if not, we still perform the checks
for the expected number of test runs for each environment.

The existing test case needed to be improved to create a more realistic
scenario, and a new test case for this use case has been added.

### DIFF
--- a/squad/core/models.py
+++ b/squad/core/models.py
@@ -276,9 +276,13 @@ class Build(models.Model):
         # dependency on squad.ci, what in theory violates our architecture.
         testjobs = self.test_jobs
         if testjobs.count() > 0:
-            # If this build has CI jobs, then it's finished if there are no
-            # outstanding test jobs.
-            return testjobs.filter(fetched=False).count() == 0
+            if testjobs.filter(fetched=False).count() > 0:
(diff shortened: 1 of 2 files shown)


danrue:
👍 thanks, this will help!

mwasilew:
The notification will still be sent on the timeout right? Than after all CI jobs complete, new notification will be triggered. At least this is how I think this should work. However with notification sent on a timeout we'll mark the ProjectStatus as notified and second (final) notification will not be sent.

terceiro:
yes, notifications on timeout will still work as before. this just avoids sending early notifications when we  know *there should be* data coming in "soon"

mwasilew:
OK, let's try it.

PR #314, opened 2018-08-10
Avoid duplicate notifications

terceiro:


danrue:
:+1:

mwasilew:
I'm not sure it helps. Usually the duplicate notifications are 'almost' identical. They only vary by 1 line, but that will be enough for hashes to differ as well.

PR #316, opened 2018-08-17
frontend: redesign build page for speed

terceiro:

PR #317, opened 2018-08-17
Build regression indication

### LINKED ISSUE #284: Add 'regressions' indication on the project summary page
In the project summary page, we list all builds , and we clearly see the number of tests for each builds, and the number of pass/failed. Can you please add an indicator to add the number of regressions as well? The data is available in the email report, but having it in the GUI would be very useful as well.

mwasilew:
Fixes #284


mwasilew:
It might be in conflict with #316 migrations.

mwasilew:
Some screenshots
build_regression1

build_regression2

squad/core/models.py
```diff
@@ -710,12 +722,28 @@ def create_or_update(cls, build):
         returns it.
         """
 
+        # avoid circular import
+        from squad.core.comparison import TestComparison
```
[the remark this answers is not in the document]
mwasilew: yeah, I know. I'm not sure how to avoid that without big code moves.

squad/frontend/templates/squad/_builds_table.html
```diff
@@ -4,8 +4,28 @@
 {% for status in statuses %}
 {% with build=status.build %}
 <a href="{% build_url build %}">
-<div class="row row-bordered build">
+<div class="row row-bordered build" ng-init="updateBuildRegressions({{ build.id }})">
```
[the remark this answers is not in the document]
mwasilew: IMHO it's better to have N async calls than do all queries before page loading. However in this case you're right. All data is already cached in ProjectStatus so I will try to drop XHR.

squad/frontend/static/squad/builds.js
```diff
@@ -10,25 +10,45 @@ function BuildsController($scope, $http) {
...
+                    html: true,
                     title: 'Fixes',
-		        	content: function () {
-				        return $('#fixes' + build_id).html();
-        			}
-		        });
+                    content: function () {
+                        return $('#fixes' + build_id).html();
+                    }
+                });
```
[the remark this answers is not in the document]
mwasilew: ugh, you're right. I didn't notice that.

mwasilew:
I removed XHR and simplified JS a bit. The circular import is still an issue though.

mwasilew:
I managed to remove the ugly import in the middle of the method but the solution isn't much prettier.

terceiro:
this needs to be rebased on master

mwasilew:
rebased, Conflicts resolved.

squad/core/models.py
```diff
@@ -707,12 +720,27 @@ def create_or_update(cls, build):
         returns it.
         """
 
+        # avoid circular import
```
terceiro: this can be removed

squad/core/models.py
```diff
@@ -724,7 +752,9 @@ def create_or_update(cls, build):
             'finished': build.finished,
             'test_runs_total': test_runs_total,
             'test_runs_completed': test_runs_completed,
-            'test_runs_incomplete': test_runs_incomplete
+            'test_runs_incomplete': test_runs_incomplete,
+            'regressions': regressions,
+            'fixes': fixes
```
terceiro: you also need to add fixes and regressions to the body of the if below. we could probably remove that redundancy with some meta-programming though.

squad/frontend/templates/squad/_builds_table.html
```diff
@@ -7,6 +7,26 @@
...
+            </div>
+            <div id="fixes{{ build.id }}">
+                <div ng-repeat="(fix_key, fix_value) in fixes[{{ build.id }}]">
+                    <h3><span ng-bind="fix_key"></span></h3>
+                    <ul>
+                        <li ng-repeat="test in fix_value"><span ng-bind="test"></span></li>
+                    </ul>
+                </div>
+            </div>
+        </div>
```
terceiro: you could just generate the data server side here, and keep the JS code just to show/hide the popover.

squad/frontend/templates/squad/builds.html
```diff
@@ -1,9 +1,28 @@
...
+{% for status in statuses %}
+    {% if status.regressions is not None %}
+        REGRESSIONS[{{ status.build.id }}] = {{ status.regressions|yaml2json }};
+    {% endif %}
+    {% if status.fixes is not None %}
+        FIXES[{{ status.build.id }}] = {{ status.fixes|yaml2json }};
+    {% endif %}
+{% endfor %}
+</script>
+<script type="text/javascript" src='{% static "squad/builds.js" %}'></script>
```
terceiro: if the actual HTML is generated server side this is not necessary

squad/frontend/templates/squad/project.html
```diff
@@ -9,21 +10,58 @@
...
+	</div>
+	<div id="fixes{{ last_build.id }}_last">
+		<div ng-repeat="(fix_key, fix_value) in fixes[{{ last_build.id }}]">
+			<h3><span ng-bind="fix_key"></span></h3>
+			<ul>
+				<li ng-repeat="test in fix_value"><span ng-bind="test"></span></li>
+			</ul>
+		</div>
+	</div>
+</div>
```
terceiro: ditto wrt generating this content server side

squad/frontend/templates/squad/project.html
```diff
@@ -9,21 +10,58 @@
...
+    {% if status.regressions is not None %}
+        REGRESSIONS[{{ status.build.id }}] = {{ status.regressions|yaml2json }};
+    {% endif %}
+    {% if status.fixes is not None %}
+        FIXES[{{ status.build.id }}] = {{ status.fixes|yaml2json }};
+    {% endif %}
+{% endfor %}
+</script>
+<script type="text/javascript" src='{% static "squad/builds.js" %}'></script>
+<script>
```
terceiro: ditto wrt not being necessary if the list of regressions or fixes is generated server side

squad/frontend/templates/squad/project.html
```diff
@@ -9,21 +10,58 @@
...
 $(document).ready(function(){
     $('[data-toggle="tooltip"]').tooltip();
 });
 </script>
 {% endblock %}
+
```
terceiro: spurious extra line

squad/frontend/templatetags/squad.py
```diff
@@ -166,3 +170,11 @@ def add_class(field, class_name):
...
     h = md5(email.encode('utf-8').strip().lower()).hexdigest()
     return 'https://www.gravatar.com/avatar/%s?s=%s&default=mm' % (h, size)
+
+
+@register.filter
+def yaml2json(text):
+    try:
+        return mark_safe(json.dumps(yaml.load(text), skipkeys=True))
+    except yaml.YAMLError:
+        return ""
```
terceiro: shouldn't be necessary if HTML for fixes/regressions is produced server side

test/core/test_project_status.py
```diff
@@ -189,3 +189,31 @@ def test_cache_test_run_counts_on_update(self):
...
+        test_run1.tests.create(name='foo', suite=self.suite, result=False)
+        ProjectStatus.create_or_update(build1)
+
+        build2 = self.create_build('2', datetime=h(9))
+        test_run2 = build2.test_runs.first()
+        test_run2.tests.create(name='foo', suite=self.suite, result=True)
+        status = ProjectStatus.create_or_update(build2)
+
+        self.assertIsNotNone(status.fixes)
+        self.assertIsNone(status.regressions)
```
terceiro: you could also have a test for updating fixes/regressions when new test runs arrive, to cover the update code path in `ProjectStatus.create_or_update`

squad/frontend/templates/squad/_builds_table.html
mwasilew: This isn't so obvious. The field value is serialized. Converting to object is done with JS. If this is done on server side at least templatetag is required.

mwasilew:
I think I fixed all comments

squad/core/migrations/0083_auto_20180816_1022.py
```diff
@@ -0,0 +1,26 @@
+# -*- coding: utf-8 -*-
```
terceiro: this migration needs to be renumbered because #324 which was merged since already has a migration `0083_*`

maybe take the opportunity and also give it a proper name

squad/core/migrations/0083_auto_20180816_1022.py
```diff
@@ -0,0 +1,26 @@
...
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+import squad.core.utils
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('core', '0082_populate_status_has_metrics'),
```
terceiro: this needs to be changed to point to the new `0083_` migration

squad/frontend/templates/squad/_regressions_popover.html
```diff
@@ -0,0 +1,49 @@
...
+            container: 'body',
+            html: true,
+            title: 'Fixes',
+            content: function () {
+                return $('#fixes' + {{ status.build.id }}).html();
+            }
+        });
+    {% endif %}
+{% endfor %}
+</script>
```
terceiro: this can be simplified by giving classes `popover-regressions` and `popover-fixes` to the buttons, adding "Regressions" and "Fixes" to a `title` attribute, hen doing something like this:

```javascript
jQuery(function($) {
    $('.popover-regressions, .popover-fixes').popover({
            placement: 'right',
            trigger: 'hover',
            container: 'body',
            html: true,
            title: function() { return $(this).attr('title') },
            content: function() { return $(this).parent().find('.hidden').html() }
    })
})
```

mwasilew: I think I got this working. PR updated.

PR #318, opened 2018-08-17
frontend: add indication of unfinished builds

### LINKED ISSUE #258: Add 'in progress' indicator to builds that are not 'finished'
Builds have a 'finished' property; but it is not reflected in the UI. It would be nice to be able to see when builds are in progress from the builds table, as well as from the individual build view.

mwasilew:
Fixes #258 

### DIFF
--- a/squad/frontend/templates/squad/_builds_table.html
+++ b/squad/frontend/templates/squad/_builds_table.html
@@ -6,6 +6,7 @@
 <a href="{% build_url build %}">
 <div class="row row-bordered build">
     <div class="col-md-2 col-sm-2">
+        {% if not build.status.finished %}<i title="Build/testing in progress" class="fa fa-cog fa-spin"></i>{% endif %}
         <strong>
             {{build.version}}
         </strong>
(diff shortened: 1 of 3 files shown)


squad/frontend/templates/squad/_builds_table.html
```diff
@@ -6,6 +6,7 @@
 <a href="{% build_url build %}">
 <div class="row row-bordered build">
     <div class="col-md-2 col-sm-2">
+        {% if not build.status.finished %}<i class="glyphicon glyphicon-refresh"></i>{% endif %}
```
terceiro: I think there should be an alt/title attribute with a description, for accessibility and also so that there is some sort of information about the "refresh" icon means.

also, the new bootstrap version dropped the glyphicons, so maybe we should be using FontAwesome here instead.

danrue: It would be nice to see % completed or number of jobs completed vs outstanding (if either are easy/possible). Maybe in the alt/title if there's no room for it on the page itself.

mwasilew: I would go with glyphicons for now. I'll create an issue to migrate to bootstrap 4 which should cover replacing all glyphicons to FA.

The number of test jobs is a bit harder to get in this view. We don't cache the number of test jobs in ProjectStatus so this would have to be queried 'live'. And it's at least 1 'count' query per build. This would be a significant performance hit.

mwasilew:
I didn't realize we're already using FA. So I switched to fa-cog instead of recycle icon. Also added 'title'

@danrue the number of test jobs is already displayed in the build page navbar.

PR #323, opened 2018-08-23
RFC: Bootstrap4

mwasilew:
All templates were migrated to bootstrap 4. This caused conflicts with DRF templates which use bootstrap 2. Our version overwrites DRF version. For this reason we have to carry additional templates for pagination as the default ones are not compatible with bootstrap 4. Also POST/PUT forms were dropped from web API. I don't think they're very useful for ordinary users anyway. Alternative is to leave Web API on it's own bootstrap and try to theme it to be in line with the rest of the page. Other changes are pretty self explanatory.

### DIFF
--- a/squad/api/renderers.py
+++ b/squad/api/renderers.py
@@ -1,0 +1,20 @@
+from rest_framework.renderers import BrowsableAPIRenderer
+
+
+class BrowsableAPIRendererWithoutForms(BrowsableAPIRenderer):
+    """Renders the browsable api, but excludes the forms."""
+
+    def get_context(self, *args, **kwargs):
+        ctx = super().get_context(*args, **kwargs)
+        ctx['display_edit_forms'] = False
+        return ctx
+
+    def show_form_for_method(self, view, method, request, obj):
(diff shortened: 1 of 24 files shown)


mwasilew:
I fixed the 'compare test' page (removed unsupported panels)

terceiro:
@mwasilew what's the status of this? does everything work for you?

I didn't review yet, but if you think it's ready, I think we should try to get this merged this right after the next release.

mwasilew:
I think I fixed all places with obsolete classes. What might be missing is new theme. B4 doesn't have any default theme and I'm not exactly sure how to make one. Theme isn't something we can't live without. I find some default colours of b4 annoying, but they kinda work.

mwasilew:
I rebased the branch on top of current master and removed a few more v3 specific bits. However I ran into a problem with 'popover'. It requires external JS library. For the moment I think I'll pause any work on this branch and focus on more important stuff. I'll push all changes I have to the branch today.

mwasilew:
Closing. I don't think it's time to move to B4 at this point.

PR #332, opened 2018-08-30
frontend: hide dropdowns in comparetest page

mwasilew:
The dropdowns for each of the search fields are now removed on lost
focus in the input field.


mwasilew:
I wanted to add keyboard support as well, but it seems to be more complicated than I expected. So only 1 commit to hide the dropdowns on blur event.

PR #334, opened 2018-08-30
Metadata separate view

mwasilew:


mwasilew:
This should solve STG-1188 from jira.

squad/frontend/views.py
```diff
@@ -225,16 +223,47 @@ def build(request, group_slug, project_slug, version):
...
+            finished=True,
+        ).order_by('build__datetime').last()
+        if status is None:
+            raise Http404()
+        build = status.build
+    else:
+        build = get_object_or_404(
+            project.builds.prefetch_related('test_runs'),
+            version=version,
+        )
```
terceiro: I don't think we need to support the "latest-finished" convention here, since we don't support it in any other view besides the main build page.

squad/frontend/views.py
```diff
@@ -225,16 +223,47 @@ def build(request, group_slug, project_slug, version):
...
+
+    context = {
+        'project': project,
+        'build': build,
+        'metadata': sorted(build.important_metadata.items()),
+        'extra_metadata': sorted(build.non_important_metadata.items()),
```
terceiro: maybe we should just return the entire metadata in the 'metadata' key, and then simplify the metadata template to render metadata in a single go

mwasilew: Yep, and only return important_metadata for the main view. Sounds like a good idea :)

mwasilew:
@terceiro I squashed and pushed.

PR #335, opened 2018-08-30
frontend: display known issues across the UI

terceiro:
Known issues/xfail results are now displayed on

* Build listings, both in the project page and in the full list of builds
* Build page
* Test run/suite page
* Test history page


terceiro:
this will probably need to have the migrations rebased after #317 is merged

terceiro:
screenshots:

project page
screen shot 2018-08-30 at 11 39 48

build page
screen shot 2018-08-30 at 11 39 58

test run/suite page
screen shot 2018-08-30 at 11 40 03

test (history) page
screen shot 2018-08-30 at 11 40 19

squad/core/history.py
```diff
@@ -9,6 +9,7 @@
 class TestResult(object):
 
     def __init__(self, test):
+        self.known_issue = test.known_issue
```
mwasilew: what happens if there is more than one active KnownIssue for a test?

squad/core/models.py
```diff
@@ -494,13 +494,22 @@ class Test(models.Model):
     name = models.CharField(max_length=256, db_index=True)
     result = models.NullBooleanField()
     log = models.TextField(null=True, blank=True)
+    known_issue = models.ForeignKey('KnownIssue', null=True, blank=True)
```
mwasilew: which KnownIssue will be selected if there is more than 1 active that matches the test name?

squad/core/models.py
```diff
@@ -494,13 +494,22 @@ class Test(models.Model):
...
 
     @property
     def status(self):
-        return {True: 'pass', False: 'fail', None: 'skip'}[self.result]
+        if self.result:
+            return 'pass'
+        elif self.result is None:
+            return 'skip'
+        else:
+            if self.known_issue_id:
```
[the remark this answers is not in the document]
terceiro: yes, this is an attempt at avoiding a database query here. every foreign key field has an implicit, associated `*_id` field. this will probably have to go away (see below)

squad/core/history.py
terceiro: good question. right now the code assumes there is only one, which is clearly wrong. I will have to rework this.

squad/core/models.py
terceiro: good point. this will probably have to be turned into a many to many relationship

terceiro:
rebased, and turned the relationship between test and known issue to many to many.

terceiro:
that was me being a lazy bastard and trying to cut corners in the test suite but PostgreSQL caught me. fixed now

mwasilew:
For some reason I can only see the known issues in the test comparison page. Do I need to do anything special after adding known issues from admin to get the data populated?

PR #345, opened 2018-09-11
frontend: Make it possible to display the charts page in fullscreen

### LINKED ISSUE #196: make it possible to display the charts page in fullscreen
Imported from https://projects.linaro.org/browse/QA-1547

stevanradakovic:
Add 'toggle fullscreen' button on top of each chart in metrics.

Fixes #196


mwasilew:
Looks great. Is there a reason why full screen is black while normal display is white (background of the chart)?

stevanradakovic:
I think that's browser specific. If you're talking about the area around the chart and not the chart itself... Otherwise, post a screenshot please.

terceiro:
I also get a black background (Firefox 62)

captura de tela de 2018-09-12 15-19-50

squad/frontend/static/main.css
```diff
@@ -212,3 +212,25 @@ table.test-results-details {
...
+
+.fullscreen-button:hover, .fullscreen-button:focus {
+  opacity: 0.75;
+  cursor: pointer;
+}
+
+.fullscreen-button:active {
+  opacity: 1;
+  box-shadow: 0 -3px 10px rgba(0, 0, 0, 0.1) inset;
+}
```
terceiro: do we really need all this CSS? I would rather have a regular bootstrap button

squad/frontend/static/squad/charts.js
```diff
@@ -247,6 +254,32 @@ function ChartsController($scope, $http, $location) {
...
+            }
+        } else {
+            if (document.exitFullscreen) {
+                document.exitFullscreen();
+            } else if (document.msExitFullscreen) {
+                document.msExitFullscreen();
+            } else if (document.mozCancelFullScreen) {
+                document.mozCancelFullScreen();
+            } else if (document.webkitExitFullscreen) {
+                document.webkitExitFullscreen();
```
terceiro: what browser versions are we  supporting exactly with all these else's?

squad/frontend/static/main.css
terceiro: also, font awesome has a fullscreen icon, maybe we could use that? 

My suggestion is to remove the in-chart title, turn that into a `<h2>`, and put the full screen button close to that header. and move the "Toggle full screen" to the title attribute of the button

PR #348, opened 2018-09-13
frontend: display reason for build being unfinished

terceiro:


squad/frontend/templates/squad/_unfinished_build.html
```diff
@@ -1,2 +1,6 @@
-{% if build and not build.status.finished %}<i title="Build/testing in progress" class="fa fa-cog fa-spin"></i> {% endif %}
+{% if build and not build.status.finished %}
+{% with reasons=build.finished.1 %}
+<i title='Build/testing in progress ({{reasons|join:"; "}})' class="fa fa-cog fa-spin"></i>
```
mwasilew: I'll merge that, but maybe we should add not only title but also the popover (like with regressions and fixes)?

terceiro:
right! I will do that.

consistency FTW

PR #350, opened 2018-09-17
doc: describe API endpoint for getting metric data

terceiro:


mwasilew:
I think we should also add support for CSV export and exporting multiple metrics in the same request. But that's out of scope for docs.

PR #357, opened 2018-09-19
frontend: Limit number of points in charts.

### LINKED ISSUE #200: limit number of points in charts
Imported from https://projects.linaro.org/browse/QA-1543
Having too much data (e.g. charting the entire history) can make the visualization process slower without necessarily making it more useful. squad should have a default limit in metrics results (e.g. latest 50 measurements/tests), and allow the user to explicitly request more than that if necessary.

It should also allow the user to select a particular period (e.g. start/end dates).

stevanradakovic:
Add hard limit of 50 points for each env in chart.
Allow user to control % of shown points via range slider.

Fixes #200

### DIFF
--- a/squad/frontend/static/main.css
+++ b/squad/frontend/static/main.css
@@ -224,3 +224,38 @@ table.test-results-details {
 .popover {
   min-width: 400px;
 }
+
+.slider {
+  position: relative;
+  margin: auto;
+  border: 1px solid #c0c0c0;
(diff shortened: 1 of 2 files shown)


terceiro:
@stevanradakovic it works great!

I have only a small suggestion for a layout change: see terceiro/squad@ca2666d

so, please apply that, and squash this PR into a single commit and we will get it merged

PR #361, opened 2018-09-28
api: fix pagination in TestRun detail routes

terceiro:
It was using the pagination settings, including field names etc from
TestRun to paginate Test's and Metric's.

### DIFF
--- a/squad/api/rest.py
+++ b/squad/api/rest.py
@@ -12,7 +12,7 @@ from rest_framework import routers, serializers, views, viewsets, status
 from rest_framework.decorators import detail_route
 from rest_framework.exceptions import NotFound
 from rest_framework.response import Response
-from rest_framework.pagination import CursorPagination
+from rest_framework.pagination import CursorPagination, PageNumberPagination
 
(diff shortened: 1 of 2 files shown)


squad/api/rest.py
```diff
@@ -622,18 +622,20 @@ def log_file(self, request, pk=None):
     @detail_route(methods=['get'], suffix='tests')
     def tests(self, request, pk=None):
         testrun = self.get_object()
-        tests = testrun.tests.prefetch_related('suite')
-        page = self.paginate_queryset(tests)
+        tests = testrun.tests.prefetch_related('suite').order_by('id')
+        paginator = PageNumberPagination()
```
mwasilew: Maybe CursorPagination would be better here? There can be a lot of tests and count(*) required for PageNumberPagination is very expensive.

mwasilew: I take this back. Just pulled the code and tested. It works. The downside is that the API doesn't provide the number of pages. But I can live with that (at least for now).

PR #362, opened 2018-10-02
doc/quickstart.rst: Add missing system dependency

chaws:
When running squad locally, i.e., installing all
dependencies with pip3 still misses out on yaml.CLoader[1]
(https://github.com/linaro/squad/blob/master/squad/ci/backend/lava.py#L209)

Just added the system dependency 'libyaml-dev' so that
when pip installs all requirements, CLoader will be available
for 'yaml' module.

[1] https://github.com/yaml/pyyaml/issues/108

### DIFF
--- a/doc/quickstart.rst
+++ b/doc/quickstart.rst
@@ -16,7 +16,7 @@ To install the dependencies::
 Alternatively to using pip, on Debian stretch or later you can install
 dependencies from the repository::
 
-    apt-get install python3-dateutil python3-django python3-celery \
+    apt-get install python3-dateutil python3-django python3-celery libyaml-dev \
       python3-django-celery python3-jinja2 python3-whitenoise python3-zmq
 
 To run the tests::


doc/quickstart.rst
```diff
@@ -16,7 +16,7 @@ To install the dependencies::
 Alternatively to using pip, on Debian stretch or later you can install
 dependencies from the repository::
 
-    apt-get install python3-dateutil python3-django python3-celery \
+    apt-get install python3-dateutil python3-django python3-celery libyaml-dev \
```
terceiro: note the paragraph above says "alternatively to using pip", so at this point it's already too late. we probably need a new paragraph before that, or an extra line before `pip install` with the required -dev  packages. One probably also wants libpq-dev for postgres, etc.

chaws:
Closing this one, not a priority right now

PR #367, opened 2018-10-05
core: TestComparison: optimize database access

terceiro:
This further improves the performance problems introduced in 0.51. On a
sample build, I have got the following number of SQL queries per request
for /api/builds/${BUILD}/email/:

* 0.50: 250
* master post-0.51: 416
* master with this patch: 91

For sure, there is room for further improvement, but this already
improves upon the state of 0.50 which was at least good enough.


terceiro:
I'm using this as my test case for the DB access performance testing I'm working on, so I don't want this merged yet.

PR #370, opened 2018-10-11
core: implement data retention policy

terceiro:
Each project can specify their own data retention policy, defaulting to
180 days. Builds older than that will be removed daily.


terceiro:
this is implemented on top of #369 so only the last commit is relevant here

terceiro:
@danrue I would like your feedback on this.

test/core/test_tasks.py
```diff
@@ -438,3 +440,38 @@ def test_dont_notify_without_patch_source(self, notify_patch_build_created):
...
+        self.project.data_retention_days = 60
+        self.project.save()
+        cleanup_old_builds()
+        cleanup_build.delay.assert_called_once_with(build.id)
+
+    def test_cleanup_build(self):
+        build_id = self.project.builds.create(version='1').id
+        cleanup_build(build_id)
+        self.assertFalse(self.project.builds.filter(id=build_id).exists())
+        self.assertFalse(self.project.builds.filter(id=build_id).exists())
```
mwasilew: duplicate line?

mwasilew:
Looks good to me. Let me send email with this change to our major users and see if they're OK with that.

terceiro:
hi all, I just updated this branch with a final version of what I think the
data retention mechanism should look like. Changes from the initial proposal:

- no cleanup is done by default
- when accessing a build that was deleted, the 404 Not Found page will say
  explicitly that.

squad/core/migrations/0096_build_keep_data.py
```diff
@@ -0,0 +1,20 @@
...
+
+    dependencies = [
+        ('core', '0095_project_data_retention_days'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='build',
+            name='keep_data',
+            field=models.NullBooleanField(help_text='Keep this build data even after the project data retention period has passed'),
```
mwasilew: Shouldn't we rather use BooleanField with default set to False?
https://docs.djangoproject.com/en/2.1/ref/models/fields/#booleanfield

terceiro: fair enough. Changed

mwasilew:
Let's see how that behaves in staging

PR #379, opened 2018-10-25
core: add tracking to EmailTemplate changes

### LINKED ISSUE #169: track history of custom templates
Imported from https://projects.linaro.org/browse/QA-1740

chaws:
It uses https://github.com/treyhunner/django-simple-history.
Simply added a 'history' field in the EmailTemplate model and set up
some settings to get it integrated with Django Admin.

Closes #169

NOTE: django-simple-history creates a 1-to-many relationship against the model,
keeping a whole copy of the changed model. There are ways to exclude fields
from being tracked, if needed.


test/core/test_historical_emailtemplate.py
```diff
@@ -0,0 +1,64 @@
...
+        self.assertEqual(original.html, 'html')
+
+    def test_revert_to_previous_version(self):
+        self.email_template.subject = 'other subject'
+        self.email_template.save()
+
+        self.email_template = self.history.earliest().instance
+        self.email_template.save()
+        self.assertEqual(len(self.email_template.history.all()), 3)
+        self.assertEqual(self.email_template.name, 'name')
```
terceiro: before testing that the attribute values are back, you need to fetch the object from the database again, otherwise you are just testing the in-memory instance that you already have, and the assestions will pass even if that was not saved to the database (e.g. if you comment the call to `save()`, the assertions about the attributes would still pass)

test/core/test_historical_emailtemplate.py
```diff
@@ -0,0 +1,64 @@
...
+        self.assertEqual(self.email_template.subject, 'subject')
+        self.assertEqual(self.email_template.plain_text, 'plain text')
+        self.assertEqual(self.email_template.html, 'html')
```
terceiro: for these assertions to be meaningful, you need to change them before reverting

test/core/test_historical_emailtemplate.py
```diff
@@ -0,0 +1,64 @@
...
+        self.email_template.save()
+
+        self.history = self.email_template.history
+
+    def test_history_creation(self):
+        self.assertEqual(len(self.history.all()), 1)
+
+    def test_new_version(self):
+        """
+        This is the main test, for default, any field that has been changed
```
terceiro: /for default/by default/

squad/settings.py
```diff
@@ -94,6 +95,7 @@
     'django.contrib.auth.middleware.SessionAuthenticationMiddleware',
     'django.contrib.messages.middleware.MessageMiddleware',
     'django.middleware.clickjacking.XFrameOptionsMiddleware',
+    'simple_history.middleware.HistoryRequestMiddleware',
```
terceiro: where is this being used?

chaws: There are some templates for Django's admin area that simple-history uses, thus requiring it to be listed in INSTALLED_APPS: https://github.com/treyhunner/django-simple-history/tree/master/simple_history/templates/simple_history.

The middleware was to enable simple-history to keep record of which user changed the template. But I looked further in their code and that is being done already in SimpleHistoryAdmin. I'm removing this line.

chaws: The middleware was to enable simple-history to keep record of which user changed the template. But I looked further in their code and that is being done already in SimpleHistoryAdmin. I'm removing this line.

test/core/test_historical_emailtemplate.py
chaws: Agreed, I changed that so the model can fetch fresh data and then check if it's been reverted.

test/core/test_historical_emailtemplate.py
chaws: Added changes to that too.

squad/core/migrations/0093_populatehistoricalemailtemplate.py
```diff
@@ -0,0 +1,30 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.11.16 on 2018-10-24 19:49
+from __future__ import unicode_literals
+
+from django.db import migrations
+import simple_history.utils
+
+
+from squad.core.models import EmailTemplate
```
terceiro: usually it's not a good idea to use model classes directly in data migrations. instead, inside the function you do something like this:

```python
EmailTemplate = apps.get_model('core', 'EmailTemplate')
```

mwasilew: hmm, it's not only a bad idea, in my experience it simply doesn't work. Creating object instances in migrations for some reason is broken. I guess this is intentional and the only proper way of creating object instances is what Antonio presented.

chaws: Humm, I agree. But one thing this that's bugging me is that `EmailTemplate = apps.get_model('core', 'EmailTemplate')` retrieves a `__fake__.EmailTemplate` class, which doesn't contain `history` attribute (only 'id', 'name', 'subject' etc). It's just a fake object with all fields cloned into it. Thus the migration fails due to missing attribute 'history'.

Strangely, if using `from django.apps import apps as django_apps` and `EmailTemplate = apps.get_model('core', 'EmailTemplate')` then it'll return `squad.core.models.EmailTemplate`, which is the correct one.

In the [documentation](https://docs.djangoproject.com/en/1.7/topics/migrations/#data-migrations) the `apps` passed on `migrations.RunPython` should be the same as `django.apps.apps`, the app registry, but it's behaving differently. 

Any ideas on a workaround?

terceiro: > Humm, I agree. But one thing this that's bugging me is that `EmailTemplate = apps.get_model('core', 'EmailTemplate')` retrieves a `__fake__.EmailTemplate` class, which doesn't contain `history` attribute (only 'id', 'name', 'subject' etc). It's just a fake object with all fields cloned into it. Thus the migration fails due to missing attribute 'history'.
> 
> Strangely, if using `from django.apps import apps as django_apps` and `EmailTemplate = apps.get_model('core', 'EmailTemplate')` then it'll return `squad.core.models.EmailTemplate`, which is the correct one.

when running from tje django shell, this is the expected behavior: to return the "current" version of the model. Now, in the migration, it needs to return the model as it existed at the point where the migration was created, and that's why it's "fake" one: it is one that has the attributes and  relationships that the model had at that point.

not sure how to workaroudn this. a quick skim of the django-simple-history documentation doesn't seem to expect that bulk creation method being run from migrations. sorry, no ideas :-/

chaws: Thanks for the input. I've came up with a workaround. It's not perfect, but I guess it solves our problem here. I looked it over simple-history's code and saw the internals of `bulk_history_create` and managed to replicate it on this migration. It basically creates one HistoricalEmailTemplate for each EmailTemplate. How does it sound to you?

PR #380, opened 2018-10-26
core: fix incorrect notification template names

mwasilew:


mwasilew:
I guess we should have a test for that but I wanted a quick fix to stop annoying error messages.

PR #381, opened 2018-10-26
Fix flake8 errors.

stevanradakovic:
Unused variables in catch blocks removed.


stevanradakovic:
Dropping since it was fixed by @mwasilew in #380

PR #388, opened 2018-11-01
Clean up is_pipeline checks.

stevanradakovic:
is_pipeline field is long removed from TestJob object.
Remove obsolete checks in the code.


mwasilew:
Do you mean it's removed from LAVA jobs? If so I guess we need to add a paragraph to the documentation stating that only LAVA v2 is supported at the moment.

stevanradakovic:
Yes, I've added the doc bit. Thanks!

PR #391, opened 2018-11-07
 Vanishing TestJobs upon changing version name.

amrohassaan:
Testjob doesn't depend on version literal anymore, rather the build id.
Build column removed from ci_testjob and all functionality uses target_build_id now.
Closes#255.


amrohassaan:
@chaws yes Charles, and I don't know if what i did is wrong or not, but I simply rebased master onto my dev branch..maybe i did that wrong and should have done in other way.

chaws:
hmm, do you remember which params you passed to `git rebase`? I usually do `git rebase -i master`

amrohassaan:
@chaws i didn't pass any params, just while on master did git rebase mydevbranch. Probably this is why it is screwed up. I will try to fix that and resubmit then.

mwasilew:
@amrohassaan I think you rebased the other way around. Branch should be rebased on top of master.

For the code itself, did you try to check how performance is affected? This is best done in the rest api as it generates a lot of queries. Example:
api/testjobs/?build=12379d17
vs
api/testjobs/?target_build__id=126

This can already be done as both fields are populated. So comparison before/after would be useful. I don't expect any change, but you never know. Also please check how long does it take to load 'test jobs' tab in the build page (before and after your patch).

terceiro:
hi, this PR has too many commits. can you please rebase against the current master?

amrohassaan:
@mwasilew I noticed just a second worth of a delay when trying both api queries you suggested. But also worth mentioning, one was done on my machine and the other was on production (qa-reports.l.o) so comparison may not be fair.
Anyways, i will cancel this PR and create a new one

PR #392, opened 2018-11-08
test: remove need to install karma globally

terceiro:


chaws:
@terceiro , did you notice any changes on your package-lock.json? I removed node_modules, then re-ran `npm install` and my package-lock.json got changed (optional:true added to a few packages). Other than that, looks good to me!

stevanradakovic:
We had this issue previously too. I think it's some sort of strange bug in npm related to: https://github.com/npm/npm/issues/17722 
My proposition is that we ignore it for now.

PR #394, opened 2018-11-12
Vanishing TestJobs upon changing version name.

amrohassaan:
Testjob doesn't depend on version literal anymore, rather the build id.
Build column removed from ci_testjob and all functionality uses
target_build_id now.

Closes#255.


squad/frontend/ci.py
```diff
@@ -16,7 +16,7 @@ def testjobs(request, group_slug, project_slug, build_version):
         raise Http404()
     testjobs = TestJob.objects.filter(
         target=project,
-        build=build.version
+        target_build=build.id
```
terceiro: this will work due because the Django ORM will take this, but it's inconsistent with all other similar changes. use just  `build` insead of `build.id` (unless I am missing something)

amrohassaan:
@mwasilew , @terceiro I've rebased that PR one more time on top of master as Milosz asked yesterday on irc.

PR #396, opened 2018-11-13
test: run javascript tests as a separate step

terceiro:
This avoids having the karma output mixed together with the dots ('.')
from the python tests.


squad/manage.py
```diff
@@ -14,8 +14,23 @@ def main():
         execute_from_command_line(sys.argv)
     finally:
         if testing:
-            import test.performance
-            test.performance.export('tmp/queries.json')
+            tests = [t for t in sys.argv[2:] if t.startswith('test.')]
+            if not tests:
+                # only run when not running specific tests
+                rc = performance_tests()
+                rc = rc or javascript_tests()
```
chaws: `rc = performance_tests()` will always be `None`, therefore `rc = rc or javascript_tests()` will always invoke `javascript_tests()`. I noticed you added two new return statements in `performance.diff`. Perhaps the intention here is to get those return codes?

terceiro: yes, good catch. I didn't pay enough attention and failed to make `performance_tests()` return an exit code. I have amended that, and also replaced `or` with `+` so that it crashes if we get None back instead of a number.

PR #397, opened 2018-11-15
api/ci.py: remove snippet that is always false

chaws:
If program passes line 22, backend will not be None at least on line 28,
therefore `backend is None` will always be false.


terceiro:
yes, good catch. in this case I think you also want to remove the assignment `backend = None` in line 20?

PR #399, opened 2018-11-19
fix testjobs resubmit

chaws:
There were two ng-app's in testjobs.jinja2, moved `ResubmitController`
to build.js. We should make one app per page and then share controllers
among ng-apps.

NOTE: on testjobs.jinja2 I just added one indentation level, but it looks like it changed lots of things


mwasilew:
LGTM

PR #403, opened 2018-11-19
core: cleanup_old_builds: leave builds from other projects alone

terceiro:


chaws:
Looks got to me, but one thing I wonder is why all builds were deleted but the one which the project's `data_retention_days` was set to 1 in staging?

from irc:
```
<mwasilew> terceiro: no, no data in any project in staging
<terceiro> huh
<mwasilew> except for the one that I set the retention do 1 day
```

Maybe the project with `data_retention_days = 1` had a build that was 1 day old?

mwasilew:
@chaws yes, that was the case. It updates itself from cron every 3h. So there should be around 8 - 9 builds always:https://staging-qa-reports.linaro.org/people/data-retention-test/ (depending on the timing of the clenup task).

PR #404, opened 2018-11-20
Metric threshold support.

### LINKED ISSUE #296: Metric threshold support
It should be possible to set the threshold for given metric. Exceeding the threshold should result in email notification similar to pass/fail notification we currently have. In other words, exceeded threshold should be considered a regression.

stevanradakovic:
Add threshold support in frontend and charts.
Implement notification support depending on the thresholds.

Fixes #296.


squad/api/rest.py
```diff
@@ -833,6 +833,23 @@ class AnnotationViewSet(viewsets.ModelViewSet):
...
+    class Meta:
+        model = MetricThreshold
+        fields = '__all__'
+
+
+class MetricThresholdViewSet(viewsets.ModelViewSet):
+
+    queryset = MetricThreshold.objects.all()
+    serializer_class = MetricThresholdSerializer
+    filter_fields = ('name', 'value', 'is_higher_better', 'project')
```
mwasilew: Please add MetricThresholdFilter and add filter_class here. There is at least one related field here (project) that can be traversed.

squad/api/rest.py
```diff
@@ -833,6 +833,23 @@ class AnnotationViewSet(viewsets.ModelViewSet):
     ordering_fields = ('id', 'build')
 
 
+class MetricThresholdSerializer(serializers.ModelSerializer):
```
mwasilew: Not a requirement but maybe using HyperlinkedModelSerializer is better? It's surely more consistent with the other serializers we have. If there is a good reason not to use it, I'm OK with that. If you decide to move to hyperlinked one, please add 'id' field manually.

squad/core/models.py
```diff
@@ -853,6 +853,25 @@ def get_regressions(self):
     def get_fixes(self):
         return self.__get_yaml_field__(self.fixes)
 
+    def get_exceeded_thresholds(self):
+        # Return a list of all (threshold, metric) objects for those
+        # thresholds that were exceeded by corresponding metrics.
+        thresholds_exceeded = []
+        for threshold in self.build.project.metricthreshold_set.all():
```
mwasilew: I have a feeling this for loop can be replaced with annotated queryset. How is the performance here with big data sets?

squad/core/notification.py
```diff
@@ -203,6 +244,12 @@ def send_status_notification(status, project=None):
 
     notification.send()
 
+    threshold_metrics = status.get_exceeded_thresholds()
```
mwasilew: IMHO threshold notifications should be send in the same message as regression/fix notifications. It's the same type of event just in context of measurement data. Is there a reason why the notification is separate?

squad/core/models.py
stevanradakovic: I optimized it a bit. I was banging my head a log with how to do some sort of join here with prefetch_related but no dice. Please review again and tell me what you think.

squad/frontend/templates/squad/_threshold_table.jinja2
```diff
@@ -0,0 +1,69 @@
...
+            <div class="col-md-10">
+              <input name="name" ng-model="currentThreshold.name" />
+            </div>
+          </div>
+          <div class="row form-group">
+            <label for="value" class="col-md-3 control-label">Value:</label>
+            <div class="col-md-10">
+              <input name="value" ng-model="currentThreshold.value" />
+            </div>
+          </div>
```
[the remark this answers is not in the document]
stevanradakovic: We search through both test runs and suites to check for metrics which are exceeding thresholds. There's no need to specify anything in the thresholds themselved.
Fixed the higher is better issue.

squad/api/rest.py
```diff
@@ -121,6 +121,14 @@ class Meta:
                   'result': ['exact', 'in']}
 
 
+class MetricThresholdFilter(filters.FilterSet):
+    project = filters.RelatedFilter(ProjectFilter, name="project", queryset=Project.objects.all(), widget=forms.TextInput)
+
+    class Meta:
+        model = MetricThreshold
+        fields = {'name': ['exact', 'in', 'startswith', 'contains', 'icontains']}
```
mwasilew: value field with 'exact', 'lt', 'lte', 'gt', 'gte' operators also sounds useful.

squad/core/models.py
```diff
@@ -853,6 +864,27 @@ def get_regressions(self):
...
+            test_runs = self.build.test_runs.all()
+            suites = Suite.objects.filter(test__test_run__build=self.build)
+            for metric in Metric.objects.filter(
+                    Q(test_run__in=test_runs) | Q(suite__in=suites),
+                    name__in=self.build.project.metricthreshold_set.values_list('name', flat=True)):
+                for threshold in self.build.project.metricthreshold_set.all():
+                    if threshold.is_higher_better:
+                        if metric.result < threshold.value:
+                            thresholds_exceeded.append((threshold, metric))
+                            continue
```
mwasilew: is continue necessary here?

squad/frontend/templates/squad/_threshold_table.jinja2
```diff
@@ -0,0 +1,95 @@
...
+      <form class="well" id="threshold_form" ng-submit="updateMetricThreshold()">
+        <div class="modal-body">
+          {% raw %}
+          <p ng-repeat="(name, errs) in errors" class="alert alert-danger"><strong>{{ name }}</strong>: {{ errs.join(', ') }}</p>
+          {% endraw %}
+
+          <div class="row form-group">
+            <label for="name" class="col-md-3 control-label">Metric name:</label>
+            <div class="col-md-10">
+              <input name="name" ng-model="currentThreshold.name" />
```
mwasilew: maybe we should use select2 with connected search? There may be a lot of metrics available. Names of metrics can be queried from SuiteMetadata object like we do in single test comparison page. Can be fixed later though

mwasilew:
Apart from the label background it looks OK to me. Other fixes can be done later. 

Here is the picture with label covering chart points:
threshold_label

Points are still interactive so this is not a problem. Other option might be moving threshold label behind the chart line (z-index). Not sure if this is possible though.

squad/frontend/static/squad/charts.js
```diff
@@ -44,6 +44,30 @@ app.factory('ChartPanel', function() {
...
+            if (threshold.name == this.metric) {
+                annotations.push({
+                    type: "line",
+                    mode: "horizontal",
+                    scaleID: "y-axis-0",
+                    value: threshold.value,
+                    borderColor: "black",
+                    borderWidth: 2,
+                    label: {
+                        backgroundColor: "#a94442",
```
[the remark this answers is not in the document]
stevanradakovic: Added transparency. Z-index is unfortunately not supported in the chartjs annotation lib. I've updated the same thing for regular annotations, it made sense to do it there as well.

PR #409, opened 2018-11-22
Long running reports

mwasilew:
This is implementation of background reports for SQUAD. It's still missing proper implementation of callbacks hence it's only submitted as WIP. Please check if there is anything obviously wrong. I tested polling and email notifications. Docs are still missing.


mwasilew:
I added callback and ran some tests locally. The remaining bits are: tests for callback and docs.

mwasilew:
Added docs and fixed tests for callback. I think it's ready for review.

doc/api.rst
```diff
@@ -210,6 +210,39 @@ following additional routes:
...
+
+  - output - mime type to be generated. Defaults to "text/plain". Can also be set
+    to "text/html". Using HTML requires HTML part of the EmailTemplate to be defined
+
+  - template - ID of the EmailTemplate to be used
+  - baseline - ID of the Build object to be used as comparison baseline. The default
+    is "previous finished" build in the same project.
+  - email_recipient - email address which is notified when report is ready
+  - callback - URL which SQUAD calls when report is ready. Call is made using POST
+    request type. Call can be secures with token
```
chaws: s/secures/secured/

squad/api/rest.py
```diff
@@ -503,53 +550,66 @@ def email(self, request, pk=None):
...
                 previous_build = Build.objects.get(pk=baseline_id)
                 baseline = previous_build.status
             except Build.DoesNotExist:
                 data = {
-                    "message": "Build %s does not exist" % baseline_id
+                    "message": "Baseline build %s does not exist" % baseline_id
                 }
-                return Response(data, status=status.HTTP_500_INTERNAL_SERVER_ERROR)
+                # return created=False to avoid running prepare_report
+                return save_delayed_report(delayed_report, data, status.HTTP_500_INTERNAL_SERVER_ERROR), False
```
chaws: in squad/ci/tasks.py you used the following snippet before `save_delayed_report`:
```
if delayed_report.template is not None:
            data.update({
                "txt": delayed_report.template.plain_text,
                "html": delayed_report.template.html
            })
```

is the intention here not to have those extra args for `data`?

mwasilew: In this line text and html part are not filled in yet. Since baseline doesn't exist it's not possible to create that content. But you're right it's a bit inconsistent.

squad/api/rest.py
```diff
@@ -503,53 +550,66 @@ def email(self, request, pk=None):
...
+        if created or force:
+            delayed_report = prepare_report(delayed_report.pk)
+        if delayed_report.status_code != status.HTTP_200_OK:
+            return Response(yaml.load(delayed_report.error_message), delayed_report.status_code)
+        if delayed_report.output_html:
+            return HttpResponse(delayed_report.output_html, content_type=delayed_report.output_format)
+        return HttpResponse(delayed_report.output_text, content_type=delayed_report.output_format)
+
+    @detail_route(methods=['get', 'post'], suffix='report', permission_classes=[AllowAny])
+    def report(self, request, pk=None):
```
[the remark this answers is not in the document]
mwasilew: no. DelayedReport is created in line 605. If you call the same URL immediately after it will trigger the notifications. Honestly I'm not sure why I did that. It doesn't look right. Notifications should only be sent when the report is ready. So the whole 'else' block should be removed (I think).

In line 607 the DelayedReport is sent for processing. Before that the status_code field is empty. After prepare_report finishes, status_code should be filled in with some value. This way client knows when the report is ready (I forgot to add this line to the docs).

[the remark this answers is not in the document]
chaws: I think the else block could be removed, or maybe just return DelayedReport status, because the notifications are already sent after report finished processing. What do you think?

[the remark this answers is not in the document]
mwasilew: You're right. I'll make changes

squad/api/rest.py
```diff
@@ -480,19 +521,25 @@ def testjobs(self, request, pk=None):
...
-        """
+    def __return_delayed_report(self, request):
         output_format = request.query_params.get("output", "text/plain")
         template_id = request.query_params.get("template", None)
         baseline_id = request.query_params.get("baseline", None)
+        email_recipient = request.query_params.get("email_recipient", None)
+        callback = request.query_params.get("callback", None)
+        callback_token = request.query_params.get("callback_token", None)
+        # keep the cached reports for 1 day by default
+        data_retention_days = request.query_params.get("keep", 1)
```
stevanradakovic: I don't see the 'keep' parameter described in the API docs. Is this intentional?

mwasilew: no, mistake. Thanks for reporting. I will add it

squad/core/tasks/__init__.py
```diff
@@ -262,6 +279,67 @@ def get_suite_version(test_run, suite):
         return None
 
 
+def save_delayed_report(delayed_report, error_message, status_code):
```
stevanradakovic: The method name is a bit confusing here. It updates two specific field in the report, so probably go in that direction.

PR #412, opened 2018-11-27
api: add schema for REST API clients

mwasilew:
In order to support standard clients SQUAD offers an API schema that is
used to describe the interface endpoints.

### DIFF
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -287,6 +287,23 @@ patchsources (/api/patchsources/)
 
 Provides access to PatchSource object.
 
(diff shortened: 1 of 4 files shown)


chaws:
I know it's not supposed to be read by humans, but what do you think of having a link to '/api/schema' in the /api index page so people know squad has it?

mwasilew:
added. However the link can only be relative as the text comes from docstring. There is no access to settings so I can't generate the full URL.

chaws:
@mwasilew I think just the relative path is fine. LKTM

PR #413, opened 2018-11-29
Add support for marking points as outliers.

### LINKED ISSUE #312: metrics: add support for marking points as outliers
Each data point is its own object, so we can add a boolean field to it called `outlier`, and then exclude outliers (or plot them outside of the line) in the charts.

Also needs the UI for users to mark points as outliers.

stevanradakovic:
Add is_outlier field to Metric object.
Show outliers in separate color in charts.
Support adding/removing is_outlier flag from the charts via menu.

Fixes #312


stevanradakovic:
WIP. Missing chart UI control for marking point as outlier.

squad/frontend/views.py
```diff
@@ -482,6 +482,24 @@ def thresholds(request, group_slug, project_slug):
     return render(request, 'squad/thresholds.jinja2', context)
 
 
+@auth
+def toggle_outlier_metric(request, group_slug, project_slug, metric_id):
```
mwasilew: I would use REST API but that can be changed later. I guess the bigger issue is that everyone can toggle the outlier flag. We probably need to address the permissions sooner than later.

mwasilew:
When changing the outliers the chart gets rendered differently before and after. Examples below:

1. mark outlier
outlier1
2. unmark outlier
outlier2
3. reload the page
outlier3

IMHO this is minor issue and I can live with that (use case is pretty dumb). However if it's easy to fix, please do so.

mwasilew:
One more thing is that 'toggle outlier' menu doesn't work in full screen. Can this be changed?

squad/frontend/views.py
stevanradakovic: I was thinking of adding auth_write decorator here. And wrt UI, do you think that we shouldn't show the 'toggle outlier' option at all if the user hasn't got the write permissions for the corresponding project? Alternative is to show the error when he tries to perform the operation but I like the first solution better.

stevanradakovic:
The issue with rendering happens only when the points happen to be in the same exact moment in time. Otherwise it works ok. Unfortunately this cannot be fixed unless we return the whole dataset from the server every time the outlier is toggled which is a bad trade-off, IMHO.

stevanradakovic:
Fix for the full screen and some other issue I found(outlier points 'jumping' to other metrics) is in.

PR #414, opened 2018-11-30
ci/backend/lava.py: add option to ignore lava suite

### LINKED ISSUE #174: register 'lava' test suite as usual tests
Imported from https://projects.linaro.org/browse/QA-1722
Create switch in Project object that tells whether to register 'lava' results as ordinary tests or consider them as 'meta-results'.

chaws:
Added a condition that checks whether `backend_settings` or `project_settings` set the option `CI_LAVA_IGNORE_TESTSUITE` in order to skip parsing results from lava suites. `project_settings` has higher priority than `backend_settings`. Fixes #174


squad/ci/backend/lava.py
```diff
@@ -271,16 +271,21 @@ def __get_publisher_event_socket__(self):
...
-        if self.settings is not None:
-            lava_infra_error_messages = self.settings.get('CI_LAVA_INFRA_ERROR_MESSAGES', [])
-            for message_re in lava_infra_error_messages:
-                try:
-                    r = re.compile(message_re, re.I)
-                    infra_messages_re_list.append(r)
-                except re.error:
-                    # ignore incorrect expressions
-                    self.log_debug("'%s' is not a valid regex" % message_re)
+        for message_re in self.settings.get('CI_LAVA_INFRA_ERROR_MESSAGES', []):
```
mwasilew: hmm, you will get AttributeError if self.settings are None or empty string.

squad/ci/backend/lava.py
```diff
@@ -314,6 +319,9 @@ def __parse_results__(self, data, test_job):
                         res_value = result['measurement']
                         metrics.update({res_name: float(res_value)})
                 else:
+                    if ignore_lava_suite:
```
mwasilew: IMHO this is wrong. The idea in #174 is to treat 'lava' as any other suite. Meaning that block 'if' above should be applied to all suites. So the proper solution should be also to apply 'ignore_lava_suite' to the if block above:
```
if ignore_lava_suite or result['suite'] != 'lava'
```

In this context variable name is misleading.

test/ci/backend/test_lava.py
```diff
@@ -309,6 +309,62 @@ def test_parse_results_metadata_with_suite_versions(self, get_results, get_detai
...
+        self.backend.backend_settings = 'CI_LAVA_IGNORE_TESTSUITE: true'
+        lava = self.backend.get_implementation()
+        testjob = TestJob(
+            job_id='1234',
+            backend=self.backend)
+        status, completed, metadata, results, metrics, logs = lava.fetch(testjob)
+
+        self.assertEqual(lava.settings.get('CI_LAVA_IGNORE_TESTSUITE'), True)
+        self.assertEqual(len(results), 1)
+        self.assertEqual(len(metrics), 1)
```
mwasilew: this is wrong, should be 2.

test/ci/backend/test_lava.py
```diff
@@ -309,6 +309,62 @@ def test_parse_results_metadata_with_suite_versions(self, get_results, get_detai
...
+        lava = self.backend.get_implementation()
+        testjob = TestJob(
+            job_id='1234',
+            backend=self.backend,
+            target=self.project)
+        status, completed, metadata, results, metrics, logs = lava.fetch(testjob)
+
+        self.assertEqual(lava.settings.get('CI_LAVA_IGNORE_TESTSUITE'), None)
+        self.assertEqual(len(results), 1)
+        self.assertEqual(len(metrics), 1)
```
mwasilew: ditto.

squad/ci/backend/lava.py
chaws: I changed the initial value from `self.setting` from `None` to `{}` so it wouldn't crash: https://github.com/chaws/squad/blob/add-project-setting/squad/ci/backend/null.py#L20

squad/ci/backend/lava.py
chaws: Oh I missed that point! It definitely makes more sense. I'll rename it to something like 'handle_lava_suite'.

test/ci/backend/test_lava.py
chaws: got it, missed the point above

PR #415, opened 2018-11-30
core: create ProjectStatus together with Build

terceiro:


mwasilew:
LGTM but needs to be updated wrt to migration dependencies

PR #418, opened 2018-12-10
Make api/data returns all results if the metrics is not specified.

stevanradakovic:
If the metrics is not specified, we should send the data for all metrics.


mwasilew:
Let's merge it for now. However I'm a bit concerned with amount of data we might return in one call. This should at least be paginated. So IMHO the proper change should be adding /metrics endpoint to REST API.

PR #424, opened 2018-12-14
api: add 'force' parameter to /build/<ID>/report

mwasilew:
This patch adds 'force' option to 'report' API. Force has the same
effect as in case of 'email' api - it invalidates the cached object and
forces re-calculation.

### DIFF
--- a/test/api/test_rest.py
+++ b/test/api/test_rest.py
@@ -220,13 +220,31 @@ class RestApiTest(TestCase):
     def test_build_report_retry(self, prepare_report_mock):
         response = self.client.get('/api/builds/%d/report/' % self.build3.id)
         self.assertEqual(202, response.status_code)
+        self.client.get('/api/builds/%d/report/' % self.build3.id)
         report_object = self.build3.delayed_reports.last()
         self.assertTrue(response.json()['url'].endswith(reverse('delayedreport-detail', args=[report_object.pk])))
         self.assertIsNotNone(report_object)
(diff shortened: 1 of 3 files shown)


test/api/test_rest.py
```diff
@@ -204,9 +204,25 @@ def test_build_report_retry(self, prepare_report_mock):
...
+
+    @patch('squad.core.tasks.prepare_report.delay')
+    def test_build_report_retry_force(self, prepare_report_mock):
+        response = self.client.get('/api/builds/%d/report/' % self.build3.id)
+        self.assertEqual(202, response.status_code)
+        report_object = self.build3.delayed_reports.last()
+        self.assertTrue(response.json()['url'].endswith(reverse('delayedreport-detail', args=[report_object.pk])))
+        self.assertIsNotNone(report_object)
+        self.assertIsNone(report_object.status_code)
+        prepare_report_mock.assert_called_once()
```
chaws: IMHO, checking if `prepare_report` was called once should've been tested for at least two requests to `/api/builds/##/report/`. Nothing crucial, but it would make more sense to me. 

Later down the test, I think having the `reset_mock` removed and then a check for something like `self.assertEqual(prepare_report-mock.call_count, 2)` after a request with `force=true` would fit better.

mwasilew: I guess reset_mock doesn't make any harm and works same way as call_count. I fixed the 1st comment.

PR #435, opened 2018-12-20
Report api fixes

### LINKED ISSUE #433: email and report API not cached when baseline is specified
This may be intentional, but I was surprised.

When requesting a report or an email, the result is normally cached and served from cache subsequently. However, if a baseline is specified, then a new email/report is generated each time.

For example, observe the 'url' returned with the following two requests when submitted multiple times:
- https://qa-reports.linaro.org/api/builds/12017/report/?template=9&baseline=11972
- https://qa-reports.linaro.org/api/builds/12017/report/?template=9

mwasilew:
Fixes #433


test/api/test_rest.py
```diff
@@ -218,6 +218,35 @@ def test_build_report_baseline(self, prepare_report_mock):
...
+
+    @patch('squad.core.tasks.prepare_report.delay')
+    def test_build_report_invalid_baseline(self, prepare_report_mock):
+        response = self.client.get('/api/builds/%d/report/?baseline=123456789' % (self.build3.id))
+        self.assertEqual(202, response.status_code)
+        report_object = self.build3.delayed_reports.last()
+        self.assertTrue(response.json()['url'].endswith(reverse('delayedreport-detail', args=[report_object.pk])))
+        report_json = self.client.get(response.json()['url'])
+        self.assertIsNone(report_json.json()['baseline'])
+        self.assertIsNotNone(report_object)
```
chaws: same as above

PR #438, opened 2018-12-27
doc: correct errata in curl example

sangorrin:
Signed-off-by: Daniel Sangorrin <daniel.sangorrin@toshiba.co.jp>


terceiro:
thanks

PR #439, opened 2019-01-02
core/notification.py: turn off escaping for text emails

chaws:
By default Django/Jinja2 escapes some caracters, e.g. ' and & to html
entities. That troubles subject content and text emails with escaped characters.

Now the behavior works as follows:
1. if template comes from string: do not escape
2. if template comes from a file and filename endswith ```.txt.jinja2```: do not escape
3. else: escape

### DIFF
--- a/squad/core/notification.py
+++ b/squad/core/notification.py
@@ -107,6 +107,7 @@ class Notification(object):
             template = custom_email_template.subject
         else:
             template = '{{project}}: {{tests_total}} tests, {{tests_fail}} failed, {{tests_pass}} passed (build {{build}})'
+
         return jinja2.from_string(template).render(subject_data)
 
     def message(self, do_html=True, custom_email_template=None):
--- a/squad/core/templatetags/squad_notification.py
+++ b/squad/core/templatetags/squad_notification.py
@@ -1,4 +1,3 @@
-from squad.core.utils import format_metadata
 from squad.jinja2 import register_global_function, register_filter
 
 
@@ -37,7 +36,5 @@ def tabulate_test_comparison(comparison, test_results=None):
 
 @register_filter
 def metadata_txt(v, key=None):
-    separator = " "
-    if key:
-        separator = "\n" + " " * (len(key) + 2)
-    return format_metadata(v, separator)
+    separator = ("\n" + " " * (len(key) + 2)) if key else " "
+    return separator.join(v) if type(v) is list else v
--- a/squad/jinja2.py
+++ b/squad/jinja2.py
@@ -1,5 +1,5 @@
 from importlib import import_module
-from jinja2 import Environment
+from jinja2 import Environment, select_autoescape
 from django.contrib.staticfiles.storage import staticfiles_storage
 from django.utils import translation
 
@@ -38,6 +38,10 @@ def environment(**options):
     env.filters.update(_local_env['filters'])
     env.tests.update(_local_env['tests'])
     env.install_gettext_translations(translation)
+    env.autoescape = select_autoescape(
+        disabled_extensions=('txt.jinja2',),
+        default_for_string=False,
+        default=True)
     return env
 
 
--- a/test/core/test_notification.py
+++ b/test/core/test_notification.py
@@ -178,6 +178,19 @@ class TestSendNotification(TestCase):
         msg = mail.outbox[0]
         self.assertEqual(0, len(msg.alternatives))
 
+    def test_dont_html_escape_metadata_for_plain_text(self):
+        self.project.subscriptions.create(email='foo@example.com')
+        self.project.html_mail = False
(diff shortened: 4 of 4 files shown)


terceiro:
When using the default template, I tried adding some `'` and `&` to metadata of a test run, and still get the values escaped (note keys `foo` and `bar):

```
datetime: 2018-09-03T21:09:34-03:00
job_id: 1.env1
job_url: http://example.com/1/1
resubmit_url: http://example.com/1/1/resubmit
foo: bar&#39;baz
bar: b&amp;b
xxx: yyy
```

mwasilew:
I think we're looking in the wrong place. I checked again in the original message and the &#39; stands for single quote (or apostrophe). It shows up in all keys except for "git commit". Now checking original test jobs it turns out that "git commit" is the only key that actually uses single quotes in the definition:
https://validation.linaro.org/scheduler/job/1901607/definition
Maybe we get incorrect values from LAVA?

chaws:
@terceiro how did you test it? Using a file template or a string template? 

@mwasilew yes, the simplest thing would be changing the EmailTemplates directly in the UI view. This PR would prevent things like that happening again for text templates and for strings

terceiro:
@chaws this test reproduces the issue that I found:

```
diff --git a/test/core/test_notification.py b/test/core/test_notification.py
index 7ea4db7..f57a99b 100644
--- a/test/core/test_notification.py
+++ b/test/core/test_notification.py
@@ -152,6 +152,19 @@ class TestSendNotification(TestCase):
         msg = mail.outbox[0]
         self.assertEqual(0, len(msg.alternatives))
 
+    def test_dont_html_escape_metadata_for_plain_text(self):
+        self.project.subscriptions.create(email='foo@example.com')+        self.project.html_mail = False+        self.project.save()+        env = self.project.environments.create(slug='myenv')
+        self.build2.test_runs.create(environment=env, metadata_file='{"foo": "foo\'bar"}')
+
+        status = ProjectStatus.create_or_update(self.build2)
+        send_status_notification(status)
+
+        msg = mail.outbox[0]
+        self.assertIn("foo'bar", msg.body)
+
 
 class TestModeratedNotifications(TestCase):
 
```

currently it fails like this (note `foo&#39;bar`):

```
$ ./manage.py test test.core.test_notification
Creating test database for alias 'default'...
System check identified no issues (0 silenced).
......F..............
======================================================================
FAIL: test_dont_html_escape_metadata_for_plain_text (test.core.test_notification.TestSendNotification)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/terceiro/src/linaro/squad/test/core/test_notification.py", line 166, in test_dont_html_escape_metadata_for_plain_text
    self.assertIn("foo'bar", msg.body)
AssertionError: "foo'bar" not found in 'Tests:   0\nFailed:  0\nPassed:  0\nSkipped: 0\nBuild:   2\nDetails: https://lemur/mygroup/myproject/build/2\n\nfoo: foo&#39;bar\n\n\nRegressions (compared to build 1)\n------------------------------------------------------------------------\n\n(none)\n\n\nFailures\n------------------------------------------------------------------------\n\n(none)\n\n\nThreshold(s) exceeded\n---------------------\n\n\nAll changes (compared to build 1)\n------------------------------------------------------------------------\n\n(none)\n\n\n--\nSQUAD\nhttps://lemur'

----------------------------------------------------------------------
Ran 21 tests in 0.854s

FAILED (failures=1)
Destroying test database for alias 'default'...
```
```

chaws:
@terceiro ,  it turns out that you forced html to be false so the [notification txt template](https://github.com/Linaro/squad/blob/master/squad/core/templates/squad/notification/diff.txt.jinja2) would be rendered. It does in fact avoid escaping (as the sample below shows), but the template itself is escaping things out by using `metadata_txt` filter, which in turn calls [format_metadata](https://github.com/Linaro/squad/blob/master/squad/core/utils.py#L51).

I tested out the behavior with the settings in this PR and they look like this:

    $ cat squad/frontend/templates/squad/lala.txt.jinja2 
    txt {{ lala }}
    $ cat squad/frontend/templates/squad/lala.html.jinja2 
    html {{ lala }}
    $ ./manage.py shell
    In [1]: from django.template.loader import render_to_string
    In [2]: print(render_to_string('squad/lala.txt.jinja2', context={"lala": "oba'oba"}))
    txt oba'oba
    In [3]: print(render_to_string('squad/lala.html.jinja2', context={"lala": "oba'oba"}))
    html oba&#39;oba

terceiro:
@chaws ok, makes sense. however AIUI, this is the situation we were running in when this issue was originally reported

chaws:
@terceiro I fixed that `metadata_txt` filter in the text template for notification. I think it's good to go now

PR #440, opened 2019-01-02
ci/backend/lava.py: fallback project_settings

chaws:
Python yaml returns a None type object when the input string is empty


mwasilew:
Looks good. Please add an unit test to cover both empty and non empty cases.

chaws:
I've added a case to cover empty project_settings, the non empty one is already there

PR #444, opened 2019-01-08
Make notification strategy a per-recipient choice.

stevanradakovic:
Move notification strategy from project based to recipient based.
Use db migration to populate all notification strategies based on
the previous per-project values.
Add a new option for notification strategy - only_on_regression.


stevanradakovic:
I've added the strategy to the user settings UI. I was forced to remove the multiple select for project subscriptions due to this change though, just a heads up.

PR #449, opened 2019-01-14
Error output missing from failed test cases in tests view.

stevanradakovic:
Add 'show info' message for failed test cases in test view.

Fixes LSS-369

### DIFF
--- a/squad/frontend/templates/squad/tests.jinja2
+++ b/squad/frontend/templates/squad/tests.jinja2
@@ -32,7 +32,11 @@
             <tr id='test-{{test.name|slugify}}' ng-show="match('test-{{test.name|slugify}}')">
                 <td><a href="{{url('test_history', args=[build.project.group.slug, build.project.slug, test.name])}}">{{test.name}}</a></td>
                 {% for status in test %}
-                <td class="{{status}}"><a href="{{url('test_history', args=[build.project.group.slug, build.project.slug, test.name])}}">{{status}}</a></td>
+                <td class="{{status[0]}}"><a href="{{url('test_history', args=[build.project.group.slug, build.project.slug, test.name])}}">{{status[0]}}</a>
+                  {% if status[1] %}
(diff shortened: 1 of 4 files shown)


squad/frontend/templates/squad/tests.jinja2
```diff
@@ -32,7 +32,11 @@
             <tr id='test-{{test.name|slugify}}' ng-show="match('test-{{test.name|slugify}}')">
                 <td><a href="{{url('test_history', args=[build.project.group.slug, build.project.slug, test.name])}}">{{test.name}}</a></td>
                 {% for status in test %}
-                <td class="{{status}}"><a href="{{url('test_history', args=[build.project.group.slug, build.project.slug, test.name])}}">{{status}}</a></td>
+                <td class="{{status[0]}}"><a href="{{url('test_history', args=[build.project.group.slug, build.project.slug, test.name])}}">{{status[0]}}</a>
+                  {% if status[1] %}
+                      <a href='#' class='pull-right' data-toggle="modal" data-target="#info_modal" data-status="{{ status[1] }}"><span class='fa fa-info-circle'></span> Show info</a>
```
terceiro: I would leave only the icon, and set  "Show info" as a tooltip. if you have many columns, several "show info"s will be a great distraction; see for example the amount of columns in https://qa-reports.linaro.org/lkft/linux-stable-rc-4.14-oe/build/v4.14.93/tests/

squad/frontend/tests.py
```diff
@@ -122,11 +124,22 @@ def get(cls, build, page, search, per_page=50):
         memo = {}
         for test in tests:
             memo.setdefault(test.full_name, {})
-            memo[test.full_name][test.test_run.environment_id] = test.status
+            memo[test.full_name][test.test_run.environment_id] = [test.status]
```
terceiro: I think the code will be a lot more clear if we use a dict with proper names for the fields instead of an array. at some point decoding the [0], [1], [2] and [3] everywhere can get confusing

PR #450, opened 2019-01-16
squad/_threshold_table.jinja2: restrict admin controls

chaws:
Threshold add/edit/del were visible to non-logged users


stevanradakovic:
Looks great, thanks @chaws

terceiro:
(not saying we already have this kind of check available though)

chaws:
@terceiro good idea, there's `writable_by` in the projects model, so it's better use that way: https://github.com/Linaro/squad/pull/458

Issue #432, opened 2018-12-18
Fixes UI incorrect


danrue:
kselftest/bpf_test_verifier_log is not a 'fix' in v4.19.10-45-g75803dcafdb3, as can be seen here: https://qa-reports.linaro.org/lkft/linux-stable-rc-4.19-oe/tests/kselftest/bpf_test_verifier_log?top=v4.19.10-45-g75803dcafdb3

Yet, it (and many others) are shown as fixes over and over again in https://qa-reports.linaro.org/lkft/linux-stable-rc-4.19-oe/ when hovering over the green check box icon.

![image](https://user-images.githubusercontent.com/409011/50182044-e54e9c00-02d3-11e9-955b-78a22748f6ca.png)

I believe this behavior was introduced in 0.56.

stevanradakovic:
@danrue Hm I fail to see 'kselftest/bpf_test_verifier_log' in the attached image or in the UI when I hover. I only see 'kselftest/bpf_test_verifier' but this one has failed in multiple occasions so it shows up correctly afais. Can you please clarify?

PR #452, opened 2019-01-17
Add management command to update project statues from scratch.

stevanradakovic:
A bug #432 was introduced which broke fixes and regressions db field
entries. This management command updates the project statuses.
It assumes that the code fix is in the code base.

### DIFF
--- a/squad/core/management/commands/update_project_statuses.py
+++ b/squad/core/management/commands/update_project_statuses.py
@@ -1,0 +1,45 @@
+import argparse
+
+from datetime import timedelta, datetime
+from django.core.management.base import BaseCommand
+from django.utils import timezone
+
+from squad.core.models import ProjectStatus, Build
+
+
+def valid_date(date):
+    try:
+        return datetime.strptime(date, "%Y-%m-%d")
+    except ValueError:
+        msg = "'{0}' is not a valid date.".format(
+            date)
+        raise argparse.ArgumentTypeError(msg)
+
+
+class Command(BaseCommand):
+
+    help = """Update project status records in order to fix a bug which compared current builds with the wrong ones."""
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            '--date-start',
+            dest="date_start",
+            default=(datetime.now() - timedelta(days=180)),
+            type=valid_date,
+            help="Start date for project status updates (default: 6 months before current date, format: YYYY-MM-DD)."
+        )
+
+    def handle(self, *args, **options):
+        self.options = options
+
+        builds = Build.objects.filter(
+            datetime__gt=timezone.make_aware(self.options['date_start']),
+            status__finished=True
+        )
+        total = builds.count()
+        for index, build in enumerate(builds):
+            ProjectStatus.create_or_update(build)
+            print('\rProgress: {1:>2}%[{0:10}]'.format(
+                '#' * int((index + 1) * 10 / total),
+                int((index + 1) * 100 / total)), end='')
+        print('\n')
--- a/test/core/test_update_project_statuses.py
+++ b/test/core/test_update_project_statuses.py
@@ -1,0 +1,46 @@
+from datetime import datetime
+
+from django.core.management import call_command
+from django.test import TestCase
+from django.utils import timezone
+
+from squad.core.models import Group, ProjectStatus
+
+
+class UpdateStatusesTest(TestCase):
+
+    def setUp(self):
+        self.group = Group.objects.create(slug='mygroup')
+        self.project = self.group.projects.create(slug='myproject')
+        self.environment = self.project.environments.create(slug='theenvironment')
+        self.suite = self.project.suites.create(slug='/')
+
+    def create_build(self, v, datetime=None, create_test_run=True):
+        build = self.project.builds.create(version=v, datetime=datetime)
+        if create_test_run:
+            build.test_runs.create(environment=self.environment)
+        return build
+
+    def test_update_status(self):
+        build1 = self.create_build('1')
+        build1.datetime = timezone.make_aware(datetime(2018, 6, 1))
+        build1.save()
+
+        status1 = ProjectStatus.objects.first()
+        status1.fixes = "fixes1"
+        status1.finished = True
+        status1.save()
+
+        self.create_build('2')
+        status2 = ProjectStatus.objects.last()
+        status2.fixes = "fixes2"
+        status2.finished = True
+        status2.save()
+
+        call_command('update_project_statuses', "--date-start", "2018-07-01")
+
+        status1.refresh_from_db()
+        status2.refresh_from_db()
+
+        self.assertEqual(status1.fixes, "fixes1")
+        self.assertNotEqual(status2.fixes, "fixes2")


squad/core/management/commands/update_project_statuses.py
```diff
@@ -0,0 +1,30 @@
...
+class Command(BaseCommand):
+
+    help = """Update project status records in order to fix a bug which compared current builds with the wrong ones."""
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            '--date-start',
+            dest="date_start",
+            default=(timezone.now() - timedelta(days=180)),
+            type=date_parser.isoparse,
```
mwasilew: I'm getting the following error here:
```
Traceback (most recent call last):
  File "./manage.py", line 5, in <module>
    main()
  File "/home/milosz/linaro/tcstorm/squad/squad/manage.py", line 14, in main
    execute_from_command_line(sys.argv)
  File "/home/milosz/linaro/tcstorm/squad_venv/lib/python3.5/site-packages/django/core/management/__init__.py", line 364, in execute_from_command_line
    utility.execute()
  File "/home/milosz/linaro/tcstorm/squad_venv/lib/python3.5/site-packages/django/core/management/__init__.py", line 348, in execute
    self.fetch_command(options.args[0]).print_help(self.prog_name, options.args[0])
  File "/home/milosz/linaro/tcstorm/squad_venv/lib/python3.5/site-packages/django/core/management/base.py", line 263, in print_help
    parser = self.create_parser(prog_name, subcommand)
  File "/home/milosz/linaro/tcstorm/squad_venv/lib/python3.5/site-packages/django/core/management/base.py", line 249, in create_parser
    self.add_arguments(parser)
  File "/home/milosz/linaro/tcstorm/squad/squad/core/management/commands/update_project_statuses.py", line 18, in add_arguments
    type=date_parser.isoparse,
AttributeError: module 'dateutil.parser' has no attribute 'isoparse'
```

This is on python 3.5.2. Isn't dateutil 3rd party package? If so we're missing a dependency.

squad/core/management/commands/update_project_statuses.py
```diff
@@ -0,0 +1,30 @@
...
+            help="Start date for project status updates (default: 6 months before current date, format: YYYY-MM-DD)."
+        )
+
+    def handle(self, *args, **options):
+        self.options = options
+
+        builds = Build.objects.filter(
+            datetime__gt=timezone.make_aware(self.options['date_start']),
```
mwasilew: I guess date_start should be checked for existence. Alternatively we can set some default value (beginning of epoch?) to update all ProjectStatus objects. Without the parameter I'm getting this:
```
Traceback (most recent call last):
  File "./manage.py", line 5, in <module>
    main()
  File "/home/milosz/linaro/tcstorm/squad/squad/manage.py", line 14, in main
    execute_from_command_line(sys.argv)
  File "/home/milosz/linaro/tcstorm/squad_venv/lib/python3.5/site-packages/django/core/management/__init__.py", line 364, in execute_from_command_line
    utility.execute()
  File "/home/milosz/linaro/tcstorm/squad_venv/lib/python3.5/site-packages/django/core/management/__init__.py", line 356, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/home/milosz/linaro/tcstorm/squad_venv/lib/python3.5/site-packages/django/core/management/base.py", line 283, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/home/milosz/linaro/tcstorm/squad_venv/lib/python3.5/site-packages/django/core/management/base.py", line 330, in execute
    output = self.handle(*args, **options)
  File "/home/milosz/linaro/tcstorm/squad/squad/core/management/commands/update_project_statuses.py", line 26, in handle
    datetime__gt=timezone.make_aware(self.options['date_start']),
  File "/home/milosz/linaro/tcstorm/squad_venv/lib/python3.5/site-packages/django/utils/timezone.py", line 285, in make_aware
    return timezone.localize(value, is_dst=is_dst)
  File "/home/milosz/linaro/tcstorm/squad_venv/lib/python3.5/site-packages/pytz/__init__.py", line 237, in localize
    raise ValueError('Not naive datetime (tzinfo is already set)')
ValueError: Not naive datetime (tzinfo is already set)
```

squad/core/management/commands/update_project_statuses.py
```diff
@@ -0,0 +1,30 @@
...
+            status__finished=True
+        )
+        for build in builds:
+            ProjectStatus.create_or_update(build)
```
mwasilew: It would be nice to show some progress to the user.

squad/core/management/commands/update_project_statuses.py
mwasilew: Also I'm missing the explanation about the date format. I tried YYYY.MM.DD and it didn't work. YYYY-MM-DD works.

mwasilew:
I tried this yesterday and the mgmt command was taking a lot of time. However I requested to update project statuses for ~6 months worth of data. Maybe we should also add option to select a project? This should speed up the whole process.

stevanradakovic:
> I tried this yesterday and the mgmt command was taking a lot of time. However I requested to update project statuses for ~6 months worth of data. Maybe we should also add option to select a project? This should speed up the whole process.

@mwasilew Hm I'm confused the default date start value is 6 months. The main point of the command is to be run once and fix all the wrong fixes and regression values in the db (the date serves to be set approximately to when the bug was introduced). Some sort of progress bar sounds nice but I don't think we should go further from that.

mwasilew:
@stevanradakovic my db is older than the bug. So I don't currently have any entries that were caused by the bug. I wanted to check how it behaves so I set the start date to some time in the past.

squad/core/management/commands/update_project_statuses.py
stevanradakovic: > Also I'm missing the explanation about the date format. I tried YYYY.MM.DD and it didn't work. YYYY-MM-DD works.

There's already a format in the help text, where else do you think should we should add the description?

mwasilew: That works. Is there a way to display help text on input parsing error?

stevanradakovic: It already does, I think the only problem here is the missing dependency.

stevanradakovic: I'm not sure what's going on here. There's already python_dateutil in requirements.txt and mention of it in test/ci/test_models.py. Strange that it doesn't work for you, can you check your dateutil version with "pip show python_dateutil" ?

terceiro: yes but we need to know what is the minimum version of dateutil that has what we need

stevanradakovic: @terceiro so should I put the minimum version to 2.7.0 in requirements.txt (and the dev one)?

terceiro: @stevanradakovic now I'm not sure. the rest of squad works fine with an older dateutil, maybe we don't want to enforce such requirement since dateutil 2.7 is relatively recent

stevanradakovic: Ok that sounds fair. I can find some other way to do command line argument validation.

PR #456, opened 2019-01-21
Refactor ng apps

### LINKED ISSUE #401: Refactor ng apps
Stevan did a PR merging two ng-apps into a single one in build.js: https://github.com/Linaro/squad/pull/398

Then Terceiro suggested we should have a single ng-app per page, sharing controllers among each other

chaws:
Closes #401 

Now each controller resides on its own file under `squad/frontend/static/squad/controllers` that can be imported by apps. Using `import/export` requires javascript to run in ES6 mode, this uses `'use strict'` directive by default, which changed the scope of some variables. Also tests now test controllers instead of apps.

All ajax calls now are made with angular. There were a few of jQuery's `$.ajax` calls spread across the code, which were replaced by `$http`.

`package-lock.json` changed a quite a lot due to update of `karma`.


squad/frontend/templates/squad/lala.html.jinja2
```diff
@@ -0,0 +1 @@
+lala {{ lala }}
```
mwasilew: is this a debug?

squad/frontend/templates/squad/lala.txt.jinja2
```diff
@@ -0,0 +1 @@
+lala {{ lala }}
```
mwasilew: ditto

squad/frontend/templates/squad/lala.html.jinja2
chaws: definitelly a debug file

squad/frontend/templates/squad/lala.txt.jinja2
chaws: just pushed an update

mwasilew:
Let's merge it and see if staging still works.

PR #458, opened 2019-01-22
squad/_threshold_table.jinja2: give more specific restriction

chaws:
Instead of just checking if user is logged in, now checks if user can make changes to that build's project.

PS: it's a better version of https://github.com/Linaro/squad/pull/450


mwasilew:
LGTM. @stevanradakovic any objections?

PR #463, opened 2019-01-23
travis: fix detection of Javascript test failures

terceiro:


scripts/travis
```diff
@@ -19,4 +19,4 @@ if [ "$TRAVIS_PULL_REQUEST" = "false" ] && [ "$TRAVIS_BRANCH" = "master" ]; then
   ./scripts/dogfood "${SUBMIT_TO}" qa/squad "$TESTENV" "$TRAVIS_BUILD_NUMBER"
 fi
 
-! grep -q '^FAILED ' .tests
+exit "$rc"
```
chaws: if line 8 is successful, then `rc` will never exist, then this line will become `exit ""`, which in turn will break `exit`. I think removing the double quotes would prevent that from happening.

terceiro: yes, I forgot to initialize `rc` to `0`; updated

PR #466, opened 2019-01-30
models.EmailTemplate: add jinja2 syntax validation

### LINKED ISSUE #176: validate syntax and semantics of custom email templates
Imported from https://projects.linaro.org/browse/QA-1714
Currently custom email templates can't be easily validated beforehand. This story is to allow validating templates at least syntactically (unclosed jinja2 tags etc.). It would be great if the templates can be validated on the existing data, but it's not mandatory here. The feature can be available from admin or user interface whichever is more convenient (in terms of writing squad code).

chaws:
Add validator to EmailTemplate's `subject`, `plain_text` and `html` that checks for syntax errors.

Closes #176


terceiro:
LGTM, but you are missing a migration. even though there are no structural changes to the data, the next time someone else tries to `makemigrations`, these changes would show up so they should be included in this PR.

chaws:
Good catch. It's also worth to note that the historical_email_template also appeared in the migration. I just pushed an update.

PR #469, opened 2019-01-31
docker-compose.yaml: add initial version

terceiro:
It starts 6 services: one for each of squad daemons (web, worker,
listener, scheduler), one for rabbitmq, and one one-off that runs the
migrations.

It uses sqlite by default, and /home is shared across the squad
containers so that they can all access the database.

As preparation, several changes had to be made in order to improve packaging
and simplifying things in general.


mwasilew:
unittest.mock and mock have different methods. So if you replace mock with unittest.mock some tests need to be updated.

terceiro:
argh, right. on python3.7 unittest.mock is already in feature-parity with the external mock package, so I didn't catch this earlier. I modified the code a littie bit, making it use unittest.mock on >= 3.7, and mock otherwise.

just pushed the update, tested locally against both 3.5 and 3.7.  let's see if the travis build runs fine.

terceiro:
during further testing, I noticed this setup can't send emails. I am researching what would be the best solution

docker-compose.yaml
```diff
@@ -0,0 +1,45 @@
...
+
+services:
+  rabbitmq:
+    image: rabbitmq
+
+  dbmigrate:
+    <<: *common_config
+    image: squad
+    command: squad-admin migrate
+    deploy:
```
chaws: I'm getting a warning on this line:

    WARNING: Some services (dbmigrate) use the 'deploy' key, which will be ignored. Compose does not support 'deploy' configuration - use `docker stack deploy` to deploy to a swarm.

terceiro: yes; that part is specifically for `docker stack deploy`, which I also tested

docker-compose.yaml
```diff
@@ -0,0 +1,45 @@
...
+volumes:
+  home:
+
+services:
+  rabbitmq:
+    image: rabbitmq
+
+  dbmigrate:
+    <<: *common_config
+    image: squad
```
mwasilew: Is it mandatory to build image before running 'docker-compose up'? Without building the image I got the error: 
```
Pulling dbmigrate (squad:)...
ERROR: The image for the service you're trying to recreate has been removed. If you continue, volume data could be lost. Consider backing up your data before continuing.

Continue with the new image? [yN]y
Pulling dbmigrate (squad:)...
ERROR: pull access denied for squad, repository does not exist or may require 'docker login'
```
I built the image locally and tagged it 'squad'. This did the trick, but I didn't find it described anywhere. Maybe we should replace 'image: squad' with
```
build:
  context: ./
```
This will build image from the local sources.

mwasilew: And my image turns out to be "bad":
```
django.template.library.InvalidTemplateLibrary: Invalid template library specified. ImportError raised when trying to load 'squad.ci.templatetags.lava': No module named 'ruamel'
```

docker-compose.yaml
```diff
@@ -0,0 +1,45 @@
...
+  volumes:
+    - home:/home
+  environment:
+    - SQUAD_CELERY_BROKER_URL=amqp://rabbitmq
+    - XDG_DATA_HOME=/home
+
+volumes:
+  home:
+
+services:
```
mwasilew: It would be helpful to add "container_name:" to each service so they're easier to debug/identify when running

terceiro: doesn't docker-compose already give nice names to the containers? for me it gives:

```
NAMES
squad_rabbitmq_1
squad_listener_1
squad_web_1
squad_worker_1
squad_scheduler_1
```

mwasilew: In my case all of the containers are named "squad".

docker-compose.yaml
terceiro: added the `build:` entry, and it works for me. docker-compose said:

```
WARNING: Image for service dbmigrate was built because it did not already exist.
To rebuild this image you must use `docker-compose build` or `docker-compose up --build`.
```

docker-compose.yaml
mwasilew: scratch that. I looked at wrong column. Container names are correct

PR #474, opened 2019-02-07
reduce RAM usage when processing new test results

terceiro:
These commits reduce a lot the RAM consumption when receiving new test results.
The main culprit is the TestComparison class, which used to load every single
test object to RAM at once. That just does not work with hundreds of thousands
of tests.


terceiro:
just updated this pull request with a better version of the memory optimization. this new version does not make the process slower; on the contrary, the process gets faster.

PR #476, opened 2019-02-11
frontend: regressions and fixes need to be in copy-pastable

stevanradakovic:
Leave popover hanging so user can select and copy-paste the
text within. The popover will close on next click on the
regression/fixes button.

### DIFF
--- a/squad/frontend/templates/squad/_regressions_and_fixes.jinja2
+++ b/squad/frontend/templates/squad/_regressions_and_fixes.jinja2
@@ -1,7 +1,7 @@
 {% with regressions=build.status.get_regressions() %}
 {% if regressions %}
-<span>
-    <i class="fa fa-exclamation-triangle text-danger popover-hover popover-regressions"></i>
+<span data-toggle="tooltip" title="Click to display regressions">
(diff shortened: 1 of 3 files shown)


squad/frontend/templates/squad/_regressions_and_fixes.jinja2
```diff
@@ -19,7 +19,7 @@
 {% with fixes=build.status.get_fixes() %}
 {% if fixes %}
 <span>
-    <i class="fa fa-check-circle text-success popover-hover popover-fixes"></i>
+    <button class="fa fa-check-circle text-success popover-regressions-fixes" data-toggle="modal" data-target="#regressions_fixes_modal"></button>
```
terceiro: why is this different than the one above?

mwasilew:
I found 2 issues:
 - there can be multiple popovers displayed at the same time
 - closing popover with "X" makes the next click on regressions/fixes icon do nothing. I have to click twice to get the popover opened.

terceiro:
nice! I have checked the issues @mwasilew pointed out and the latest version of the PR fixes them

PR #479, opened 2019-02-12
Add query result limit to api/data

### LINKED ISSUE #462: Add query result limit to api/data
Api calls from Google data studio time out in most cases because of large datasets. We should add support to limit the amount of data retrieved from api/data.

stevanradakovic:
Limit amount of data in api/data API call by date range.

Fixes #462

### DIFF
--- a/squad/api/data.py
+++ b/squad/api/data.py
@@ -26,6 +26,8 @@ def get(request, group_slug, project_slug):
     project = get_object_or_404(group.projects, slug=project_slug)
 
     metrics = request.GET.getlist('metric')
+    date_start = request.GET.get('date_start', None)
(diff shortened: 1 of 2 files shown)


chaws:
@stevanradakovic I think tests need to be updated. There're a few breaking, also I ran `test_all_metrics` locally and got `{':tests:': {'env1': []}, 'bar/baz': {'env1': []}, 'foo': {'env1': []}}` from my debug prints, I think it was supposed to contain metrics within these arrays

mwasilew:
We'll still fail if there is a lot of data with given time interval. Let's merge it and come up with sth better in the future (like limit the number of data points fetched).

PR #481, opened 2019-02-13
Refactor project status rebuild command.

stevanradakovic:
Include date-end argument.
Setup sequential update of the objects.

### DIFF
--- a/squad/core/management/commands/update_project_statuses.py
+++ b/squad/core/management/commands/update_project_statuses.py
@@ -32,19 +32,32 @@ class Command(BaseCommand):
             type=valid_date,
             help="Start date for project status updates (default: 6 months before current date, format: YYYY-MM-DD)."
         )
+        parser.add_argument(
+            '--date-end',
+            dest="date_end",
+            default=datetime.now(),
+            type=valid_date,
+            help="End date for project status updates (default: today, format: YYYY-MM-DD)."
+        )
 
     def handle(self, *args, **options):
         self.options = options
 
         builds = Build.objects.filter(
-            datetime__gt=timezone.make_aware(self.options['date_start']),
+            datetime__range=(
+                timezone.make_aware(self.options['date_start']),
+                timezone.make_aware(self.options['date_end'])),
             status__finished=True
         )
         total = builds.count()
-        for index, build in enumerate(builds):
-            ProjectStatus.create_or_update(build)
 
+        logger.info("Updating ProjectStatus objects from builds...")
+        logger.info("Total build count in selected date range: %s" % total)
+
+        for index, build in enumerate(builds.iterator()):
+            ProjectStatus.create_or_update(build)
             if index % 100 == 0:
                 logger.info('Progress: {1:>2}%[{0:10}]'.format(
                     '#' * int((index + 1) * 10 / total),
-                    int((index + 1) * 100 / total)), end='')
+                    int((index + 1) * 100 / total)))
+        logger.info('Progress: {1:>2}%[{0:10}]'.format('#' * 10, 100))


squad/core/management/commands/update_project_statuses.py
```diff
@@ -32,19 +33,39 @@ def add_arguments(self, parser):
...
+
+        offset = 0
+        step = 50
+        while offset <= build_count:
+            builds = Build.objects.filter(
+                datetime__range=(
+                    timezone.make_aware(self.options['date_start']),
+                    timezone.make_aware(self.options['date_end'])),
+                status__finished=True
+            )[offset:offset + step]
```
terceiro: you can use `Build.objects.filter(...).iterate()` to do that

squad/core/management/commands/update_project_statuses.py
```diff
@@ -32,19 +33,39 @@ def add_arguments(self, parser):
...
+            for index, build in enumerate(builds):
+                ProjectStatus.create_or_update(build)
+                sys.stdout.write('\rProgress: {0}/{1}'.format(
+                    offset + index + 1, build_count))
+                sys.stdout.flush()
```
terceiro: please don't use sys.stdout as we don't have a way of turning  that off when running tests; use logger instead

terceiro: also, logging at every single object is too much; think of several hundred thousand objects ... and sending \r will usually make logs unreadable.

squad/core/management/commands/update_project_statuses.py
stevanradakovic: Nice, I never used it, great advice!

squad/core/management/commands/update_project_statuses.py
```diff
@@ -32,19 +32,36 @@ def add_arguments(self, parser):
...
         builds = Build.objects.filter(
-            datetime__gt=timezone.make_aware(self.options['date_start']),
+            datetime__range=(
+                timezone.make_aware(self.options['date_start']),
+                timezone.make_aware(self.options['date_end'])),
             status__finished=True
         )
         total = builds.count()
-        for index, build in enumerate(builds):
+        for index, build in enumerate(builds.iterator()):
```
mwasilew: index will always be 1, so there won't be any progress bar. This is what I'm getting:
```
[2019-02-18 11:28:01 +0000] [INFO] Updating ProjectStatus objects from builds...
[2019-02-18 11:28:01 +0000] [INFO] Total build count in selected date range: 99
[2019-02-18 11:28:01 +0000] [INFO] Progress:  1%[          ]
[2019-02-18 11:30:01 +0000] [INFO] Progress: 100%[##########]
```

squad/core/management/commands/update_project_statuses.py
```diff
@@ -32,19 +32,36 @@ def add_arguments(self, parser):
...
+            dest="date_end",
+            default=datetime.now(),
+            type=valid_date,
+            help="End date for project status updates (default: today, format: YYYY-MM-DD)."
+        )
 
     def handle(self, *args, **options):
         self.options = options
 
+        build_count = Build.objects.filter(
```
mwasilew: isn't this redundant? It's the same as variable 'total' below.

squad/core/management/commands/update_project_statuses.py
stevanradakovic: Nah it's because we update status every 100 records and you only have 99 in selected date range. Please try with more and if this still persists please let me know. We have to set the limit to something after all...

squad/core/management/commands/update_project_statuses.py
stevanradakovic: Thanks, fixed.

squad/core/management/commands/update_project_statuses.py
mwasilew: :)

PR #482, opened 2019-02-13
frontend: prevent displaying 'secrets' from LAVA job definition

mwasilew:

### DIFF
--- a/squad/ci/models.py
+++ b/squad/ci/models.py
@@ -1,6 +1,8 @@
 import json
 import logging
 import traceback
+import yaml
+from io import StringIO
 from django.db import models
 from django.utils import timezone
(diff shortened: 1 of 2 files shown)


chaws:
The definitions look way better once yaml is parsed, but I think it'd be useful to let the user know there was a `secrets` block in the definition. Since [documentation](https://validation.linaro.org/static/docs/v2/publishing-artifacts.html?highlight=secrets) says that secrets is a key-value array, I think it's worth  replacing secrets to something like `****`

mwasilew:
I guess this can be done. Not very costly.

mwasilew:
I pushed new version of the patch.

chaws:
LGTM

PR #486, opened 2019-02-16
core: GroupManager.accessible_to: restore and fix project count

terceiro:
We need to go through the accessible projects to be able to count how
many projects in each group are accessible to each user. i.e. if a group
has one public and one private project, an anonymous user should see it
as having only one project.


mwasilew:
LGTM. I'm not sure why syntax checker fails the way it does. Please fix that and this can be merged.

terceiro:
fixed

PR #490, opened 2019-03-01
Add initial UI for self-service usage

terceiro:
This pull request adds the ability to:

- Create new groups (authenticated users only)
- Create users namespaces (groups in the format "~${username}"), to host
  personal projects
- On groups where one is a member:
  - Edit group setttings
  - Delete the group
  - Create new projects
- On projects where one is a member:
  - Edit project settings
  - Delete the project


squad/frontend/urls.py
```diff
@@ -7,8 +7,14 @@
...
 from . import tests
 from . import ci
+from . import group_settings
+from . import project_settings
 from . import user_settings
-from squad.core.models import slug_pattern
+from squad.core.models import slug_pattern, group_slug_pattern
+
+
+group_and_user = (group_slug_pattern, slug_pattern)
```
chaws: I might be missing something, but wouldn't `group_and_project` be a more appropriate name for this?

terceiro: yes! good catch

(updated)

squad/core/models.py
```diff
@@ -29,6 +31,8 @@
 
 slug_pattern = '[a-zA-Z0-9][a-zA-Z0-9_.-]*'
 slug_validator = RegexValidator(regex='^' + slug_pattern)
```
mwasilew: Is this variable still used anywhere?

squad/frontend/group_settings.py
```diff
@@ -30,6 +31,28 @@ def settings(request, group_slug):
...
+
+
+@auth_write
+def new_project(request, group_slug):
+    group = request.group
+    if request.method == "POST":
+        form = NewProjectForm(request.POST, instance=Project(group=group))
+        if form.is_valid():
+            project = form.save()
+            return redirect(reverse('project-settings', args=[group.slug, project.slug]))
```
mwasilew: I think exceptions are not handled properly here. For example attempt to create project with slug that already exists in database results in ugly exception rather than error message in the UI.

squad/frontend/group_settings.py
```diff
@@ -31,6 +32,28 @@ def settings(request, group_slug):
...
+
+
+@login_required
+def new_group(request):
+    if request.method == "POST":
+        form = NewGroupForm(request.POST, instance=Group())
+        if form.is_valid():
+            group = form.save()
+            group.setup_for(request.user)
+            return redirect(reverse('group-settings', args=[group.slug]))
```
mwasilew: Same as with 'new project'. In this case there is no crash but there is no error message either.

squad/frontend/templates/squad/index.jinja2
```diff
@@ -27,4 +27,10 @@
 </div>
 
 
+{% if user.is_authenticated %}
```
mwasilew: Maybe this block should be moved to the top? The button is a bit 'buried' in the page.
Should we allow creating new groups to all authenticated users? This seems a bit dangerous to me.

squad/core/models.py
terceiro: yes, `Project` still uses it

squad/frontend/group_settings.py
```diff
@@ -76,6 +78,65 @@ def form_valid(self, form):
...
 
+class GroupRelationshipForm(forms.ModelForm):
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.fields['group'].widget = forms.HiddenInput()
+        self.fields['group'].disabled = True
+
+
+class GroupMemberForm(GroupRelationshipForm):
```
mwasilew: It would be nice to have alphabetic ordering on users field. Right now it's most likely using pk as ordering field.

mwasilew:
It looks OK. Ordering of users is just a minor issue and I can live with that.

PR #492, opened 2019-03-07
Fix api/data call date arguments.

stevanradakovic:
Previously we passed date_start and date_end as strings to
get_metric_data which was incorrect because make_aware expects
datetime objects and not strings. This patch fixes that.

### DIFF
--- a/squad/api/data.py
+++ b/squad/api/data.py
@@ -1,4 +1,7 @@
 import json
+
+from datetime import datetime
+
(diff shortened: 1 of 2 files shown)


mwasilew:
Did you think about creating unit tests for that code?

terceiro:
hey @mwasilew I thought the same thing, but ended up merging anyway before I noticed your comment. sorry

PR #500, opened 2019-03-28
core: remove duplicate memberships before making them unique

terceiro:
The migration will keep the membership with higher access.


terceiro:
I'm merging myself as this should unbreak stagiing (and void the same issue during the production upgrade).

PR #502, opened 2019-04-01
core: drop obsolete Token model

terceiro:
We have been using the token model from django_rest_framework for a
while now, so the Token model from squad.core is just cruft.


mwasilew:
Do we run any tests on RFT token creation/deletion?

terceiro:
not explicitly, but all of the API tests test DRF tokens indirectly

PR #509, opened 2019-04-12
fix yaml.load default Loader warning

chaws:
PyYAML has deprecated the default `yaml.load`. It recommends explicitly specify
the loader or calling a shortcut function, which is done in this commit.

Source: https://github.com/yaml/pyyaml/wiki/PyYAML-yaml.load(input)-Deprecation

### DIFF
--- a/squad/api/rest.py
+++ b/squad/api/rest.py
@@ -407,9 +407,9 @@ class ProjectStatusSerializer(serializers.HyperlinkedModelSerializer):
     def to_representation(self, instance):
         ret = super().to_representation(instance)
         if instance.regressions is not None:
-            ret['regressions'] = json.dumps(yaml.load(ret['regressions']))
+            ret['regressions'] = json.dumps(instance.get_regressions())
         if instance.fixes is not None:
-            ret['fixes'] = json.dumps(yaml.load(ret['fixes']))
+            ret['fixes'] = json.dumps(instance.get_fixes())
         return ret
 
     class Meta:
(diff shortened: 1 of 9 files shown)


requirements.txt
```diff
@@ -13,7 +13,7 @@ Markdown
 msgpack>=0.5.0
 psycopg2
 python_dateutil
-PyYAML
+PyYAML>=5.1
```
terceiro: this forces an upgrade of PyYAML in the docker image (which should use Debian packages from Debian buster). Also `full_load` is only available in that version. Please don't do this version bump, and replace `load` with `safe_load` instead of `full_load` in the code. This way you both make the PyYAML usage safer in a way that works with both PyYAML 3 and PyYAML 5, and will also fix this warning.

chaws: I replaced most calls for `safe_load`. The only ones left out were ProjectStatus `fixes` and `regressions` because they use python deserializer (due to OrderedDict), which behaves like a threat for `safe_load`, breaking it. Instead of changing `fixes` and `regressions`, I chose to keep their old `load` function, just passing the default `Loader` to it, thus getting rid of the warning. Also I've changed tests a bit to make sure they're loading correctly.

chaws:
@terceiro On a side note, I realized that `.travis.yml` still uses pip to install python packages. During this current PR, I noticed that tests ran in docker failed because they used an older version of pyyaml if compared to non-docker test runs, which passed. Should we tweak `.travis.yml` to stop using pip and force it to use Debian packages instead?

terceiro:
@chaws we still use pip in production so let's not change that for now

PR #510, opened 2019-04-12
Dockerfile: cache installed dependencies

terceiro:
Having the installation of packages after WORKDIR and COPY forces a
reinstallation of the dependencies every time there is any change to the source
tree. By installing the dependencies directly on top of the base image, that
layer will be used from the cache and only needs to be rebuilt if the list of
deopendencies change.

The resulting image has a couple of extra layers, but is the same size as
before:

squad/after   latest   3e6bc1fd3f33  6 seconds ago   280MB
squad/before  latest   e09e3b1ead22  2 minutes ago   280MB

(Also, install iproute2 together with auto-apt-proxy so that the later works properly)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,12 +1,8 @@
 FROM debian:buster-slim
 
-WORKDIR /app
-COPY . ./
-
 RUN apt-get update -q=2 && \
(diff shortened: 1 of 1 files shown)


terceiro:
@fboudra please have a look

PR #511, opened 2019-04-15
i18n improvements

terceiro:


terceiro:
I want to cut a release soon, this is the last bit that I really need in the release, and the changes are mostly trivial so I merged myself

PR #513, opened 2019-04-15
dev-docker: fix to work with current Dockerfile

terceiro:


test/javascript.py
```diff
@@ -7,7 +7,10 @@
 
 
 def javascript_tests():
-    if os.path.exists(karma):
+    if not shutil.which('nodejs'):
```
chaws: Is there a scenario where "node_modules/.bin/karma" exists where "nodejs" doesn't? When installing npm, Debian seems to be installing nodejs too.

chaws:
Maybe we should have javascript tests within the dev-docker? Just a suggestion...I added npm and chromium to apt in dev dockerfile then added [no-sandbox](https://github.com/karma-runner/karma-chrome-launcher/issues/189#issuecomment-440423763) to karma config and it worked fine, it just added a much greater amount of time and space

test/javascript.py
terceiro: the scenario that I hit, which is I had this stuff installed in my host machine, but not in the container created by dev-docker

(I tried to add nodejs etc to it, but chromium crashes weirdly inside docker and I didn't want to investigate that at the moment)

terceiro: (just saw you other comment)

terceiro:
@chaws so that's the trick, --no-sandbox. I will amend

PR #514, opened 2019-04-22
frontend: rewrite build settings with Django forms/views

terceiro:


chaws:
LGTM, soo much less code : )

PR #519, opened 2019-05-02
Show "Projects" tab only if user had specific permission

chaws:
The "Projects" tab allows users to navigate from project settings and creating new projects. If the user does not have such privileges, there will be only one tab pointing to the current page, making no use
for it.


mwasilew:
Makes sense to me. I propose merging it.

stevanradakovic:
Looks good

PR #520, opened 2019-05-03
Subscribe api

mwasilew:
Creates subscribe/unsubscribe endpoints for /api/projects/<id>/. These are used to create/remove email subscriptions.

### DIFF
--- a/squad/api/rest.py
+++ b/squad/api/rest.py
@@ -1,9 +1,12 @@
 import json
 import yaml
 
-from django.core.exceptions import MultipleObjectsReturned
+from django.db.models import Q
+from django.core import exceptions as core_exceptions
+from django.core.validators import validate_email
+from django.contrib.auth.models import User
 from squad.api.filters import ComplexFilterBackend
-from squad.core.models import Annotation, Group, Project, ProjectStatus, Build, TestRun, Environment, Test, Metric, MetricThreshold, EmailTemplate, KnownIssue, PatchSource, Suite, SuiteMetadata, DelayedReport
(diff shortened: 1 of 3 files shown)


squad/api/rest.py
```diff
@@ -401,6 +404,39 @@ def test_results(self, request, pk=None):
...
+    @detail_route(methods=['post'], suffix='unsubscribe')
+    def unsubscribe(self, request, pk=None):
+        subscriber_email = request.data.get("email", None)
+        if subscriber_email is None:
+            raise serializers.ValidationError("'email' field is mandatory.")
+        try:
+            validate_email(subscriber_email)
+        except core_exceptions.ValidationError:
+            raise serializers.ValidationError("Invalid 'email' field value.")
+        subs = Subscription.objects.filter(Q(email=subscriber_email) | Q(user__email=subscriber_email))
```
chaws: Wouldn't this unsubscribe a user/email that could be subscribed in different projects? Perhaps passing project as in the subscribe endpoint would do it.

terceiro: yes, the filter needs to be more strict

squad/api/rest.py
```diff
@@ -401,6 +404,39 @@ def test_results(self, request, pk=None):
...
+        if subs.exists():
+            for sub in subs:
+                sub.delete()
```
terceiro: you don't need to check for emptiness, or to loop over the items. just call `subs.delete()` and it will do the right thing.

squad/api/rest.py
terceiro: you should probably have a test for only unsubscribing from the current project

mwasilew: good point. Will fix it.

squad/api/rest.py
mwasilew: oh, I didn't know :)

chaws:
LGTM, all fixes were made, merging it

PR #524, opened 2019-05-09
Document translation process

terceiro:
Also, rename scripts/new-language to scripts/translate and adapt it to
be used both to create new languages, as well as updating existing ones.


scripts/translate
```diff
@@ -0,0 +1,24 @@
+#!/bin/sh
+
+set -eu
+
+apps='core frontend'
+
+for lang in "$@"; do
```
chaws: If no parameter is given, the script will just exit silently. Maybe it would be nice adding a small check if 'lang' is empty

doc/translating.rst
```diff
@@ -0,0 +1,17 @@
...
+
+    $ ./scripts/translate pt_BR
+    squad/core/locale/pt_BR/LC_MESSAGES/django.po: 0 translated messages, 22 untranslated messages.
+    squad/frontend/locale/pt_BR/LC_MESSAGES/django.po: 0 translated messages, 212 untranslated messages.
+
+The script will output the names of the translation files that need to be
+edited, together with the current translation statistics. Edit the `*.po`
+files, commit, and send a pull request.
+
+This procedure is valid both for creating new translations, and for updating
```
chaws: When running `./scripts/translate`, there's no new messages added to the po files, i.e. if I add a new `_()` or `N_()` to the code. Is that the intention here or should the user run `./scripts/update-translation-files` before that?

scripts/translate
terceiro: nice catch. I added this check, and removed the loop; it will only take one language at a time.

doc/translating.rst
terceiro: `./scripts/update-translation-files` should be run by a maintainer when close to a release, and when no more strings should be added before the release (this is usually called "string freeze").

`./scripts/translate` should be run by each translator.

chaws: didn't know about "string freeze", but grasped it by context :)

scripts/translate
chaws: ack

chaws:
LGTM

PR #526, opened 2019-05-13
frontend/project_settingspy: add enabled plugins list to settings

### LINKED ISSUE #525: Plugins settings missing from web UI for project/group admins
It isn't possible to enable/disable plugins when using the self-service web UI. The only option to change plugin settings for the project is to use admin interface.

chaws:
Closes #525

### DIFF
--- a/squad/frontend/project_settings.py
+++ b/squad/frontend/project_settings.py
@@ -12,7 +12,9 @@ from squad.frontend.queries import get_metrics_list
 class ProjectForm(forms.ModelForm):
     class Meta:
         model = Project
-        fields = ['name', 'is_public', 'is_archived', 'description']
+        fields = ['name', 'is_public', 'is_archived', 'description',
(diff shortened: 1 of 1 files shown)


squad/frontend/project_settings.py
```diff
@@ -12,7 +12,7 @@
 class ProjectForm(forms.ModelForm):
     class Meta:
         model = Project
-        fields = ['name', 'is_public', 'is_archived', 'description']
+        fields = ['name', 'is_public', 'is_archived', 'description', 'enabled_plugins_list']
```
mwasilew: could you also add timeouts?

chaws: "notification timeout" and "wait for notification"? I also thought about "data rentention", what do you think?

mwasilew: makes sense.

chaws: added

PR #527, opened 2019-05-15
Polish translation for SQUAD UI

mwasilew:

### DIFF
--- a/squad/core/locale/pl_PL/LC_MESSAGES/django.po
+++ b/squad/core/locale/pl_PL/LC_MESSAGES/django.po
@@ -1,0 +1,108 @@
+# Translations template for squad.core.
+# Copyright (C) 2019 ORGANIZATION
+# This file is distributed under the same license as the squad.core project.
+# FIRST AUTHOR <EMAIL@ADDRESS>, 2019.
+#
+#, fuzzy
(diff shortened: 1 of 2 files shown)


chaws:
LGTM, but I can't understand a thing of Polish besides common Latin words :)

mwasilew:
I'm pretty sure here are some typos :)

mwasilew:
Some plural forms are incorrect. I'm not sure how to fix it and not affect other languages. For example in the main page there should be "1 project" or "X projects". Same applies to Polish but currently plural form is used everywhere. Moreover depending on the number Polish plural changes. There would be "2 projekty" but "5 projektów". Any ideas how this can be done?

chaws:
In jinja2 there's the `trans` block (https://github.com/Linaro/squad/blob/master/squad/frontend/templates/squad/login.jinja2#L6), but not sure how to do that in plain python

chaws:
@mwasilew just out of curiosity: in Polish do you use the plural form for number 0 (zero) as well? E.g. 0 tests, 1 test, 2 tests...

mwasilew:
@chaws yes, in most cases 0 will be accompanied with plural form (well, one of them).

PR #535, opened 2019-05-20
ci: add support for LAVA multinode jobs

mwasilew:
This patch adds support for multinode jobs. In case of multnode LAVA API
calls return list in the following form: [job_id.0, job_id.1, ...].
SQUAD will use each of the IDs in the list to create new TestJob object.
This way we ensure that results from all test jobs are assigned to
proper build and fetched.

### DIFF
--- a/doc/ci.rst
+++ b/doc/ci.rst
@@ -154,4 +154,24 @@ Example LAVA backend settings:
     CI_LAVA_SEND_ADMIN_EMAIL: False
     CI_LAVA_HANDLE_SUITE: True
 
+Multinode
++++++++++
+
+SQUAD supports fetching results from LAVA multinode jobs. There are however
+a few limitations with this setup:
+ - All results from multinode will share environment name
+   Since test jobs are submitted via SQUAD using the environment from submit
+   URL there is no way for SQUAD to distinguish between different environmens
+   on different parts of multinode job.
+ - Resubmit will repeat the whole set
+   In SQUAD all parts of multinode job will share the multinode definition.
+   For this reason re-submitting any part of the multinode job will result
+   in new multinode job that includes all parts.
+ - Each part of the multinode job will be retrieved separately
+   This means that each part will create a TestRun in SQUAD. This should not
+   be a major issue as all results will still be available. Users need to make
+   sure that the test names don't overlap as SQUAD will not have any means of
+   distinguishing between identically named tests from different parts of
+   multinode job.
+
 .. vim: ts=4 sw=4 et=1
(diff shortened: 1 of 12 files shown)


mwasilew:
I noticed one serious problem with current approach - environment name is copied across all multinode jobs. This might not be desirable but I don't have any good idea how to solve the problem.

squad/ci/models.py
```diff
@@ -134,10 +134,20 @@ def __postprocess_testjob__(self, test_job):
                 logger.error("Plugin postprocessing error: " + str(e) + "\n" + traceback.format_exc())
 
     def submit(self, test_job):
-        test_job.job_id = self.get_implementation().submit(test_job)
+        job_id_list = self.get_implementation().submit(test_job)
+        if isinstance(job_id_list, list):
```
chaws: `self.get_implementation().submit(test_job)` will always return a list of IDs, so this condition will always be true

mwasilew: It should but I don't know how to enforce it. Anyway, point taken, I'll update it.

chaws:
Maybe we could get the environment of each job by calling [scheduler.jobs.configuration](https://validation.linaro.org/api/help/#scheduler.jobs.configuration). But documentation says it'll return 404 if job hasn't started yet. I don't know if queued jobs will return 404 or not

mwasilew:
SQUAD environment doesn't directly map to LAVA device or device type. So that's no use. We could parse it from the job definition directly if it was the case.

terceiro:
the code looks good enough to me

wrt the envionment: a multinode job usually specifies a specific pairing of devices, and I would expect that pairing to the "the environment". I don't think we should try to handle the device types of the lava jobs ... if one wants separate environments, don't use multinode jobs.

chaws:
Also from that issue, the extra node seems to be supporting the main target anyways. IMHO I'd say it's OK to copy the env name.

mwasilew:
I'll add a few words on this limitation to docs.

squad/ci/models.py
mwasilew: It initially looked like one line change but ended with 2 commits and half of the tests updated :)

mwasilew:
Docs pushed. IMHO it's ready now.

doc/ci.rst
```diff
@@ -154,4 +154,24 @@ Example LAVA backend settings:
...
+Multinode
++++++++++
+
+SQUAD supports fetching results from LAVA multinode jobs. There are however
+a few limitations with this setup:
+ - All results from multinode will share environment name
+   Since test jobs are submitted via SQUAD using the environment from submit
+   URL there is no way for SQUAD to distinguish between different environmens
+   on different parts of multinode job.
+ - Resubmitt will repeat the whole set
```
chaws: s/Resubmitt/Resubmit/

mwasilew: thanks, fixed.

Issue #199, opened 2018-01-10
include metrics summaries (aka geomeans) in charts


mwasilew:
Imported from https://projects.linaro.org/browse/QA-1544

Issue #198, opened 2018-01-10
initialize charts with overal summary on all environments


mwasilew:
Imported from https://projects.linaro.org/browse/QA-1545
Currently the charts page starts empty. Instead, we should display by default the overall summary on all environments.

terceiro:
Yet another failure is properly describing what each issue actually means. :-)

"overall summary" is the geomean of all metrics for a given build/environment. You will be able to get that from the `metrics_summary` field in `Status` objects, with `suite=None`. the data set should be given by something like

```
env = ... # the environment you want
project = ... # the project you want
Status.objects.filter(suite=None, test_run__environment=env, test_run__build__project=project)

```

With that said, you could reword the commit message (dropping mention to #198) and the pull request title and we could get this merged right away and change to the overall summary later. after all showing the test pass % by default is better than displaying nothing.

terceiro:
Just remove the issue # from the commit message so the issue doesnt get closed.
 The link from the PR will stay there forever, its ok

chaws:
@terceiro in your comment above, you mentioned that the `metrics_summary` of `Status` when `suite=None` represents the metrics summary (geomean) of all metrics for a given Build/Environment.

I've run this query but I'm still getting multiple `metrics_summary` per Build/Environment. Then it occurred to me that because a single build has multiple `TestRun`s, when `suite=None` it actually represents metrics summary for a single TestRun/Environment, not for a Build/Environment.

And because `ProjectStatus` contains `metric_summary` of a build (of all environments), we still don't have a simple way to get the overall metrics summary per build per environment.

Am I missing anything here?

terceiro:
@chaws yes, I think you got it right

this part was designed very early and ended up a bit outdated with regards to the whole Build/TestRun design. We indeed don't have this data per build and environment

chaws:
This was already solved in #537

PR #537, opened 2019-05-22
Metrics summary per build per environment

chaws:
Closes https://github.com/Linaro/squad/issues/199 then https://github.com/Linaro/squad/issues/198.

See https://github.com/Linaro/squad/issues/198 for context.

This patch adds `BuildSummary` that fills our need to have metrics summary per build per environment. That's what's used to create the overall summary in charts. Also, because issue 198 depends on 199, I've put them both in this PR, separate commits though.

Also it add a the command `compute_build_summaries` that computes data for a specific date, and if a given project is passed, it computes data for builds of a single project as well.


squad/core/models.py
```diff
@@ -1102,6 +1102,33 @@ def __init__(self, build, environment=None):
         self.has_metrics = len(values) > 0
 
 
+class BuildSummary(models.Model):
+    build = models.ForeignKey(Build, related_name='metrics_summary')
+    environment = models.ForeignKey(Environment)
+    metrics_summary = models.FloatField(null=True)
+    has_metrics = models.BooleanField(default=False)
```
terceiro: maybe we also want the test result counts here, instead of only metrics

squad/core/management/commands/compute_build_summaries.py
```diff
@@ -0,0 +1,99 @@
...
+                return
+
+            logger.info('Filtering builds from project "%s"' % (project_name))
+            builds = builds.filter(project=project)
+
+        logger.info('Computing metrics summary for %s builds' % (builds.count()))
+        if show_progress:
+            logger.info('Showing progress, one dot means one processed build')
+
+        if project:
```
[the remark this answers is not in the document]
chaws: Hmm, the thing is that `project` is optional, which in that case the environments needed to be from each `build.project`'s. I managed to shorten it to a single loop by removing the first one completely. The `__get_environments__` already takes care of both cases

terceiro:
@chaws I was thinking that we can solve without storing any extra data. You can calculate the geomean for a given environment by just taking the geomean for each test run in that environment, and taking the geomean of them. e.g.:

```python
In [1]: from squad.core.statistics import geomean

In [2]: x = [1,2,3]

In [3]: y = [4,5,6]

In [4]: geomean(x)
Out[4]: 1.8171205928321397

In [5]: geomean(y)
Out[5]: 4.93242414866094

In [6]: geomean(x+y)
Out[6]: 2.993795165523909

In [7]: geomean([geomean(x), geomean(y)])
Out[7]: 2.993795165523909
```

chaws:
@terceiro I actually tried some math as well, but mine didn't match up like yours due to number of elements. If both `x` and `y` have the same number of elements, then that's true, if number of elements is different, then the equality is not true. See:

```python
In [1]: from squad.core.statistics import geomean

In [2]: x = [1, 2, 3, 4]

In [3]: y = [5, 6, 7]

In [4]: geomean(x)
Out[4]: 2.213363839400643

In [5]: geomean(y)
Out[5]: 5.943921952763129

In [6]: geomean(x + y)
Out[6]: 3.3800151591412964

In [7]: geomean([geomean(x), geomean(y)])
Out[7]: 3.627128604621783
```
If you use greater numbers, the difference grows proportionally. 
The question is: should we trust that the number of tests in all testruns will always be the same?

terceiro:
ack, ignore what I said about that then

chaws:
Yeah, I thought about a similar strategy to combine geomeans and reduce the number of *Summary classes, but couldn't find a way out :(

squad/core/models.py
chaws: I've pushed a new update with tests_pass, tests_fail, tests_xfail, tests_skip. Also `BuildSummary` extends `TestSummary` now. I didn't bother keeping `regressions` and `fixes`. Should I?

PR #538, opened 2019-05-22
Improvements to the import_data utility

terceiro:


squad/core/management/commands/import_data.py
```diff
@@ -75,6 +80,15 @@ def import_build(self, directory):
         for envdir in glob(os.path.join(directory, '*')):
             self.import_environment(build_id, envdir)
 
+        if not self.options['dry_run']:
+            build = self.project.builds.get(version=build_id)
```
chaws: Because now there's a check for `metadata.json`, the build with `version=build_id` may have never been created, causing this line to raise `DoesNotExist`.

terceiro: yes, good catch. fixed

chaws:
LGTM

PR #543, opened 2019-06-04
Add UI to compare 2 builds from the same project

### LINKED ISSUE #191: add UI to compare 2 builds from the same projects
Imported from https://projects.linaro.org/browse/QA-1626
we already have a cross-project comparison, and the same implementation will support comparing 2 builds of the same project just fine. we just need a UI that allows the user to select which 2 builds to compare, like in art-reports:
 * in the build list, allow to select 2 builds, and compare from there
 * the the build page, allow to enter a build number to compare that build with the current one.

chaws:
Closes #191 

Compare build page:
Screenshot from 2019-06-03 14-28-49

Compare build in build list:
Screenshot from 2019-06-03 17-25-06

Compare build in build page:
Screenshot from 2019-06-04 08-46-56

### DIFF
--- a/squad/frontend/comparison.py
+++ b/squad/frontend/comparison.py
@@ -1,4 +1,4 @@
-from django.shortcuts import render
+from django.shortcuts import render, get_object_or_404
 from django.core.paginator import Paginator
 
 from squad.core.models import Project
@@ -35,3 +35,33 @@ def compare_test(request):
 
     context = {}
     return render(request, 'squad/compare.jinja2', context)
+
+
+def compare_builds(request):
+    project_slug = request.GET.get('project')
+    comparison = None
+    project = None
+    if project_slug:
+        group_slug, project_slug = project_slug.split('/')
+        project = get_object_or_404(Project, group__slug=group_slug, slug=project_slug)
+
+        baseline_build = request.GET.get('baseline')
+        target_build = request.GET.get('target')
+        if baseline_build and target_build:
+            baseline = get_object_or_404(project.builds, version=baseline_build)
+            target = get_object_or_404(project.builds, version=target_build)
+            comparison = TestComparison.compare_builds(baseline, target)
+
+            try:
+                page = int(request.GET.get('page', '1'))
+            except ValueError:
+                page = 1
+            paginator = Paginator(tuple(comparison.results.items()), 50)
+            comparison.results = paginator.page(page)
+
+    context = {
+        'project': project,
+        'comparison': comparison,
+    }
+
+    return render(request, 'squad/compare_builds.jinja2', context)
--- a/squad/frontend/static/squad/build.js
+++ b/squad/frontend/static/squad/build.js
@@ -1,13 +1,16 @@
 import {AnnotationController} from './controllers/annotation.js'
+import {BuildCompareController} from './controllers/build_compare.js'
 import {FilterController} from './controllers/filter.js'
 import {ResubmitController} from './controllers/resubmit.js'
 import {Config as appConfig} from './config.js'
+import {attach_select2} from './attach_select2.js'
 
 var app = angular.module('Build', []);
 
-app.value('build', window.build);
-
 appConfig(app, ['httpProvider']);
+
+app.value('build', window.build);
+app.factory('attach_select2', ['$http', attach_select2]);
 
 app.controller(
     'AnnotationController',
@@ -37,5 +40,14 @@ app.controller(
         '$location',
         '$timeout',
         ResubmitController
+    ]
+);
+
+app.controller(
+    'BuildCompareController',
+    [
+        '$scope',
+        'attach_select2',
+        BuildCompareController
     ]
 );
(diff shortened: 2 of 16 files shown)


squad/frontend/comparison.py
```diff
@@ -35,3 +35,36 @@ def compare_test(request):
 
     context = {}
     return render(request, 'squad/compare.jinja2', context)
+
+
+def compare_builds(request):
+    project_slug = request.GET.get('project')
+    user = request.user
+    projects = Project.objects.accessible_to(user).prefetch_related('group')
```
terceiro: this can (and eventually will) be too many projects.

maybe we can start with this, and come up later with an UI element to select a project by typing (parts of) its name.

squad/frontend/templates/squad/_builds_table.jinja2
```diff
@@ -1,16 +1,24 @@
-<div class='highlight-row'>
+<div class='highlight-row' ng-app='Build' ng-controller='BuildComparisonController'>
+<input type="hidden" ng-init="compare_button_label='{{_('Compare builds')}}'" />
+<input type="hidden" ng-init="invalid_number_of_builds_msg='{{_('Two builds need to be selected!')}}'" />
+<input type="hidden" ng-init="invalid_selected_builds_msg='{{_('Selected builds must be different!')}}'" />
+<input type="button" ng-value="compare_button_label" class="btn btn-default" ng-click="compareBuilds('{{statuses.0.build.project.full_name}}')" />
```
terceiro: `<input>` elements require an enclosing `<form>` tag. it may work without it, but it's not valid HTML.

squad/frontend/templates/squad/_builds_table.jinja2
```diff
@@ -1,16 +1,24 @@
...
 {% for status in statuses %}
 {% with build=status.build %}
 <a href="{{build_url(build)}}">
 <div class="row row-bordered build">
+    <div class="col-md-1 col-sm-1">
```
terceiro: this takes too much horizontal space. Add this to the same column as the unfinished/regressions/fixes column, but inside a div that has a fixed width so that the entire row doesn't move around when the second radio button is not visible.

squad/frontend/static/squad/build.js
```diff
@@ -19,6 +20,15 @@ app.controller(
...
 
+app.controller(
+    'BuildComparisonController',
+    [
+        '$scope',
+        '$window',
+        BuildComparisonController
+    ]
+)
+
```
terceiro: this should probably be added to a separate flle. build.js is supposed to be included in pages where a single build is shown; all the rest of the code imported in build.js would not be useful in the build list

squad/frontend/templates/squad/_builds_table.jinja2
```diff
@@ -53,4 +61,9 @@
 </a>
 {% endwith %}
 {% endfor %}
+<input type="button" ng-value="compare_button_label" class="btn btn-default" ng-click="compareBuilds('{{statuses.0.build.project.full_name}}')" />
 </div>
+
+{% block javascript %}
+<script type="module" src='{{static("squad/build.js")}}'></script>
+{% endblock %}
```
terceiro: see my comment above; I think we should not include here all of the code that only applies to a single build

squad/frontend/static/squad/build.js
chaws: I think I'll place it in here `squad/frontend/static/squad/compare.js`

squad/frontend/comparison.py
chaws: I've added a select2 box for project and builds. Now by clicking on it, a few options will show up automagically, then options will be filtered out as you type

squad/frontend/templates/squad/_builds_table.jinja2
chaws: had to remove some space from the last column (with the time), and added the radio buttons inside the unfinished/regressions/fixes columns, each surrounded by a `div.row`, this way it guarantees that the rest of columns wouldn't move. Couldn't find a better way for it to take less horizontal space though

squad/frontend/templates/squad/_builds_table.jinja2
chaws: yeah, you're right. Created a new file called `build_list.js` that should contain all js for build list, then moved that controller in.

terceiro:
I tried this locally, and found a few UI issues:

* when I click the radio buttons in the build list, I'm sent to the build page instead of just selection the radio button
* in the project selection, if I type $group/ (i.e. group name followed by a slash), I would expect to get all projects inside the $group, but instead I get nothing.
* in the compare builds page, all select boxes take up the entire width of the page. That is very unecessary.

chaws:
Good catch!

It was my bad, I ended up experimenting only on chrome, and there those radio buttons worked as expected. Apparently in FF, having a radio within an anchor prioritizes the latter click event. While I put some work in it, I didn't find any better solution other than keeping the radio buttons outside of  `a`. (evt.preventDefault() and evt.stopPropagation just didn't work for this matter)

Regarding the project selection, I've added more filters to the api so now searches using project's full_name are possible.

Lastly, regarding the needless used space, I've cut it in half. Please take a look and let me know if it looks better now. I'm a total wreck in layouts & design :)

PS: I just realized that the api and the user_menu templates are not translated.

PR #544, opened 2019-06-05
LAVA test logs

### LINKED ISSUE #539: How to get case log from squad?
See this: 

Capture111

How to get the failure case log by directly click the `red fail link` in `All test results`?

mwasilew:
Fixes #539

### DIFF
--- a/squad/ci/backend/lava.py
+++ b/squad/ci/backend/lava.py
@@ -8,6 +8,7 @@ import yaml
 import xmlrpc
 import zmq
 
+from io import BytesIO
 from zmq.utils.strtypes import u
 
 from xmlrpc import client as xmlrpclib
@@ -56,12 +57,12 @@ class Backend(BaseBackend):
                 data['results'] = self.__get_testjob_results_yaml__(test_job.job_id)
 
                 # fetch logs
-                logs = ""
+                raw_logs = BytesIO()
                 try:
-                    logs = self.__get_job_logs__(test_job.job_id)
+                    raw_logs = BytesIO(self.__download_full_log__(test_job.job_id))
                 except Exception:
                     self.log_warn(("Logs for job %s are not available" % test_job.job_id) + "\n" + traceback.format_exc())
-                return self.__parse_results__(data, test_job) + (logs,)
+                return self.__parse_results__(data, test_job, raw_logs)
         except xmlrpc.client.ProtocolError as error:
             raise TemporaryFetchIssue(self.url_remove_token(str(error)))
         except xmlrpc.client.Fault as fault:
@@ -205,10 +206,27 @@ class Backend(BaseBackend):
         response = requests.get(url, params=payload)
         return response.content
 
-    def __get_job_logs__(self, job_id):
-        logger.debug("Retrieving logs job: %s" % job_id)
-        log_data = self.__download_full_log__(job_id)
-        return self.__parse_log__(log_data)
+    def __download_test_log__(self, raw_log, log_start, log_end):
+        return_lines = ""
+        if not log_start:
+            return return_lines
+        log_start_line = int(log_start)
+        log_end_line = None
+        if log_end:
+            log_end_line = int(log_end)
+        else:
+            log_end_line = log_start_line + 2  # LAVA sometimes misses the signals
+        raw_log.seek(0)
+        counter = 0
+        for line in raw_log:
+            counter += 1
+            if counter < log_start_line:
+                continue
+            return_lines = return_lines + line.decode("utf-8")
+            if counter >= log_end_line:
+                break
+        raw_log.seek(0)
+        return return_lines
 
     def __parse_log__(self, log_data):
         returned_log = ""
@@ -287,7 +305,7 @@ class Backend(BaseBackend):
             selected_setting = tmp_setting
         return selected_setting
 
-    def __parse_results__(self, data, test_job):
+    def __parse_results__(self, data, test_job, raw_logs):
         if hasattr(test_job, 'target') and test_job.target.project_settings is not None:
             ps = yaml.safe_load(test_job.target.project_settings) or {}
             handle_lava_suite = self.__resolve_setting__(ps, 'CI_LAVA_HANDLE_SUITE', False)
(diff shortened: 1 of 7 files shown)


squad/ci/backend/lava.py
```diff
@@ -314,10 +318,16 @@ def __parse_results__(self, data, test_job):
                 if handle_lava_suite or result['suite'] != 'lava':
                     suite = result['suite'].split("_", 1)[-1]
                     res_name = "%s/%s" % (suite, result['name'])
+                    res_log = None
+                    if 'log_start_line' in result.keys() and \
```
chaws: I think you can do `result.get('log_start_line') is not None`, it's less complex

test/api/tests.py
```diff
@@ -60,6 +61,17 @@ def test_receives_tests_file(self):
...
+
+    def test_receives_tests_file_with_logs(self):
+        with open(tests_log_file) as f:
+            self.client.post(
+                '/api/submit/mygroup/myproject/1.0.0/myenvironment',
+                {'tests': f}
+            )
+        self.assertIsNotNone(models.TestRun.objects.last().tests_file)
+        self.assertNotEqual(0, models.Test.objects.count())
+        self.assertIsNotNone(models.Test.objects.last().log)
```
terceiro: it would be nice to also test that the test result arrives correctly

test/api/tests.py
```diff
@@ -68,6 +80,16 @@ def test_receives_tests_file_as_POST_param(self):
...
+        self.assertIsNone(models.Test.objects.last().log)
+
+    def test_receives_tests_file_as_POST_param_with_logs(self):
+        self.client.post(
+            '/api/submit/mygroup/myproject/1.0.0/myenvironment',
+            {'tests': '{"test1": {"result": "pass", "log": "test log"}}'}
+        )
+        self.assertIsNotNone(models.TestRun.objects.last().tests_file)
+        self.assertNotEqual(0, models.Test.objects.count())
+        self.assertIsNotNone(models.Test.objects.last().log)
```
terceiro: ditto

squad/ci/backend/lava.py
terceiro: IMO `in` is just fine

squad/ci/backend/lava.py
```diff
@@ -314,10 +318,16 @@ def __parse_results__(self, data, test_job):
...
+                            'log_end_line' in result.keys() and \
+                            result['log_start_line'] is not None and \
+                            result['log_end_line'] is not None:
+                        res_log = self.__download_test_log__(test_job.job_id, result['log_start_line'], result['log_end_line'])
```
terceiro: it looks like this will make an HTTP request for every single test result? we should probably download the full log once, and do this separation locally instead

squad/ci/backend/lava.py
mwasilew: I think it's a good point - 2 lines fewer.

squad/ci/backend/lava.py
mwasilew: This will be tricky. Any hint how to retrieve only certain lines from the local file? Lava stores offsets to read from the file.

mwasilew: I mean retrieve only certain lines without loading the whole file to memory.

terceiro: if you `open()` the file, it won't be read into memory as a whole unless you say so. if you have the offsets, you can `seek()` to them; otherwise you can read line by line with `readline()`

mwasilew:
Please ignore last update. I just rebased my branch. I'll do requested changes next week.

mwasilew:
I re-wrote some parts of log fetching. Please review now. A few more unit tests should be added to cover corner cases. I tested locally and it works well.

squad/ci/backend/lava.py
```diff
@@ -205,10 +206,27 @@ def __download_full_log__(self, job_id):
...
+        log_end_line = None
+        if log_end:
+            log_end_line = int(log_end)
+        else:
+            log_end_line = log_start_line + 2  # LAVA sometimes misses the signals
+        raw_log.seek(0)
+        counter = 0
+        for line in raw_log:
+            counter += 1
+            if counter <= log_start_line:
```
chaws: I think this line should be `counter < log_start_line`, otherwise the first log line will always be missing.

squad/ci/backend/lava.py
```diff
@@ -205,10 +206,27 @@ def __download_full_log__(self, job_id):
...
+    def __download_test_log__(self, raw_log, log_start, log_end):
+        return_lines = ""
+        if not log_start:
+            return return_lines
+        log_start_line = int(log_start)
+        log_end_line = None
+        if log_end:
+            log_end_line = int(log_end)
+        else:
+            log_end_line = log_start_line + 2  # LAVA sometimes misses the signals
```
chaws: Do you think it's nice to check whether `log_start_line <= log_end_line`? I was just thinking of it now, but most likely it's not needed.

test/ci/backend/test_lava.py
```diff
@@ -33,6 +35,8 @@
     {'duration': '',
      'job': '1234',
      'level': 'None',
+     'log_end_line': '4',
+     'log_start_line': '5',
```
chaws: `log_start_line` is greater than `log_end_line`

mwasilew: I know. I did that on purpose.

chaws: Oh, nvm then :)

squad/ci/backend/lava.py
mwasilew: I wasn't sure about that but you might be right. 1st line is usually not that important anyway but I'll change it.

squad/ci/backend/lava.py
mwasilew: I didn't see such case in LAVA. End always at least 1 greater than Start. I saw cases where lava missed the start or end by a few lines.

chaws:
@mwasilew thanks for applying the updates :)

I just have once last minor concern about test logs (I promise), e.g.: https://github.com/Linaro/squad/blob/master/squad/frontend/templates/squad/_test_run_test.jinja2#L52

This will display logs as `test.log`, and if that's extracted from LAVA directly, it'll be in YAML syntax. Should we care parsing the test log to contain only messages?

mwasilew:
@chaws  I know and I thought about that. There are a few options:
 - strip timestamps and leave messages only
 - filter messages from target only
 - leave as is
For now I think it's better to leave as is. The reason for that is that start and stop markers aren't always accurate. Leaving in yaml format and not filtering lava debug allows to understand what is happening.

Having said that, if there are complains it's easy to change. So I would leave it as it is and wait for feedback.

chaws:
LGTM

Issue #197, opened 2018-01-10
add dynamic summary to charts


mwasilew:
Imported from https://projects.linaro.org/browse/QA-1546
A dynamic summary is the summary of the selected metrics (as in ART-reports)

chaws:
Like this one?
![Screenshot from 2019-05-02 19-26-01](https://user-images.githubusercontent.com/2254825/57110815-3c4b8f80-6d10-11e9-8973-7f7c793adcb4.png)

mwasilew:
Yes. This data needs to be created in the database first.

terceiro:
note that **dynamic** summary specifically is calculated at runtime as it depends on the selected metrics (in the UI), so in art-reports it was calculated on the fly, and should probably be done in the same way for squad.

the ones that we want to have in the database are the metric summaries; we want one global (for all metrics), and then one per suite (that's  #199)

chaws:
Similar to: https://art-reports.linaro.org/#/stats/?branch=master&environment=bullhead_32_a53_aot&environment=bullhead_32_a57_aot&environment=bullhead_64_a53_aot&environment=bullhead_64_a57_aot&environment=flounder_32_denver_aot&environment=flounder_64_denver_aot&environment=NUC6i5SYH_32_i5-6260U_aot&environment=NUC6i5SYH_64_i5-6260U_aot&environment=sailfish_32_kryo_aot&environment=sailfish_64_kryo_aot&environment=VOYOv1_32_Pentium-N4200_aot&environment=VOYOv1_64_Pentium-N4200_aot&benchmark=:summary:&startDate=1554606000.415&limit=-100

PR #545, opened 2019-06-05
charts: add dynamic metric summary

terceiro:


chaws:
That's interesting because even if two metrics are not related, the geomean normalizes the data.

The only downside is that depending on the data, it may take quite a few seconds to process the dynamic metrics chart, but that's inevitable. 

This closes https://github.com/Linaro/squad/issues/197.

PR #549, opened 2019-06-07
frontend: fix floatThead

chaws:
I noticed that the plugin `floatThead` as in *float table head*, was named wrong in download scripts as `floatThread`. This PR renames floatThread to floatThead and add `thead` where needed.

Hopefully it won't conflict with https://github.com/Linaro/squad/pull/543, but if it does, I'll update this PR.


terceiro:
when you re-run the download script, you will be left with a new tarball under `frontend/static/vendor/`. please include that in the PR (and remove the previous one)

chaws:
Missed that one! Already pushed the new tarball and added a small note in download.conf about that.

PR #556, opened 2019-06-22
Spelling: Delete project

comradekingu:
There is no linking from Weblate onto here.
The only info presented is a link to [the docs](https://hosted.weblate.org/projects/squad/#information), which don't seem to lead here.


terceiro:
@comradekingu thanks for the patch

and you are right, the weblate URL should point here instead; I will do that.

PR #558, opened 2019-06-27
metrics: display range of values

terceiro:


terceiro:
Screenshot_2019-06-27 SQUAD -

squad/frontend/static/squad/controllers/charts.js
```diff
@@ -231,16 +229,50 @@ function ChartPanel($http, DATA) {
...
+                        label: ':low:' + env.name,
+                        data: low,
+                        backgroundColor: env.range_color,
+                        pointRadius: 0,
+                        pointHoverRadius: 0,
+                        borderColor: '#ffffff00', // transparent
+                        yidx: 6,
+                        fill: '+1'
+                    }
+                    )
```
chaws: i think this parenthesis should've been in line above, right after closing curly braces

squad/frontend/static/squad/controllers/charts.js
```diff
@@ -683,6 +725,19 @@ function ChartsController($scope, $http, $location, $compile, ChartPanel, DATA)
         })
     }
 
+    $scope.filterByDate = function(chartPanel, minLimit, maxLimit) {
+        console.log(chartPanel)
```
chaws: leftover debugging...

chaws:
I noticed that there's an if condition to check whether `low != line` and `high != line`, but that only worked for integer values.

In the screenshot below, it shows ':low:env' and ':high:env' displaying same values

Screenshot from 2019-07-01 17-26-37

squad/frontend/static/squad/controllers/charts.js
terceiro: fixed

squad/frontend/static/squad/controllers/charts.js
terceiro: yes, nice catch

terceiro:
> I noticed that there's an if condition to check whether `low != line` and `high != line`, but that only worked for integer values.

actually it doesn't work for arrays at all:

Issue #201, opened 2018-01-10
build comparison (metrics-based)


mwasilew:
Imported from https://projects.linaro.org/browse/QA-1519
similar to what we now have in art-reports, i.e. we want to compare metrics across the projects. metrics comparison also needs to be added to the existing project comparison

chaws:
Just adding up. It should be similar to: https://art-reports.linaro.org/#/builds/compare/?from=3986&to=3985

PR #559, opened 2019-06-28
Metric comparison

chaws:
Add option for metric based comparison for builds/projects. It's using the same template from test comparison. It shows the mean, the standard deviation between parenthesis and the number of measurements when hovering on the table cell.

Screenshot from 2019-06-28 17-21-56

Screenshot from 2019-06-28 17-24-21

### DIFF
--- a/squad/core/comparison.py
+++ b/squad/core/comparison.py
@@ -1,33 +1,33 @@
 from collections import OrderedDict
-from django.db.models import F
+from itertools import groupby
+import statistics
 
 
-from squad.core.utils import parse_name
+from squad.core.utils import parse_name, join_name
 from squad.core import models
-from squad.core.utils import join_name
 
 
-class TestComparison(object):
+class BaseComparison(object):
     """
     Data structure:
 
     builds: [Build]
     environments: Build → [EnvironmentName(str)]
-    results: TestName(str) → ((Build,EnvironmentName(str)) → TestResults)
+    results: Name(str) → ((Build,EnvironmentName(str)) → Results)
 
     The best way to think about this is to think of the table you want as
     result:
 
-    +-------+---------------+---------------+
-    |       |    build 1    |    build 2    |
-    +       +-------+-------+-------+-------+
-    |       | env 1 | env 2 | env 1 | env 2 |
-    +-------+-------+-------+-------+-------+
-    | test1 |       |       |       |       |
-    | test2 |       |       |       |       |
-    | test3 |       |       |       |       |
-    | test4 |       |       |       |       |
-    +-------+-------+-------+-------+-------+
+    +---------+---------------+---------------+
+    |         |    build 1    |    build 2    |
+    +         +-------+-------+-------+-------+
+    |         | env 1 | env 2 | env 1 | env 2 |
+    +---------+-------+-------+-------+-------+
+    | result1 |       |       |       |       |
+    | result2 |       |       |       |       |
+    | result3 |       |       |       |       |
+    | result4 |       |       |       |       |
+    +---------+-------+-------+-------+-------+
 
     `builds` is the list of builds (the top header row)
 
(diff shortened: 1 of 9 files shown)


squad/core/comparison.py
```diff
@@ -107,18 +78,96 @@ def diff(self):
...
         return self.__diff__
 
+
+class MetricComparison(BaseComparison):
+
+    def __get_build__(self, build_id):
+        for build in self.builds:
+            if build.id == build_id:
+                return build
+        return None
```
terceiro: you probably want to use a `dict` here

squad/frontend/comparison.py
```diff
@@ -2,16 +2,22 @@
 from django.core.paginator import Paginator
 
 from squad.core.models import Project
-from squad.core.comparison import TestComparison
+from squad.core.comparison import TestComparison, MetricComparison
+
+
+def __get_comparison_class(comparison_type):
+    return MetricComparison if 'metric' == comparison_type else TestComparison
```
terceiro: I think a regular multi-line if is easier to read

squad/frontend/templates/squad/compare_projects.jinja2
```diff
@@ -29,6 +29,16 @@
...
+  <div class="form-group">
+    <label>{{ _('compare by') }}</label>
+    <div>
+      <label for="test-comparison">{{ _('tests') }}</label>
+      <input id="test-comparison" name="comparison_type" type="radio" value="test" checked="checked"/>
+      &nbsp;{{ _('or') }}&nbsp;
+      <label for="metric-comparison">{{ _('metrics') }}</label>
+      <input id="metric-comparison" name="comparison_type" type="radio" value="metric" />
+    </div>
+  </div>
```
terceiro: this does not work for internationalization; you should always use full sentences instead, e.g.
[ ] compare by tests
[ ] compare by metrics

when you break up the sentences, you assume that all languages have the same flow structure as English, and that is not the case.

chaws: That's true, I didn't consider that into the equation. I've added a group label as "Comparison type", then added the options as you suggested. Looks much better indeed.

chaws:
@terceiro , just pushed new updates with your reviews and with mean and stdev being calculated in python using 'statistics' module. It's supposed to be supported in Python 3+.

squad/frontend/templates/squad/compare_builds.jinja2
```diff
@@ -36,10 +36,20 @@
...
+		  <div class="form-group">
+			<label>{{ _('Comparison type') }}</label>
+			<div>
+			  <input id="test-comparison" name="comparison_type" type="radio" value="test" checked="checked"/>
+			  <label for="test-comparison">{{ _('compare by tests') }}</label>
+			  <br />
+			  <input id="metric-comparison" name="comparison_type" type="radio" value="metric" />
+			  <label for="metric-comparison">{{ _('compare by metrics') }}</label>
+			</div>
+		  </div>
```
terceiro: the indentation here is wrong

squad/frontend/templates/squad/compare_builds.jinja2
```diff
@@ -36,10 +36,20 @@
...
           <div class="form-group">
             <input type="hidden" name="project" value="{{ project.full_name }}" />
             <input onclick="this.form.submit()" value="{{ _('Compare') }}" type="button" class="btn btn-default" />
-            or
+            {{ _('or') }}
```
terceiro: I think we should drop this "or". everybody knows what to do when there is a sequence of buttions

squad/frontend/templates/squad/compare_builds.jinja2
chaws: hm, I my vim got confused with tabs. Fixed that, thanks for spotting that.

squad/frontend/templates/squad/compare_builds.jinja2
chaws: agreed.

chaws:
Nitpicks are always welcome :)

chaws:
Closed #201

PR #563, opened 2019-07-10
core: add management command to create/update auth tokens

mwasilew:


mwasilew:
This was requested for creating/updating tokens when working in docker-compose environments with LAVA.

terceiro:
can we add a "set-token" action to the existing `users` command instead?

terceiro:
or a --token option to the existing add and update actions there

mwasilew:
I think 1st option is better/easier

chaws:
LGTM

Issue #307, opened 2018-07-31
add support for notification of gerrit patches


terceiro:
see #190 for context

PR #564, opened 2019-07-11
Gerrit plugin

chaws:
Closes https://github.com/Linaro/squad/issues/307

* Fix dropdown for the "implementation" field in "Djando admin > patch sources > add patch source"
  * Turns out it wasn't an issue related to deploy, as I previously suspected
* Add Gerrit plugin supporting HTTP and SSH
  * For the SSH part:
  Gerrit only allows SSH commands if it's using keys for authentication
* Document the notification plugins Github and Gerrit


squad/core/migrations/0121_auto_20190711_1829.py
```diff
@@ -0,0 +1,25 @@
+# -*- coding: utf-8 -*-
```
mwasilew: Could you rename the migration so the file name is descriptive

squad/core/migrations/0121_auto_20190711_1829.py
```diff
@@ -0,0 +1,25 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('core', '0120_buildsummary'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='patchsource',
+            name='password',
```
mwasilew: not sure if it makes it any more secure, but maybe you should try sth like this: 
https://djangosnippets.org/snippets/1330/

squad/core/migrations/0121_auto_20190711_1829.py
chaws: definitely forgot that one, Thanks for heads up

squad/core/migrations/0121_auto_20190711_1829.py
chaws: Now password is xor'd with `SECRET_KEY` before being stored into the database. Is not as secure as the link you referenced, but avoids storing plain text.

chaws: Just a note: It was a great idea to obfuscate the password before storing in the database, otherwise it'd be displayed in plain text in https://qa-reports.linaro.org/api/patchsources/. A good thing to do is also obfucate token as well, if added to a patchsource of type github. Or just prevent these fields from be displayed in the API would do it too.

mwasilew: you can make the fields 'write only' in the API. This is a good practice for passwords/tokens.

PR #568, opened 2019-07-16
api: rest: fix rest-framework detail_route

chaws:
Replace "detail_route" decorator by "action(detail=True)". As of django-restframework >= 3.8,  the "detail_route" has been completely removed [1], causing SQUAD travis tests to break.

[1] https://www.django-rest-framework.org/community/3.8-announcement/#deprecations

### DIFF
--- a/squad/api/rest.py
+++ b/squad/api/rest.py
@@ -14,7 +14,7 @@ from django.http import HttpResponse
 from django.urls import reverse
 from django import forms
 from rest_framework import routers, serializers, viewsets, status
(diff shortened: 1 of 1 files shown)


mwasilew:
Thanks @chaws! I was planning on doing this but it was low priority. We don't have great tests for API so some more testing will be required on staging instance.

chaws:
@mwasilew yeah, there's still plenty of room to improve api testing. Perhaps we can leave an open issue for that and merge this PR so the next PRs won't be breaking due to drf update.

Issue #186, opened 2018-01-10
rethink the "compare projects" UI


mwasilew:
Imported from https://projects.linaro.org/browse/QA-1644
original request:

> 13. comparison between projects only inside the team so we don't compare squad to zephyr for example

we should only be able to compare projects inside the same group. also the top-level "compare" link does not make much sense.

milosz.wasilewski:
> In some cases comparison between projects in the group also makes no sense (like in lkft compare CTS to LTP). Maybe it's a better idea to have a list of 'comparable projects'? It's also not easy to guarantee that the compared projects share the same versions, envs and tests.

mwasilew:
This patch: https://github.com/mwasilew/squad/commit/fb7e101fac5fec57e86d0a86a7efed6c0bbd65d1 adds Test object endpoint to API and allows more complex filtering. Adding UI to use these features is the next step.

terceiro:
@mwasilew I don't get the relation between these API endpoints and the "compare projects" UI. maybe you posted your comment on the wrong issue?

mwasilew:
@terceiro there was a discussion (in the email thread) about getting the comparison results through API. Without this endpoint it required hundreds of API calls. With the test endpoint it's 2-3 calls. There is still issue with performance.

The relationship with this issue is that I'm planning to implement the compare UI re-using the API. I'll try to push the changes to the branch so it's clearer what I mean.

chaws:
After https://github.com/Linaro/squad/issues/201, https://github.com/Linaro/squad/issues/191 and https://github.com/Linaro/squad/issues/178 adding more "compare" options, there'll be maybe too many "compare" tabs, we should make dropdown menu with all “compare” options:
- 2+ builds from the same project
- Different test suites from the same build
- etc

chaws:
@mwasilew In https://github.com/Linaro/squad/pull/543 I've added "Compare builds", then I used "Compare" menu to create a dropdown menu with "Compare Builds", "Compare Projects" and "Compare Tests".

Regarding this issue per se, I thought about having on the same page a dropdown box, in which the user can pick a group, then all projects belonging to that group would be displayed next to a select box as we have today. Does this seem reasonable?

mwasilew:
@chaws I think this would make a step in the right direction.

PR #569, opened 2019-07-17
Improve project compare

chaws:
Closes https://github.com/Linaro/squad/issues/186

This PR includes the following improvements:
 - restrict project comparison to projects from the same group only
   - change group by typing its name or slug under "Group" header
 - by selecting a project, a dropdown box appears next to it for build selection
   - by default, the latest build is selected
   - change build by typing its version string

Screenshot from 2019-07-17 19-37-59


terceiro:
LGTM. there is an issue that I will report separately, to not delay this any longer

Issue #554, opened 2019-06-17
TypeError: 'NoneType' object is not callable


keveryang:
This happened after the first time running command './manage.py migrate' and './manage.py addsuperuser':

  Applying sessions.0001_initial... OK
Exception ignored in: <function WeakValueDictionary.__init__.<locals>.remove at 0x7f8eeef8ac80>
Traceback (most recent call last):
  File "/usr/lib/python3.5/weakref.py", line 117, in remove
TypeError: 'NoneType' object is not callable
Exception ignored in: <function WeakValueDictionary.__init__.<locals>.remove at 0x7f8eeef8ac80>
Traceback (most recent call last):
  File "/usr/lib/python3.5/weakref.py", line 117, in remove
TypeError: 'NoneType' object is not callable

chaws:
Hi @keveryang , does that happen every time you run `./manage.py migrate`? How is your set up?

keveryang:
My Host is Debian 9.5(stretch), and here is the packages I have install:
sudo apt install libyaml-dev
sudo apt install python3-pip
sudo pip3 install --upgrade jinja2
sudo pip3 install -r requirements-dev.txt

It does not happen again when I run './manage.py migrate', maybe because all the migrations has been done:
$ ./manage.py migrate
Operations to perform:
  Apply all migrations: admin, auth, authtoken, ci, contenttypes, core, sessions
Running migrations:
  No migrations to apply.

chaws:
This is a bug in Python < 3.6: https://bugs.python.org/issue29519

I've seen that happening other times, specially at the end of tasks, but that didn't mess up with the task itself. I guess we'll have to specify the minimal version of Python to be 3.6+

Thanks for reporting, @keveryang

PR #576, opened 2019-08-08
Migrate to django2

chaws:
This is a work in progress patch that migrates our base framework from Django 1.11 LTS to 2.2 LTS. The changes are described below:

Django 1.11 -> 2.2 [1]
- s/from django.core.urlresolvers import reverse/from django.urls import reverse/
- s/MIDDLEWARE_CLASSES/MIDDLEWARE/ (settings.py)
  - MIDDLEWARE has a new format
- s/django.shortcuts.render_to_response()/django.shortcuts.render()/
- s/User.is_authenticated()/User.is_authenticated/
- s/User.is_anonymous()/User.is_anonymous/
- d/SessionAuthenticationMiddleware/ (settings.py)
- s/@register.assignment_tag/@register.simple_tag/
- s/ForeignKey()/ForeignKey(on_delete=models.CASCADE)/
- s/OneToOne()/OneToOne(on_delete=models.CASCADE)/
- d/Manager.use_for_related_fields/ (squad/core/models.py:539)

Python >= 3.3
- mock is now part of Python3 std library [2]
- 'logger.warn' is deprecated in favor of 'logger.warning' [5]
- should we enforce >= 3.6?
  - weakref bug: https://bugs.python.org/issue29519

Django Rest Framework >= 3.9.2 
- 3.9.2 is ensures support to Django 2.2 [3]
- s/detail_route/action(detail=True)/

Django Filters >= 2.0 [4]
- s/Filter.name/Filter.field_name/

Django Rest Framework Filters >= 1.0.0dev0
- current version doesn't support Django 2.2
- we should wait for 1.0 to be released

References:
[1] https://eldarion.com/blog/2017/12/26/10-tips-upgrading-django-20/
[2] https://docs.python.org/3/library/unittest.mock.html
[3] https://www.django-rest-framework.org/community/release-notes/#392
[4] https://django-filter.readthedocs.io/en/master/guide/migration.html
[5] https://docs.python.org/3/library/logging.html#logging.Logger.warning


terceiro:
FWIW, I tested this on Debian unstable, and after updating a couple of
packages¹² I got it to work just fine, both passing all tests and running the
UI successfully on my machine.

¹ [djangorestframework-filters](https://tracker.debian.org/news/1054466/accepted-djangorestframework-filters-100dev0-1-source-into-unstable/)
² [python-django-extensions](ttps://tracker.debian.org/news/1054468/accepted-python-django-extensions-221-1-source-into-unstable/)

The diff seems reasonable, so overall this looks good to me.

chaws:
Thanks for testing it! 

> ...after updating a couple of packages...

Cool! Did you have any issues with `django_extensions`?

Also, I saw that in Debian buster the default Python3 is 3.7, so I guess it's safe to state that squad is compatible with Python3. I mean that because of the [weakref](https://github.com/Linaro/squad/issues/554#issuecomment-515172222) bug.

terceiro:
> Cool! Did you have any issues with `django_extensions`?

I often use runserver_plus and shell_plus, and at least runserver_plus didn't work. I didn't note the actual error, but upgrading to the newest upstream version made it work.

chaws:
@terceiro , I've updated travis' build environment to "xenial". The previous one, "trusty", had an old version of Sqlite3 (5 years old!) and it was causing tests to crash because Django 2.2 required a more recent version of it, whereas Django 2.1 didn't. Nevertheless, I opted to update travis' to a more recent distro rather than downgrading Django.

Also, I managed to overcome incompatibilities between "django_restframework_filters" by adding `compat.py`. It seems it's a common place to add compatibility code in packages. Please correct me if you think it's weird.

terceiro:
migrated to  #585

PR #577, opened 2019-08-09
Improvements to dev-docker

terceiro:
I have been using dev-docker recently and these changes make the experience a
lot better for me.

### DIFF
--- a/dev-docker
+++ b/dev-docker
@@ -15,7 +15,7 @@ dockerfile="$datadir"/../Dockerfile.dev
   echo "    groupadd -g $(id -g) $(id -gn) && \\"
   echo "    useradd -m -u $(id -u) -g $(id -g) -s /bin/bash ${USER}"
   echo "USER ${USER}"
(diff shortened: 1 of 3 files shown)


dev-docker
```diff
@@ -11,7 +11,7 @@ mkdir -p "$datadir"
 dockerfile="$datadir"/../Dockerfile.dev
 (
   sed -e '1,/WORKDIR/!d' "$basedir"/Dockerfile
-  echo 'RUN apt-get install -q2 snakefood flake8 git python3-wheel chromium nodejs && \'
+  echo 'RUN apt-get install -q2 snakefood flake8 git python3-wheel chromium nodejs ruby-foreman procps && \'
```
chaws: Pretty neat, I've ran foreman a couple of times back in school,

Perhaps it would be nice to add rabbit-mq too? I've ran `./dev-docker` then `foreman start`, and the worker was never able to connect to a queue server.

terceiro: yes, good point. added rabbitmq-server as dependency and also added a rabbitmq-server process to `Procfile` (see the last commit I pushed)

PR #578, opened 2019-08-14
Added 'Reuse' SPDX compliance tool into travis config

amrohassaan:
Added 'Reuse' SPDX compliance tool into travis.yml to test all newly added files for compliance and report on missing headers


mwasilew:
Sth is wrong with the setup. I would like 'Reuse' to be an additional 'environment' rather than pre-requisite to run any other tests. So if SPDX fails we still know that other tests were passing.

.travis.yml
```diff
@@ -23,6 +23,8 @@ env:
 
 before_install:
   - if test -n "$DATABASE"; then ./scripts/travis-postgresql; fi
+  - docker pull fsfe/reuse:latest
```
chaws: I think `docker run` already pulls `fsfe/reuse` automagically when it's not locally available.

terceiro:
> Sth is wrong with the setup. I would like 'Reuse' to be an additional 'environment' rather than pre-requisite to run any other tests. So if SPDX fails we still know that other tests were passing.

+1 FWIW

PR #579, opened 2019-08-14
some fixes/improvements to the notification system

terceiro:


mwasilew:
I think we need to clearly state this fix in release notes (I mean fetching incomplete jobs). This is a fairly major change in behavior.

Issue #566, opened 2019-07-12
Build compare not useful to LKFT


danrue:
I tried the new build compare feature today. Thanks!

Here's what it looks like: https://qa-reports.linaro.org/_/comparebuilds/?project=lkft/linux-stable-rc-5.1-oe&baseline=v5.1.17-137-gde182b90f76d&target=v5.1.17

I notice two things. First, well, there are 92 pages of results. Second is, most results consist of NA vs PASS. 

How could this be improved? Perhaps some filters to control the comparison. I'd like to be able to see a comparison of things that went from pass to fail, or fail to pass, for example.

chaws:
@danrue that always got me thinking. At first, I think I can add an option that does something similar to art-reports (and to what you commented): show only differences. For now, test results diff are only in notifications, but I think it'll be useful in the front-end as well.

What do you think?

Browsing throughout the 92 pages, I think most of them will disappear when "diff-mode" is on.

danrue:
Hi @chaws! diff mode would help, especially if it let you be particular about which transitions you care to see. Even with diff mode, there may be too many pages of results, so we may want to discard any diffs that only consist of going to or from N/A, for example.

On an unrelated note, is there any way to use this to compare builds across projects? It's a use-case we have, and which I do via the API periodically. Looking at the url parameters, it seems not possible, but I know the back end and API support it. Not sure how you'd support it in the UI though.

Thanks again,
Dan

chaws:
@danrue 

> Hi @chaws! diff mode would help, especially if it let you be particular about which transitions you care to see. Even with diff mode, there may be too many pages of results, so we may want to discard any diffs that only consist of going to or from N/A, for example.

I like the "transitions" idea! After I talked to @mwasilew , we came up with a matrix of transitions that allows multiple filtering choices:

### Display selected transitions
| from/to | pass                   | fail                   | xfail                  | skip                   | n/a                    |
|:-------:|------------------------|------------------------|------------------------|------------------------|------------------------|
| pass    | -                      | <ul><li>[x] </li></ul> | <ul><li>[ ] </li></ul> | <ul><li>[ ] </li></ul> | <ul><li>[ ] </li></ul> |
| fail    | <ul><li>[x] </li></ul> | -                      | <ul><li>[ ] </li></ul> | <ul><li>[ ] </li></ul> | <ul><li>[ ] </li></ul> |
[...]

PR #581, opened 2019-08-16
frontend: filter comparison by transitions

chaws:
Closes https://github.com/Linaro/squad/issues/566

Add a transitions table that allow users to select which transitions from a comparison to be shown.

- All tests that have exact same results along builds and environments are **not** shown
- By default only *pass-to-fail* and *fail-to-pass* transitions are shown
- If a test has at least one differing result, all its results across builds & envs are displayed
- If the transitions selected didn't pick up any change in a specific environment, that column will **not** be displayed
- If no transitions are selected, the frontend will automatically go back to default (*pass-to-fail* and *fail-to-pass*)
- As boxes are checked/unchecked, the frontend will be updated accordingly
- Check "ignore all transitions" box to go back to regular comparison view

Screenshot from 2019-08-16 12-59-14

### DIFF
--- a/squad/core/comparison.py
+++ b/squad/core/comparison.py
@@ -1,6 +1,7 @@
 from collections import OrderedDict
 from django.db.models import F
 from itertools import groupby
+from functools import reduce
 import statistics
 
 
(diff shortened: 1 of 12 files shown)


terceiro:
there is a now a merge conflict, please rebase

terceiro:
can the transition selection be made more compact?

chaws:
@terceiro , rebased and table shrank (see image below). Also I noticed there was a bug with project comparison, which I included a commit in between to fix that. Let me know if you want that fix to be in a separate PR.

Screenshot from 2019-08-19 20-33-51

terceiro:
AIUI "ignore all transitions" should actually say "show all tests".

also I found an issue, see the screenshots below. unless I check "ignore all transitions", no tests are shown.

terceiro:
FWIW I cannot reproduce this last issue anymore

terceiro:
this issue I mentioned last seem to only happen when comparing projects, but not when comparing builds from the same project

chaws:
"show all tests" make more sense indeed. I've renamed it and updated the PR

Regarding your issue, it seems that you're not comparing same environments. See

issue

PR #582, opened 2019-08-19
Added reuse as a test env in travis.

amrohassaan:
Since old PR seems closed, pushing new one with required changes. Added reuse as a test job in travis.yml

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,7 @@ env:
   - SCRIPT="./dev-docker ./manage.py test"
   - SCRIPT="./scripts/test-docker"
   - SCRIPT="./manage.py compilemessages" INSTALLPIP=yes
+  - SCRIPT="docker run -v $(pwd):/squad --workdir /squad fsfe/reuse reuse lint"
 
 before_install:
   - if test -n "$DATABASE"; then ./scripts/travis-postgresql; fi


.travis.yml
```diff
@@ -20,6 +20,7 @@ env:
   - SCRIPT="./dev-docker ./manage.py test"
   - SCRIPT="./scripts/test-docker"
   - SCRIPT="./manage.py compilemessages" INSTALLPIP=yes
+  - SCRIPT="docker run -v $(PWD):/src --workdir /src fsfe/reuse reuse lint"
```
chaws: I think you mean `squad` instead of `src`, which is where reuse will scan our source files.

amrohassaan: Ack, thanks @chaws

.travis.yml
```diff
@@ -20,6 +20,7 @@ env:
   - SCRIPT="./dev-docker ./manage.py test"
   - SCRIPT="./scripts/test-docker"
   - SCRIPT="./manage.py compilemessages" INSTALLPIP=yes
+  - SCRIPT="docker run -v $(pwd):/squad --workdir /squad/squad fsfe/reuse reuse lint"
```
terceiro: why `/squad/squad`? with this you are running from inside the `squad` subdirectory, instead of from the root of the repository.

amrohassaan: I was just having a chat with @chaws right now about it ..we thought /squad is where main dir under the docker container and then /squad/squad is where the source files are. We were having doubts and not sure as well..so It should be 
docker run -v $(pwd):/squad --workdir /squad/ fsfe/reuse reuse lint?

chaws: @terceiro should we run reuse on non-source files as well?

mwasilew: It should be run in top level dir. We have scripts and tests dirs which are also part of SQUAD.

chaws: ack

amrohassaan: @mwasilew exactly yes.

terceiro:
the commit message mentions a fix, but what is being fixed will never appear in the repository. so use a commit message that just says what the patch does in terms of the current contents of the repository, not in terms of previous versions of the patch,

amrohassaan:
@terceiro ack. Amended comment and repushed.

PR #584, opened 2019-08-20
Dockerfile: don't install packages from pip

terceiro:
This ensures that the final docker image will contain only Debian
packages.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -36,7 +36,7 @@ RUN ln -sfT container_settings.py /app/squad/local_settings.py && \
     python3 -m squad.frontend && \
     ./manage.py collectstatic --noinput --verbosity 0 && \
     ./manage.py compilemessages && \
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -36,7 +36,7 @@ RUN ln -sfT container_settings.py /app/squad/local_settings.py && \
     python3 -m squad.frontend && \
     ./manage.py collectstatic --noinput --verbosity 0 && \
     ./manage.py compilemessages && \
-    python3 setup.py develop && \
+    python3 setup.py develop --no-deps && \
```
chaws: How did you figure out this option? I didn't find this at https://setuptools.readthedocs.io/en/latest/setuptools.html#develop-deploy-the-project-source-in-development-mode

terceiro: `python3 setup.py develop --help`

PR #590, opened 2019-08-30
Fix broken JS on test history page

chaws:
Example: https://qa-reports.linaro.org/lkft/linux-next-oe/tests/kselftest/binderfs_binderfs_test

When clicking on the "i" icon, a modal window is expected to pop up. Instead, it was triggering an exception:

`Uncaught SyntaxError: Unexpected token d in JSON at position 45`

Also, this fixes: [LSS-145](https://projects.linaro.org/browse/LSS-145)

### DIFF
--- a/squad/frontend/templates/squad/test_history.jinja2
+++ b/squad/frontend/templates/squad/test_history.jinja2
@@ -43,7 +43,7 @@
               <td class='{{result.status|slugify}}'>
                 <a href="{{project_url(result.test_run)}}">{{result.status}}</a>
                 {% if result.info['test_description'] or result.info['suite_instructions'] or result.info['test_instructions'] or result.info['test_log'] %}
-                  <a href='#' data-toggle="modal" data-target="#info_modal" data-info="{{ result.info }}"><span data-toggle="tooltip" data-placement="right" title="{{ _('Show info') }}" class='fa fa-info-circle'></span></a>
+                  <a href='#' data-toggle="modal" data-target="#info_modal" data-info="{{ to_json(result.info) }}"><span data-toggle="tooltip" data-placement="right" title="{{ _('Show info') }}" class='fa fa-info-circle'></span></a>
                 {% endif %}
                 {% if known_issues %}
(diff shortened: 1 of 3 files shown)


squad/frontend/templatetags/squad.py
```diff
@@ -196,3 +198,8 @@ def crispy(context, form, **options):
     for option, value in options.items():
         setattr(helper, option, value)
     return render_crispy_form(form, helper=helper, context=context)
+
+
+@register_global_function()
+def to_json(d):
+    return json.dumps(d)
```
mwasilew: shouldn't you catch potential exception here?

chaws: yeah, you're right. That's TypeError. I'll make it return an empty string if `d` can't be parsed to json

squad/frontend/templates/squad/test_history.jinja2
```diff
@@ -43,7 +43,7 @@
               <td class='{{result.status|slugify}}'>
                 <a href="{{project_url(result.test_run)}}">{{result.status}}</a>
                 {% if result.info['test_description'] or result.info['suite_instructions'] or result.info['test_instructions'] or result.info['test_log'] %}
-                  <a href='#' data-toggle="modal" data-target="#info_modal" data-info="{{ result.info }}"><span data-toggle="tooltip" data-placement="right" title="{{ _('Show info') }}" class='fa fa-info-circle'></span></a>
+                  <a href='#' data-toggle="modal" data-target="#info_modal" data-info="{{ to_json(result.info) }}"><span data-toggle="tooltip" data-placement="right" title="{{ _('Show info') }}" class='fa fa-info-circle'></span></a>
```
terceiro: it seems that json.dumps will always use double quotes for strings, so AFAICT for this to work the HTML should be changed to use single quotes?

chaws: That's true, but Jinja will render those into html entities. The javascript bit of this page will convert them back to quotes when interpreting such string back to JS objects.

PR #593, opened 2019-09-04
yaml_validator: drop usage of simple yaml.load

terceiro:
This is insecure, and deprecated.


chaws:
Is this for Axel's issue?

PR #605, opened 2019-10-02
ci: models: backend: handle duplicated testjob as new exception

chaws:
Add DuplicatedTestJob exception to handle cases where a testjob is being submitted twice. Before this patch, we were considering that case as `InvalidMetadata`.

### DIFF
--- a/squad/ci/models.py
+++ b/squad/ci/models.py
@@ -69,60 +69,58 @@ class Backend(models.Model):
             self.really_fetch(test_job)
 
     def really_fetch(self, test_job):
-        implementation = self.get_implementation()
-
         test_job.last_fetch_attempt = timezone.now()
+
+        implementation = self.get_implementation()
+        results = implementation.fetch(test_job)
+        if not results:
+            test_job.save()
+            return
+
+        status, completed, metadata, tests, metrics, logs = results
+
+        test_job.job_status = status
+        if not completed:
+            tests = {}
+            metrics = {}
+            test_job.can_resubmit = True
+
+        if completed and not tests and not metrics:
+            # test job produced no results
+            # mark it incomplete
+            completed = False
+            test_job.can_resubmit = True
+
+        metadata['job_id'] = test_job.job_id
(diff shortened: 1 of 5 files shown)


squad/ci/models.py
```diff
@@ -68,60 +68,58 @@ def fetch(self, test_job):
...
+                log_file=logs,
+                completed=completed,
+            )
+            test_job.testrun = testrun
+        except InvalidMetadata as exception:
+            test_job.failure = str(exception)
+
+        # mark test job as fetched to prevent resubmission
+        # on next fetch attempt
+        test_job.fetched = True
```
mwasilew: I think this block should be inside 'except' block. Otherwise you will mark job as fetched even if creating TestRun fails. I'm not sure this is a good idea.

squad/ci/models.py
```diff
@@ -68,60 +68,58 @@ def fetch(self, test_job):
             self.really_fetch(test_job)
 
     def really_fetch(self, test_job):
-        implementation = self.get_implementation()
-
         test_job.last_fetch_attempt = timezone.now()
-        test_job.save()
```
terceiro: the intention of saving here is to ensure that if the fetching process crashes with an exception, the date of the last fetch attempts is still recorded. if you only save after the actual fetching, that date will never be updated for jobs that cause a crash

mwasilew: The problem is that there might be a few reasons for not an exception here. One of them is that more than one worker picket the fetching task. When the first one finishes it saves actual result to database (for example incomplete job with infra error). When second worker picks the task it overwrites this error with squad internal error saying 'sorry, there already is a test job with this ID'. I asked to simply ignore this exception as it's not very useful from user point of view.

chaws:
@mwasilew, rebased it with master

squad/ci/models.py
mwasilew: We went through the code and it looks like the timestamp will be saved in squad.ci.tasks.fetch (as it is now). So I'm OK to merge this.

PR #610, opened 2019-10-10
Improve fetch performance

chaws:
This improves the overall performance of ProjectStatus.create_or_update. The two main bottlenecks where in TestSummary and TestComparison, consuming a lot of memory and making thousands of queries to db, respectively. I've managed to perform some heavy tasks in db.

In my local database (October last year's dump from production) I got some benchmarks that reduced the time in half using this [patch](https://pastebin.linaro.org/view/20455d97) compared to [master](https://pastebin.linaro.org/view/5416423a). Maybe you guys can help out just to be sure it'll improve performance overall.

### DIFF
--- a/squad/core/comparison.py
+++ b/squad/core/comparison.py
@@ -164,11 +164,16 @@ class TestComparison(BaseComparison):
 
     __test__ = False
 
-    def __init__(self, *builds):
+    def __init__(self, *builds, regressions_and_fixes_only=False):
         self.__intermittent__ = {}
+        self.regressions_and_fixes_only = regressions_and_fixes_only
         BaseComparison.__init__(self, *builds)
 
     def __extract_results__(self):
+        if self.regressions_and_fixes_only:
+            self.__regressions_and_fixes__(self.builds[0], self.builds[1])
+            return
+
         test_runs = models.TestRun.objects.filter(
             build__in=self.builds,
         ).prefetch_related(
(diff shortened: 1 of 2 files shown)


squad/core/models.py
```diff
@@ -1126,13 +1121,6 @@ def __init__(self, build, environment=None):
                 self.tests_xfail += 1
             elif test.status == 'skip':
                 self.tests_skip += 1
-            else:
-                raise InvalidTestStatus(test.status)
-            if test.status == 'fail':
```
mwasilew: I think this is used in email templates. At least it can be.

squad/core/models.py
```diff
@@ -1013,7 +1013,6 @@ def create_or_update(cls, build):
             status.metrics_summary = metrics_summary.value
             status.has_metrics = metrics_summary.has_metrics
             status.last_updated = now
-            finished, _ = build.finished
```
mwasilew: I think you also need to change next line to copy finished status from Build object. Otherwise it will not change ever.

squad/core/models.py
chaws: ah, that's why is good to have another pair of eyes, replaced it!

squad/core/models.py
chaws: I removed this line because `build.finished` is already called above (https://github.com/Linaro/squad/blob/master/squad/core/models.py#L996). Is this second call on line 1026 needed for some special case? I noticed build.finished computes the finished state for every call.

mwasilew: Correct. I didn't look high enough :)

squad/core/comparison.py
```diff
@@ -164,11 +164,16 @@ class TestComparison(BaseComparison):
...
+    def __init__(self, *builds, regressions_and_fixes_only=False):
         self.__intermittent__ = {}
+        self.regressions_and_fixes_only = regressions_and_fixes_only
         BaseComparison.__init__(self, *builds)
 
     def __extract_results__(self):
+        if self.regressions_and_fixes_only:
+            self.__regressions_and_fixes__(self.builds[0], self.builds[1])
+            return
+
```
terceiro: I don't see `regressions_and_fixes_only` being used anywhere. So `__regressions_and_fixes__` below is not being called anywhere.

squad/core/models.py
```diff
@@ -1114,38 +1113,38 @@ def __init__(self, build, environment=None):
         self.tests_fail = 0
         self.tests_xfail = 0
         self.tests_skip = 0
-        self.failures = OrderedDict()
+        self.failures = defaultdict(list)
```
terceiro: the use of `OrderedDict` is usually to ensure consistent ordering of items in the UI, both web and email

squad/core/comparison.py
chaws: Ah, forgot to use that in `ProjectStatus.create_or_update`

squad/core/models.py
chaws: Oh, missed that one

PR #612, opened 2019-10-11
ci: allow disabling listening to some backends

terceiro:
This is useful for a few use cases:

* LAVA instances that don't provide a ZMQ interface
* Having remote backends configured locally for debugging once in a
  while, without being constantly connected to them.


chaws:
LGTM

PR #613, opened 2019-10-16
Default zero expected testruns

chaws:
Setting `expected_test_runs` to zero will force builds to finish earlier. Some points worth noting:

* In the first valid call to `api/submit`, a notification might be sent out if there is no submitted testjob waiting to be fetched
* A `finished` build can still receive more testruns, but `send_status_notification` will already be sent


danrue:
:+1: setting this field is a common chore in LKFT. Thanks. Notifications shouldn't be sent out early - that's what "wait before notification" is for.

Issue #592, opened 2019-09-03
Plugin to parse LAVA logs to find kernel issues


danrue:
We would like to be able to catch kernel panics, oopses, and warnings that may occur during LKFT runs. Each of these things should produce a "pass" or "fail" test result.

For example, we could have a dictionary like the following:
```py
kernel_tests = {
  'kernel_panic': r'',
  'kernel_oops': r'',
}
```

For each of these items, the regular expression will be run against the LAVA log and a test result will be produced.

Some regular expressions can already be found in LAVA @ https://git.lavasoftware.org/lava/lava/blob/master/lava_common/constants.py#L93-99

LAVA's implementation is not sufficient because it only runs against the boot portion of the job. LKFT would like to parse the entire kernel log for such errors.

I'm not sure how test names should be produced. It should be unique per job, but not per build.

cc @mrchapp @mattface @nareshkamboju

mwasilew:
It's the same request as in squadplugins: https://github.com/Linaro/squadplugins/issues/3

danrue:
@chaws pointed out https://github.com/Linaro/squad/blob/master/squad/plugins/linux_log_parser.py, which seems to already do exactly what I proposed in this issue. I've enabled it on https://qa-reports.linaro.org/lkft/linux-next-oe-sanity/ and https://qa-reports.linaro.org/lkft/linux-next-oe/ and will roll it out to the rest of the LKFT branches if it works as expected.

danrue:
This isn't quite right.

![image](https://user-images.githubusercontent.com/409011/64352753-2ef62880-cfc2-11e9-905b-c17da9e8481b.png)

In LKFT's case, we run about 20 test jobs per environment, but we only get one check-oops result per environment. It should be one per job.

https://qa-reports.linaro.org/lkft/linux-next-oe/build/next-20190904/#!?details=47906

mwasilew:
After a chat with @danrue we agreed that the plugin should report panics/oopses/warnings for each LAVA test job (which translatest to TestRun). It will take a metadata key (to be agreed) and add the key to the test name. So for example for ltp-syscalls test job it will produce:
* ltp-syscalls-panics
* ltp-syscalls-oopses
* ltp-syscalls-warnings
[...]

PR #614, opened 2019-10-17
plugins: linux_log_parser: improve kernel log parsing

chaws:
Closes https://github.com/Linaro/squad/issues/592 and https://github.com/Linaro/squadplugins/issues/3. This PR has been moved from https://github.com/Linaro/squadplugins/pull/13

I managed to put several regex'es together so that findall is called only once. Not sure if that has better performance than running each regex on its own. Also, there's a pre-regex that filters-out all non-kernel messages (that don't start with [number.number]), thus reducing log size considerably (specially in jobs with lots of target output).

By default, there are 6 tests, each represented by a regex, all under kernel-issues suite:

if a test has no match, result=True
if a test has at least one match, result=False
if a test matches more than once, all occurrences are concatenated and saved to its log. Meaning there will always be one test per build per testrun (2 Oopses == 1 test)

**NOTE:** It also addresses Dan's point of having multiple testsruns per environment. It appends `job_id` to the testname. Although helpful distinguishing tests across different test jobs, it won't be helpful finding regressions and fixes due to test name uniqueness.

### DIFF
--- a/squad/plugins/linux_log_parser.py
+++ b/squad/plugins/linux_log_parser.py
@@ -1,67 +1,88 @@
-from squad.core.plugins import Plugin as BasePlugin
+import logging
+import re
+from squad.plugins import Plugin as BasePlugin
 
 
-class Issue(object):
-    name = None
-    found = False
-    done = False
-    max_lines = None
+logger = logging.getLogger()
 
-    def __init__(self):
-        self.__log__ = []
+REGEX_NAME = 0
+REGEX_BODY = 1
 
-    @classmethod
-    def find(cls, log):
-        issues = [issue_type() for issue_type in cls.__subclasses__()]
-        for line in log.splitlines():
-            for issue in issues:
-                issue.feed(line)
-        return issues
+MULTILINERS = [
+    ('check-kernel-panic', r'Kernel panic - not syncing.*?end Kernel panic - not syncing'),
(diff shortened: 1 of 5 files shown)


squad/plugins/linux_log_parser.py
```diff
@@ -1,67 +1,87 @@
...
-            for issue in Issue.find(testrun.log_file):
-                testrun.tests.create(
-                    suite=suite,
-                    name='check-' + issue.name,
-                    result=(not issue.found),
-                    log=issue.log
-                )
+        log = self.__kernel_msgs_only(testrun.log_file)
+        suite, _ = testrun.build.project.suites.get_or_create(slug='linux-log-parser')
+        test_name_suffix = ('-%s' % testrun.job_id) if testrun.job_id else ''
```
terceiro: I would use the job name instead. this way if the jobs are named correctly, then you have recurring test names across builds.

chaws: Makes sense. I remember a while ago @mwasilew saying that this could also be a field in the metadata. At any rate, I've changed `postprocess_testrun` to `postprocess_testjob` so the plugin would have easy access to the testjob.

chaws:
Fixes https://github.com/Linaro/squad/issues/633

PR #615, opened 2019-10-17
core: tasks: ReceiveTestRunData: ignore tests and metrics with long names

chaws:
Ignore tests and metrics that have names longer than 256 characters. We should consider increasing the max length of the name field in the future though.

This avoids getting an exception: DataError value too long for type character varying(256)

### DIFF
--- a/squad/core/tasks/__init__.py
+++ b/squad/core/tasks/__init__.py
@@ -206,6 +206,9 @@ class ParseTestRunData(object):
             issues[issue.test_name].append(issue)
 
         for test in test_parser()(test_run.tests_file):
+            # TODO: remove check below when test_name size changes in the schema
+            if len(test['test_name']) > 256:
+                continue
             suite = get_suite(
                 test_run,
                 test['group_name']
@@ -226,6 +229,9 @@ class ParseTestRunData(object):
                 test_obj.known_issues.add(issue)
 
         for metric in metric_parser()(test_run.metrics_file):
+            # TODO: remove check below when test_name size changes in the schema
+            if len(metric['name']) > 256:
+                continue
             suite = get_suite(
                 test_run,
                 metric['group_name']
--- a/test/core/test_tasks.py
+++ b/test/core/test_tasks.py
@@ -263,10 +263,10 @@ class CommonTestCase(TestCase):
     def setUp(self):
         group = Group.objects.create(slug='mygroup')
         project = group.projects.create(slug='mygroup')
-        build = project.builds.create(version='1.0.0')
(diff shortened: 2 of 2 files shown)


squad/core/tasks/__init__.py
```diff
@@ -206,6 +206,8 @@ def __call__(test_run):
             issues[issue.test_name].append(issue)
 
         for test in test_parser()(test_run.tests_file):
+            if len(test['test_name']) > 256:
```
mwasilew: please add TODO comment here so we know that this should be removed when the DB schema changes

chaws: sure

terceiro: is there a reason to not increase the max size already?

whatever the new size is, we should document the limit explicitly. otherwise we can run into insanely long names again in the future.

chaws: No particular reason. What's a good number? Also, do you think this may affect database performance? I have no clue on how postgres' data layout.

mwasilew: I thought 256 was long enough... and than came CTS :(

chaws: Java....

terceiro: > No particular reason. What's a good number? Also, do you think this may affect database performance? I have no clue on how postgres' data layout.

not sure. you would have to check. since tests is by far our largest table, any schema changes in it can be pretty disruptive as a migration that requires rewriting the underlying storage for the rows will take forever.

maybe we could go ahead with ignoring the tests with names that are too large for now. another option would be to hash these large names and then have another, small table that maps these hashes to the actual names.

chaws: I agree with that. Running a simple SQL, I roughly got the amount of space used just to store test_names:

```
productionqareports=> SELECT SUM(length(name)) FROM core_test;
    sum     
------------
 9139882540
```

Since each character is 1 byte, the number above results in ~8GB.

Do you guys think it'd be a good idea to move all test names to an external table? I think so

mwasilew:
Let's merge it to avoid crash for now and follow up with proper fix.

squad/core/tasks/__init__.py
chaws: Just for the record there are 126552 unique test names in production: 

```
productionqareports=> SELECT COUNT(t.*) FROM (SELECT DISTINCT name FROM core_test ORDER BY name ASC) AS t;
 count  
--------
 126552
```

So given that qa-reports is over 200 million records in test table, that gives an average of 1500+ copies of each test name.

PR #617, opened 2019-10-22
core: ProjectStatus: delay fixes and regressions

chaws:
Delay calculation of fixes and regressions to after a build has finished status. This will reduce amount of time to run UpdateProjectStatus

### DIFF
--- a/squad/core/models.py
+++ b/squad/core/models.py
@@ -973,6 +973,7 @@ class ProjectStatus(models.Model, TestSummaryBase):
         returns it.
         """
 
+        build_finished, _ = build.finished
         test_summary = build.test_summary
         metrics_summary = MetricsSummary(build)
         now = timezone.now()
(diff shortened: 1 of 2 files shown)


terceiro:
I think we can try this. Let's get it in so we can test on staging.

danrue:
Is there any middle ground on this? We do use the fixes and regressions often while builds are still under test.

mwasilew:
I think this can be done using a separate task in a separate queue. We don't want to delay fetch operation. I also think we should try creating a separate set of queues for each LAVA instance. This won't help on it's own but will allow to have workers dedicated for LAVA backend.

In any case let's see how this works in staging. It already has enough data to cause the delay in fetch operation. I'll test it with @chaws tomorrow and report back. We're not rolling this out to production yet.

danrue:
I wonder about more generalized solutions for the queue problems.

For example, what if each project, or each group, had its own queue, and the workers used round-robin to pick items from one queue after another. This way, smaller projects or groups didn't get starved, and larger projects still got equal or greater time (more time when other projects had no queue tasks). I don't suppose celery supports any such mechanisms?

mwasilew:
Celery does round robin on queues as far as I understand. Your approach is just a bit more complicated version of my proposal. Projects will use CI backends and I'm proposing round robin on these. In any case I didn't yet try to implement this and it's not easy in celery to make dynamic task routing. However I think this just moves the problem a bit into the future. Proper solution is to make the regression/fixes calculations quicker.

chaws:
Doing https://github.com/Linaro/squad/pull/620 might calculate regressions and fixes a bit better, thus doing UpdateProject status all the time wouldn't delay the whole thing

PR #619, opened 2019-10-24
Use status for build summary

chaws:
Do fast count when `with_failures=False`, given that a testrun summary is already recorded in `Status`, we can reuse those numbers to retrieve build's summary much faster, instead of counting test by test.

### DIFF
--- a/squad/core/models.py
+++ b/squad/core/models.py
@@ -974,7 +974,7 @@ class ProjectStatus(models.Model, TestSummaryBase):
         """
 
         build_finished, _ = build.finished
-        test_summary = build.test_summary
+        test_summary = TestSummary(build, with_failures=False)
         metrics_summary = MetricsSummary(build)
         now = timezone.now()
         test_runs_total = build.test_runs.count()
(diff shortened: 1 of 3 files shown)


squad/core/models.py
```diff
@@ -1104,7 +1104,7 @@ class TestSummary(TestSummaryBase):
 
     __test__ = False
 
-    def __init__(self, build, environment=None):
+    def __init__(self, build, environment=None, with_failures=True):
```
terceiro: not sure about this new conditional. maybe we should move collecting the failures to the same code that already calculates fixes and regressions (and with this scan every single in only one place)

The only place that actually seems to use this list of failures is the notification code, which will already also use fixes and regressions anyway.

chaws: makes sense to me. I almost removed that a while ago, but @mwasilew pointed it out that notification emails might use that :)

Since the topic is hot, I'm trying to think a way of calculating regressions and fixes without having to rely on loading all tests into memory. If only there was a way to quickly get tests directly thru a hash or something...

chaws:
I'm closing this, as it would also hit a bottleneck for big numbers of testruns. I'm working on a PR with an improved version of this

PR #621, opened 2019-10-29
Test comparison

chaws:
After https://github.com/Linaro/squad/pull/620, staging DB just took forever to calculate TestComparison for builds with over 1000 testruns. Postgres queries were taking an insane amount of time to return (10+ hours). This patch fetch tests using batch of 400 test runs ids at a time. I've run a few queries in staging and production and this number seems to be the sweet spot.

### DIFF
--- a/squad/core/comparison.py
+++ b/squad/core/comparison.py
@@ -5,7 +5,7 @@ from functools import reduce
 import statistics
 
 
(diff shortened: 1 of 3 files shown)


chaws:
All tests have passed. I'm merging this PR myself so we can see it in action on staging.

PR #622, opened 2019-10-30
core: comparison: remove the use of iterator over tests

chaws:
Remove the use of iterator. This causes postgres to increase number of locks some how.


chaws:
All checks passed, I'm merging this myself then test it one more time for Lite, Lkft and a android-lkft(the ones with hundreds of thousands of tests) before rolling it out to production.

PR #626, opened 2019-11-04
Reduce number of queries

chaws:
This PR changes several snippets in the code, trying to minimize the number of trips to db, applying less load on it too:

1. Re-adjust [chunk size](https://github.com/Linaro/squad/blob/master/squad/core/comparison.py#L190) from 400 to 100. In a single threaded architecture (dev), 400 testrun-ids at a time proved to be the sweet spot, but on a multi-threaded one (staging and production), having multiple calls to that snippet makes things slow again.

2. Prefetch known_issues from failing tests. If a testrun contains 1000 tests with known_issues, this would result on 1000 extra queries in the DB. I just moved them to a separate method to be called after tests are processed

3. Simplify calculations for TestSummary using States instead of summing test-by-test. I also added what @terceiro asked in https://github.com/Linaro/squad/pull/619#discussion_r339012500.

### DIFF
--- a/squad/core/models.py
+++ b/squad/core/models.py
@@ -22,7 +22,7 @@ from django.utils.translation import ugettext_lazy as N_
 from simple_history.models import HistoricalRecords
 
 from squad.core.utils import parse_name, join_name, yaml_validator, jinja2_validator
-from squad.core.utils import encrypt, decrypt
+from squad.core.utils import encrypt, decrypt, split_list
 from squad.core.comparison import TestComparison
 from squad.core.statistics import geomean
 from squad.core.plugins import Plugin
@@ -1099,12 +1099,6 @@ class NotificationDelivery(models.Model):
         return (not created)
 
 
-class InvalidTestStatus(Exception):
-
-    def __init__(self, status):
-        super(InvalidTestStatus, self).__init__('Invalid status: %s' % status)
-
-
 class TestSummary(TestSummaryBase):
 
     __test__ = False
@@ -1114,38 +1108,21 @@ class TestSummary(TestSummaryBase):
         self.tests_fail = 0
         self.tests_xfail = 0
         self.tests_skip = 0
-        self.failures = OrderedDict()
 
         query_set = build.test_runs
         if environment:
             query_set = query_set.filter(environment=environment)
-        test_runs = query_set.prefetch_related(
-            'environment',
-            'tests',
-            'tests__suite'
-        ).order_by('id')
 
-        tests = OrderedDict()
-        for run in test_runs.all():
-            for test in run.tests.all():
-                tests[(run.environment, test.suite, test.name)] = test
+        query_set = query_set.values('id').order_by('id')
 
-        for context, test in tests.items():
-            if test.status == 'pass':
-                self.tests_pass += 1
-            elif test.status == 'fail':
-                self.tests_fail += 1
-            elif test.status == 'xfail':
-                self.tests_xfail += 1
-            elif test.status == 'skip':
-                self.tests_skip += 1
-            else:
-                raise InvalidTestStatus(test.status)
-            if test.status == 'fail':
-                env = test.test_run.environment.slug
-                if env not in self.failures:
-                    self.failures[env] = []
-                self.failures[env].append(test)
+        test_runs_ids = [test_run['id'] for test_run in query_set]
+        for chunk in split_list(test_runs_ids, chunk_size=100):
+            status = Status.objects.filter(suite=None, test_run_id__in=chunk)
+            for s in status:
+                self.tests_pass += s.tests_pass
+                self.tests_fail += s.tests_fail
+                self.tests_xfail += s.tests_xfail
+                self.tests_skip += s.tests_skip
 
 
 class MetricsSummary(object):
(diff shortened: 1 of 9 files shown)


squad/core/models.py
```diff
@@ -1114,38 +1108,24 @@ def __init__(self, build, environment=None):
...
-            if test.status == 'fail':
-                env = test.test_run.environment.slug
-                if env not in self.failures:
-                    self.failures[env] = []
-                self.failures[env].append(test)
+
+        # 'build' needs to be explicitly passed as one of the wanted fields, otherwise Django
+        # will generate one query per testrun to fetch it's build, even though we're not using it.
+        # We actually only the the id
+        query_set = query_set.only('build').order_by('id')
```
terceiro: in my local tests, adding `only('build')` is what made several queries be executed! e.g.:

```
 build = Build.objects.last()
In [2]: query_set = build.test_runs
In [3]: query_set.all()

<< 1 query >>

In [4]: query_set.order_by('id')

<< 1 query (same as above plus the ORDER BY) >>

Out[4]: SELECT "core_testrun"."id",
       "core_testrun"."build_id",
       "core_testrun"."environment_id",
       "core_testrun"."created_at",
       "core_testrun"."completed",
       "core_testrun"."datetime",
       "core_testrun"."build_url",
       "core_testrun"."job_id",
       "core_testrun"."job_status",
       "core_testrun"."job_url",
       "core_testrun"."resubmit_url",
       "core_testrun"."data_processed",
       "core_testrun"."status_recorded"
  FROM "core_testrun"
 WHERE "core_testrun"."build_id" = 126
 ORDER BY "core_testrun"."id" ASC
 LIMIT 21


Execution time: 0.002020s [Database: default]

<QuerySet [<TestRun: #10593>, <TestRun: #10595>, <TestRun: #10594>, <TestRun: #10596>, <TestRun: #10597>, <TestRun: #10598>, <TestRun: #10599>, <TestRun: #10600>, <TestRun: #10601>, <TestRun: #10602>, <TestRun: #10603>, <TestRun: #10604>, <TestRun: #10605>, <TestRun: #10606>, <TestRun: #10607>, <TestRun: #10608>, <TestRun: #10609>, <TestRun: #10610>, <TestRun: #10612>, <TestRun: #10611>, '...(remaining elements truncated)...']>

In [5]: query_set.only('build').order_by('id')
<< several queries. one like this: >>

Out[5]: SELECT "core_testrun"."id",
       "core_testrun"."build_id"
  FROM "core_testrun"
 WHERE "core_testrun"."build_id" = 126
 ORDER BY "core_testrun"."id" ASC
 LIMIT 21


Execution time: 0.001103s [Database: default]

<< then A LOT like this: >>

SELECT "core_testrun"."id",
       "core_testrun"."job_id"
  FROM "core_testrun"
 WHERE "core_testrun"."id" = 10614


Execution time: 0.000244s [Database: default]

<QuerySet [<TestRun: #10593>, <TestRun: #10595>, <TestRun: #10594>, <TestRun: #10596>, <TestRun: #10597>, <TestRun: #10598>, <TestRun: #10599>, <TestRun: #10600>, <TestRun: #10601>, <TestRun: #10602>, <TestRun: #10603>, <TestRun: #10604>, <TestRun: #10605>, <TestRun: #10606>, <TestRun: #10607>, <TestRun: #10608>, <TestRun: #10609>, <TestRun: #10610>, <TestRun: #10612>, <TestRun: #10611>, '...(remaining elements truncated)...']>
```

chaws: Fixed it, with `models.TestRun.objects.filter(build=build).only('id')` instead of 'build.test_runs.only('id')'.

terceiro: that still generates 1 extra query for each test run. You want values():

```python
query_set.values('id').order_by('id')
```

the result is then a list of dict's, not TestRun objects.

chaws: Updated to use values now : )

chaws:
@terceiro , I've addressed your comments above and the ones from irc. Now `split_dict` does not create extra objects while slicing the dictionary.

chaws:
Just for the record, I've compared performance between fetching States using testrun_id chunks against using all testrun_ids at once fetching data with Django's `iterator`. Here's the snippet of the little benchmark I've run:

```python
from squad.core import models
import time

build_lite = models.Group.objects.filter(slug='lite').first().projects.filter(slug='zephyr-upstream').first().builds.filter(version='master').first()

def split_list(the_list, chunk_size):
    result_list = []
    while the_list:
        result_list.append(the_list[:chunk_size])
        the_list = the_list[chunk_size:]
    return result_list

def with_chunks():
    global build_lite
    global split_list
    global models
    testruns = build_lite.test_runs.values('id').order_by('id')
    tr_ids = [tr['id'] for tr in testruns.all()]
    ids = []
    for chunk in split_list(tr_ids, 100):
        status = models.Status.objects.filter(suite=None, test_run_id__in=chunk)
        for s in status:
            ids.append(s.id)
    print('with_chunks: len(ids) = %d' % len(ids))

def with_iterator():
    global build_lite
    global models
    ids = []
    for status in models.Status.objects.filter(suite=None, test_run__in=build_lite.test_runs.all()):                                                                                                               
        ids.append(status.id)
    print('with_iterator: len(ids) = %d' % len(ids))

start = time.time(); with_chunks(); print(time.time() - start);
start = time.time(); with_iterator(); print(time.time() - start);
```

I got local results:

```
with_chunks: len(ids) = 9088
0.7558376789093018
with_iterator: len(ids) = 9088
4.862984657287598
```

Staging and production had similar behavior. It turns out that using all ids at once is very expensive compared to using 100 at a time. I've noticed the same pattern when querying the test table.

PR #630, opened 2019-11-19
Dockerfile: ignore version constraints from requirements.txt

terceiro:
The version constraints in requirements.txt are needed when installing
from pypi, but when installing from Debian packages we can just trust
what's in the repository since we do run the tests against that.

The issue manifests when you call the installed commands (e.g.
squad-admin), but not when you call ./manage.py. So, also add an
explicit call to squad-admin to scripts/test-docker so we know if we
introduce a regression.


chaws:
LGTM

PR #632, opened 2019-11-22
celery: add support to SQS

chaws:
This adds support to Amazon SQS


chaws:
Let's do SQS only in deployment

PR #637, opened 2019-11-26
settings: add support to prefix-named queues

chaws:
This allows celery to create queues with user defined prefix names
ref: http://docs.celeryproject.org/en/latest/getting-started/brokers/sqs.html#queue-prefix

### DIFF
--- a/doc/install.rst
+++ b/doc/install.rst
@@ -206,6 +206,10 @@ __ https://docs.djangoproject.com/en/1.11/topics/email/
   Useful when multiple environments share the same broker.
   Defaults to ``''``.
 
+* ``SQUAD_CELERY_POLL_INTERVAL``: Number of seconds a worker will sleep
+  after an empty answer from SQS before the next polling attempt.
+  Defaults to ``1``.
+
 User management
 ---------------
 
--- a/doc/install.rst
+++ b/doc/install.rst
@@ -133,6 +133,10 @@ useful::
 
 By default, workers listen to all queues.
 
+For message brokers that support prefixed-queue names, SQUAD has the optional
+environment variable `SQUAD_CELERY_QUEUE_NAME_PREFIX`, that, if set, will
+prepend it before all queues.
+
 Further configuration
 ---------------------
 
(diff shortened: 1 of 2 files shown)


mwasilew:
I think some mention about this feature in this chapter: https://squad.readthedocs.io/en/latest/install.html#worker-configuration would be handy.

terceiro:
> I think some mention about this feature in this chapter: https://squad.readthedocs.io/en/latest/install.html#worker-configuration would be handy.

yes please

chaws:
I added a small note at the end of `worker-configuration` section.

doc/install.rst
```diff
@@ -133,6 +133,10 @@ useful::
 
 By default, workers listen to all queues.
 
+For message brokers that support prefixed-queue names, SQUAD has the optional
+setting `SQUAD_CELERY_QUEUE_NAME_PREFIX`, that, if set, will prepend it before
```
terceiro: it's actually an environment variable, not a setting. it needs to be listed also in `doc/install.rst`, together with the rest of supported environment variables

chaws: Right! I reworded `setting` to `environment variable`, the correct name indeed. Also added it to the environment variables below in the same file.

chaws:
@terceiro @mwasilew FYI: I miss interpreted SQS pricing system. I originally thought that it was free up to 1 million messages per month. It turns out, it's not messages, but requests.

In one hand this is good because a single request can fetch multiple messages, but on the other hand when the queue is empty, it takes one request to realize that the queue is empty, then it needs to keep polling the server until it finds new messages. 

In my toy copy of qa-reports running SQS, it passed the free 1 million tier in just a few days. Calculating why this happened is quite easy: 60 requests per second * 60 min in an hour * 24 hours in a day * 9 workers (3 vms, each running 3 threads of workers [production are 5 threads]). That's over 700k requests just to keep workers alive for one day!

Setting the polling interval to 60s/1min should drop the number of requests to just a few dozen thousands a day.

mwasilew:
@chaws makes sense. I think 1 min is a good compromise. Will this also apply to rabbitmq?

chaws:
@mwasilew I've looked a little bit, but couldn't find rabbitmq using that type of setting. I guess in that case, the setting is just ignored.

PR #640, opened 2019-12-10
Add coverage testing with pytest

terceiro:
Our current coverage is almost 79%, so we will fail tests if it goes
under 78. With time we want to increase that threashold as high as
feasible.

It would be nice to switch to pytest completely at some point, but there
are more changes needed for that to happen

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,5 @@
+[run]
+source = squad
+omit = squad/wsgi.py
+       squad/*/apps.py
+       squad/*settings.py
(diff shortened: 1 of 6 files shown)


chaws:
Cool! One question: do we need to bump up the minimum coverage threshold every time it increases?

terceiro:
> Cool! One question: do we need to bump up the minimum coverage threshold every time it increases?

yes. the idea is to revisit that number as the coverage increases.

PR #641, opened 2019-12-10
Fix test results page

chaws:
This fixes the test results page with tests with no metadata and also makes it load a bit faster

### DIFF
--- a/squad/frontend/tests.py
+++ b/squad/frontend/tests.py
@@ -125,17 +125,16 @@ class TestResultTable(list):
         for test in tests:
             memo.setdefault(test.full_name, {})
             memo[test.full_name][test.test_run.environment_id] = [test.status]
-            if test.metadata.description or test.suite.metadata.instructions_to_reproduce or test.metadata.instructions_to_reproduce or test.log:
-                error_info = {
-                    "test_description": test.metadata.description,
-                    "suite_instructions": test.suite.metadata.instructions_to_reproduce,
-                    "test_instructions": test.metadata.instructions_to_reproduce,
-                    "test_log": test.log
-                }
-                memo[test.full_name][test.test_run.environment_id].append(
-                    json.dumps(error_info))
-            else:
-                memo[test.full_name][test.test_run.environment_id].append(None)
+
+            error_info = {
+                "test_description": test.metadata.description if test.metadata else '',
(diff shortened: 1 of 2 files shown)


squad/frontend/tests.py
```diff
@@ -46,18 +46,31 @@ def __init__(self):
         self.paginator = self
         self.paginator.num_pages = 0
         self.number = 0
+        self.all_tests = []
+
+    def __get_all_tests__(self, build, search):
+        self.all_tests = []
```
terceiro: duplicates what's already in the constructor

squad/frontend/tests.py
```diff
@@ -46,18 +46,31 @@ def __init__(self):
...
+
+        test_runs = TestRun.objects.filter(build=build).values('id')
+        test_runs_ids = [test_run['id'] for test_run in test_runs]
+        for chunk in split_list(test_runs_ids, chunk_size=100):
+            query_set = Test.objects.filter(test_run_id__in=chunk).prefetch_related('suite')
+            if search:
+                query_set = query_set.filter(name__icontains=search)
+
+            tests = query_set.only('id', 'suite', 'name', 'result', 'has_known_issues').order_by()
+            self.all_tests += tests
```
terceiro: this looks to me like you are loading all the tests, and then using the potentially huge list to calculate the pagination. the point of pagination is not having to load everything from the database in the first place.

squad/frontend/tests.py
chaws: fixed

squad/frontend/tests.py
chaws: @terceiro previous code was running 3 significantly heavy queries: the first query did string matching + distinct() + count() just to get the count of tests. The second query (again with string matching) was ordering tests by number of failures/xfails/skips/passes, the result of this second query was just to build a dynamic [filter](https://github.com/Linaro/squad/blob/master/squad/frontend/tests.py#L119) for the third query that generates content for the test results page, which used suite_id and test name to query tests. Although test name is indexed, this would still take some time process.

All that was replaced by 2 much lighter queries: first one that does string matching and return all tests for a build. All previous calculations are made in memory: pagination, ordering and selection of the tests to be displayed. The second query simply passes the wanted tests ids.

Running these changes locally for an android-lkft project, it lowered loading times from 4000ms to just a couple of 100ms. Also, we are already doing something similar for workers, when loading tests to generate regressions and fixes.

PR #642, opened 2020-01-08
Enabled filtering on created_at date for builds in api

amrohassaan:
Enabled filtering on created date for builds in rest api. This will be required for pulling specific builds from the client side reporting tool.

### DIFF
--- a/squad/api/rest.py
+++ b/squad/api/rest.py
@@ -82,7 +82,9 @@ class BuildFilter(filters.FilterSet):
     class Meta:
         model = Build
         fields = {'version': ['exact', 'in', 'startswith'],
-                  'id': ['exact']}
+                  'id': ['exact'],
+                  'created_at': ['exact', 'lt', 'lte', 'gt', 'gte'],
+                  }
 
 
 def unordered_build_queryset(request):
--- a/test/api/test_rest.py
+++ b/test/api/test_rest.py
@@ -446,6 +446,11 @@ class RestApiTest(APITestCase):
         data = self.hit('/api/builds/%d/testjobs/' % self.build.id)
         self.assertEqual(1, len(data['results']))
 
+    def test_build_filter(self):
+        created_at = str(self.build3.created_at.isoformat()).replace('+00:00', 'Z')
+        data = self.hit('/api/builds/?created_at=%s' % created_at)
+        self.assertEqual(1, len(data['results']))
+
     def test_testjob(self):
         data = self.hit('/api/testjobs/%d/' % self.testjob.id)
         self.assertEqual('myenv', data['environment'])


mwasilew:
LGTM. I'm not sure if the filter_override is needed. I had date filtering working without it. Since I don't have time to test it, could you try without it? Some unit tests would also be handy (I know it's PITA in case of REST API).

amrohassaan:
@mwasilew without the filter_override it works only for 'date' but not 'lte' and 'gte'. Not sure if those could be needed one day. If only 'date' would suffice, then we can remove that part.
Ack for the Unittests..will add those and repush.

mwasilew:
@amrohassaan ah, I missed that. The 'date' should be changed to 'exact'. This should do the trick and I think lte and gte should work this way.

amrohassaan:
@mwasilew:
http://localhost:8000/api/builds/?created_at__exact=2018-10-07 doesn't work, but
http://localhost:8000/api/builds/?created_at__date=2018-10-07 works. Sorry but not sure what you mean.

mwasilew:
I just tried and gt, gte, lt, lte work for me without the override. They require fancy datetime format though :( Example query that worked for me:
http://127.0.0.1:8000/api/builds/?created_at__gt=2019-10-27T21:17:28.467707Z

my patch:
```
diff --git a/squad/api/rest.py b/squad/api/rest.py
index 5c44cf5..7390712 100644
--- a/squad/api/rest.py
+++ b/squad/api/rest.py
@@ -82,7 +82,9 @@ class BuildFilter(filters.FilterSet):
     class Meta:
         model = Build
         fields = {'version': ['exact', 'in', 'startswith'],
-                  'id': ['exact']}
+                  'id': ['exact'],
+                  'created_at': ['exact', 'lt', 'lte', 'gt', 'gte'],
+                  'datetime': ['exact', 'lt', 'lte', 'gt', 'gte']}
```

amrohassaan:
That fancy date is what i wanted to avoid users having to pass. Complex IMHO. 
Tried it, all works but **not** exact. gte does the trick of grabbing the exact date plus whatever following dates. I just thought to make it easier on users passing dates not datetime or timestamp.

mwasilew:
you don't need to add any parameters with exact. Just use '='. What you're doing is traversing the object and that's not necessary in this case.

amrohassaan:
@mwasilew ah, got it. In this case, users will be expected ballpark the timestamp eg.  created_at__gt=2018-07-07T00:00:00Z. Ack.

mwasilew:
@amrohassaan this should be possible to customize. I'm trying to find a proper setting for DATETIME_INPUT_FORMAT.

amrohassaan:
@mwasilew 
Tried the following to no avail:
Added `created_at = filters.IsoDateTimeFilter()` to BuildFilter class and changed the BuildSerializer adding:
`created_at = IsoDateTimeField()
 created_at.input_formats = [IsoDateTimeField.ISO_8601] + [DateTimeField.input_formats]`
on time and then this 
`created_at = serializers.DateTimeField(input_formats=[IsoDateTimeField().ISO_8601, DateTimeField.input_formats])`
another time but still it doesn't accept any datetime input except the same one we are trying to avoid.
Not sure what I am missing there.

mwasilew:
@amrohassaan it might be just not working right. I think you're doing it OK. Since you're mainly concerned with library access I propose we stick to the 'long' datetime format and only add filtering capabilities. This way you can hide this from user by parsing input in the library. We can deal with direct human input later. How does this sound?

amrohassaan:
@mwasilew alright. sounds good.
FYI https://github.com/carltongibson/django-filter/issues/668
Dated 2017, talking about the same issue

PR #643, opened 2020-01-14
api: rest.py: enable filtering testruns by 'completed'

amrohassaan:
Enable filter testruns by 'completed' status. This is required to filter only meaningful testruns for downstream report processing.

### DIFF
--- a/squad/api/rest.py
+++ b/squad/api/rest.py
@@ -101,8 +101,9 @@ class TestRunFilter(filters.FilterSet):
         model = TestRun
         fields = {'job_id': ['exact', 'in', 'startswith'],
                   'job_status': ['exact', 'in', 'startswith'],
-                  'data_processed': ['exact', 'in'],
-                  'status_recorded': ['exact', 'in']}
+                  'data_processed': ['exact'],
+                  'status_recorded': ['exact'],
+                  'completed': ['exact']}
 
 
 class TestJobFilter(filters.FilterSet):
--- a/test/api/test_rest.py
+++ b/test/api/test_rest.py
@@ -23,11 +23,14 @@ class RestApiTest(APITestCase):
         self.build2 = self.project.builds.create(version='2', datetime=t2)
         t3 = timezone.make_aware(datetime.datetime(2018, 10, 3, 1, 0, 0))
         self.build3 = self.project.builds.create(version='3', datetime=t3)
+        t4 = timezone.make_aware(datetime.datetime(2018, 10, 4, 1, 0, 0))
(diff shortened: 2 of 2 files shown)


squad/api/rest.py
```diff
@@ -102,7 +102,8 @@ class Meta:
         fields = {'job_id': ['exact', 'in', 'startswith'],
                   'job_status': ['exact', 'in', 'startswith'],
                   'data_processed': ['exact', 'in'],
-                  'status_recorded': ['exact', 'in']}
+                  'status_recorded': ['exact', 'in'],
+                  'completed': ['exact', 'in']}
```
mwasilew: "in" makes little sense in this case as 'completed' can only have 2 values.

mwasilew:
You need to fix the number of builds in the tests.

squad/api/rest.py
amrohassaan: @mwasilew I thought exactly the same when adding this but i followed the pattern of the two booleans above it: 'data_processed' and 'status_processed'. Should I fix all  those three to be only 'exact' as well?

amrohassaan:
> You need to fix the number of builds in the tests.

not sure what you mean exactly @mwasilew. I created build4 and followed pattern and test passed. Can you explain more?

amrohassaan:
> You need to fix the number of builds in the tests.

@mwasilew yes, saw the Travis error. will fix that

squad/api/rest.py
mwasilew: I guess so. For Booleanfield there should only be 'exact'.

test/api/test_rest.py
```diff
@@ -468,6 +471,10 @@ def test_testruns(self):
         data = self.hit('/api/testruns/%d/' % self.testrun.id)
         self.assertEqual(self.testrun.id, data['id'])
 
+    def test_testruns_filter(self):
+        data = self.hit('/api/testruns/?completed=%s&build__id=%s' % (self.testrun.completed, self.build4.id))
```
chaws: You declared `self.testrun4` but used `self.testrun` instead. It worked because TestRun objects have `completed=True` by default (https://github.com/Linaro/squad/blob/master/squad/core/models.py#L609)

chaws:
Commit message: `s/api: rest.py:/api: rest:/` (just remove '.py' to harmonize with older commit messages :) )

test/api/test_rest.py
amrohassaan: Thanks @chaws for catching this. Fixed and repushed.

PR #644, opened 2020-01-16
frontend: project_settings: add advanced settings tab

chaws:
This will allow project admins to edit `project_settings` field in the self service menu, without having to go to Django admin page

### DIFF
--- a/squad/frontend/project_settings.py
+++ b/squad/frontend/project_settings.py
@@ -17,6 +17,12 @@ class ProjectForm(forms.ModelForm):
                   'notification_timeout', 'data_retention_days']
 
 
+class ProjectFormAdvanced(forms.ModelForm):
+    class Meta:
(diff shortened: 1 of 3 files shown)


squad/frontend/project_settings.py
```diff
@@ -35,6 +41,24 @@ def settings(request, group, project):
     return render(request, 'squad/project_settings/index.jinja2', context)
 
 
+@auth_write
+def advanced_settings(request, group, project):
+    if request.method == "POST":
+        form = ProjectFormAdvanced(request.POST, instance=request.project)
+        if form.is_valid():
```
mwasilew: will this use yaml_validator?

chaws: Yes, it throws a warning like `Dictionary object expected` when it doesn't match yaml syntax

PR #645, opened 2020-01-21
api, rest: enable displaying null instead of url for empty metrics on…

amrohassaan:
display null for empty 'metrics' on testruns endpoint instead of url that points to metrics with zero count.
This should save requests for empty metrics on clients attempting to generate reports.


test/api/test_rest.py
```diff
@@ -471,6 +471,14 @@ def test_testruns(self):
         data = self.hit('/api/testruns/%d/' % self.testrun.id)
         self.assertEqual(self.testrun.id, data['id'])
 
+    def test_testruns_null_metrics_attr(self):
```
chaws: Cool test. One more thing though: could you make one test where there are actual metrics for a testrun? This would test the other way around, when 'metrics' and 'metrics_file' attributes aren't `None`

squad/api/rest.py
```diff
@@ -762,15 +762,24 @@ class EnvironmentViewSet(ModelViewSet):
     ordering_fields = ('id', 'slug', 'name')
 
 
+class HyperlinkedMetricIdentityField(serializers.HyperlinkedIdentityField):
+    def get_url(self, *args):
+        testrun = args[0]
+        if testrun.metrics.count() == 0:
```
terceiro: it seems that this will produce one count(*) query for every test run in a list ...

squad/api/rest.py
```diff
@@ -762,15 +762,24 @@ class EnvironmentViewSet(ModelViewSet):
...
+        else:
+            return super().get_url(*args)
+
+
 class TestRunSerializer(serializers.HyperlinkedModelSerializer):
 
     id = serializers.IntegerField(read_only=True)
     tests_file = serializers.HyperlinkedIdentityField(view_name='testrun-tests-file')
-    metrics_file = serializers.HyperlinkedIdentityField(view_name='testrun-metrics-file')
+    metrics_file = HyperlinkedMetricIdentityField(view_name='testrun-metrics-file')
```
terceiro: any reason to not also do the same for `tests_file`? it seems inconsitent that one will always have the link while the other may not.

squad/api/rest.py
```diff
@@ -762,15 +762,24 @@ class EnvironmentViewSet(ModelViewSet):
...
     metadata_file = serializers.HyperlinkedIdentityField(view_name='testrun-metadata-file')
     log_file = serializers.HyperlinkedIdentityField(view_name='testrun-log-file')
     tests = serializers.HyperlinkedIdentityField(view_name='testrun-tests')
-    metrics = serializers.HyperlinkedIdentityField(view_name='testrun-metrics')
+    metrics = HyperlinkedMetricIdentityField(view_name='testrun-metrics')
```
terceiro: ditto for `tests`

squad/api/rest.py
amrohassaan: @terceiro yes was just talking this thought with @chaws today, the reason i held off, was thinking that no sensible testrun would carry no tests unless it is not completed. Any requestor for results for downstream analysis would most likely query completed ones ...**i think**..... We aren't sure! But it makes sense from uniformity and compliance PoV to have both. What do you think?

mwasilew: @amrohassaan there can be a testrun with only metrics and no tests.

squad/api/rest.py
amrohassaan: Done!

test/api/test_rest.py
amrohassaan: Done!

chaws:
Hey @amrohassaan , thanks for the updates. A few more items to comment:

1. @amrohassaan said:
> I have squished the fix for the missing 'has_mterics' for testrun in status table with the original PR for defaulting metrics and tests to null in testruns api view.

Cool! Almost a one-liner! Because old testruns's statuses might have `has_metrics=False` in production database, maybe it'd be a good idea to add a migration to fix that. I remember checking it in production database and it's just a couple of millions, it should be fairly fast to run that. Otherwise most testruns that do have metrics would still have `status.has_metrics=False` thus causing `null` for `metrics[_file]`. Also, it'd be good to have it as separate commits in this PR.


2. @amrohassaan said:
> Also ran black against test_rest.py to better align/format code.

I also agree that we should agree on a standard coding style, but the changes that **black** did on the code kinda caused some noise when I was reviewing the features you've implemented. it looks a bit out of scope of this PR, perhaps you can open another one changing the code project-wide, instead of just a single file.


3. @amrohassaan said:
> One thing i noticed with running tests is the reported regressions in db calls which I am not sure if a drastic issue or just trivial:
> ```
> DATABASE PERFORMANCE REGRESSIONS
> --------------------------------
> Unit: number of database queries
> 
> url:/api/builds/92/testruns/: 5 -> 13
> url:/api/testruns/292/: 1 -> 2
> url:/api/testruns/?completed=True&build__id=219: 1 -> 2
> url:/api/testruns/306/metrics/: 2 -> 3
> ```

* `url:/api/builds/92/testruns/: 5 -> 13`, if you look at line https://github.com/Linaro/squad/blob/master/squad/api/rest.py#L632, you'll see that it does `test_runs.order_by('-id')` (there is no prefetch_related to Status), and two lines after it uses `TestRunSerializer`, which checks for `testrun.status`.  It makes one extra query for each testrun belonging to that build. You can just copy the prefetch there.

* `url:/api/testruns/292/: 1 -> 2` | `url:/api/testruns/?completed=True&build__id=219: 1 -> 2` | `url:/api/testruns/306/metrics/: 2 -> 3`, that's due to the prefetch_related, it's OK

amrohassaan:
> Hey @amrohassaan , thanks for the updates. A few more items to comment:
> 
> 1. @amrohassaan said:
> 
> > I have squished the fix for the missing 'has_mterics' for testrun in status table with the original PR for defaulting metrics and tests to null in testruns api view.
> 
> Cool! Almost a one-liner! Because old testruns's statuses might have `has_metrics=False` in production database, maybe it'd be a good idea to add a migration to fix that. I remember checking it in production database and it's just a couple of millions, it should be fairly fast to run that. Otherwise most testruns that do have metrics would still have `status.has_metrics=False` thus causing `null` for `metrics[_file]`. Also, it'd be good to have it as separate commits in this PR.
> 
>@chaws migration added. I have ran the fix manually as in shell on my local dev and it took around ~ 15 min as said earlier on IRC. So applying this migration didn't do anything off course when i tried it locally. Please check it.
> 1. @amrohassaan said:
> 
> > Also ran black against test_rest.py to better align/format code.
> 
> I also agree that we should agree on a standard coding style, but the changes that **black** did on the code kinda caused some noise when I was reviewing the features you've implemented. it looks a bit out of scope of this PR, perhaps you can open another one changing the code project-wide, instead of just a single file.
> 
I've added separated the two topics each in its commit. One for the null metrics url in rest + its tests and the other for fixing the has_metrics for testrun in status table + its test + its migration. Did so coz it is the way recommended in Django docs to bundle migrations along with their changes for good practice.
> 1. @amrohassaan said:
> 
> > One thing i noticed with running tests is the reported regressions in db calls which I am not sure if a drastic issue or just trivial:
> > ```
> > DATABASE PERFORMANCE REGRESSIONS
> > --------------------------------
> > Unit: number of database queries
> > 
> > url:/api/builds/92/testruns/: 5 -> 13
> > url:/api/testruns/292/: 1 -> 2
> > url:/api/testruns/?completed=True&build__id=219: 1 -> 2
> > url:/api/testruns/306/metrics/: 2 -> 3
> > ```
> 
> * `url:/api/builds/92/testruns/: 5 -> 13`, if you look at line https://github.com/Linaro/squad/blob/master/squad/api/rest.py#L632, you'll see that it does `test_runs.order_by('-id')` (there is no prefetch_related to Status), and two lines after it uses `TestRunSerializer`, which checks for `testrun.status`.  It makes one extra query for each testrun belonging to that build. You can just copy the prefetch there.
> * `url:/api/testruns/292/: 1 -> 2` | `url:/api/testruns/?completed=True&build__id=219: 1 -> 2` | `url:/api/testruns/306/metrics/: 2 -> 3`, that's due to the prefetch_related, it's OK
Thank you! That is fixed now.

chaws:
Looks good to me

chaws:
Just for the record, running the migration to fix the has_metric flag shouldn't take too much time. This is what I got from production today:

```python
>>> s = Status.objects.filter(has_metrics='f', metrics_summary__gt=0, suite_id=None)
>>> s.count()
261292
```

PR #647, opened 2020-01-30
Allow users to switch test results table layout

chaws:
This was asked by @danrue and @mwasilew when the number of environments is too big (I'm assuming a number > 8) viewing suites in the build view gets hard.

I added an option so that the user can toggle the layout to this one below:
Screenshot from 2020-01-30 02-17-03

Clicking '+' a table like the one below is shown:
Screenshot from 2020-01-30 02-17-25

**NOTE 1** for builds with number of environments less than 8, the default behavior won't change. If that's more than 8, the 'box' view will kick in, but users can still click on the table layout to have the old style back. The opposite is valid, meaning if a build has less than 8 environments, users can simply click on 'box' view.

**NOTE 2** as requested by @danrue and @mrchapp , environments are now sorted in alphabetical order.

### DIFF
--- a/squad/frontend/templates/squad/_test_results_suitebox.jinja2
+++ b/squad/frontend/templates/squad/_test_results_suitebox.jinja2
@@ -1,0 +1,32 @@
+        {% for suite in test_results.suites %}
+        <table id='test-results-{{ suite.slug }}' class='test-results-box table table-bordered'>
+            <thead>
+                <tr>
+                    <th colspan="2">
+                        {{ suite }}
+                        {% set hide_suite = 'hide_' + suite.slug.replace('-', '_') -%}
+                        <a ng-click="{{ hide_suite }} = !{{ hide_suite }}" class="toggle-test-results">
+                            <i ng-hide="!{{ hide_suite }}" class="fa fa-minus"></i>
+                            <i ng-hide="{{ hide_suite }}" class="fa fa-plus"></i>
+                        </a>
+                    </th>
+                </tr>
+            </thead>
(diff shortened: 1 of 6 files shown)


chaws:
@mwasilew I've also added group by suite option and changed the way they're switched: 

Screenshot from 2020-02-06 15-03-30

Screenshot from 2020-02-06 15-09-37

danrue:
This looks like a really nice improvement. Thanks @chaws !

squad/frontend/templates/squad/_test_results_suitebox.jinja2
```diff
@@ -0,0 +1,32 @@
...
+                <tr>
+                    <th>{{ _('Environment') }}</th>
+                    <th>{{ _('Test Results') }}</th>
+                </tr>
+                {% for environment, statuses in suite.environments %}
+                <tr>
+                    <td>{{ environment }}</td>
+                    <td>
+                        {% for status in statuses %}
+                        <a href="{{testrun_suite_tests_url(project.group, project, build, status)}}">{% include "squad/_test_results_summary.jinja2" %}</a>
```
amrohassaan: @chaws, just for completion of data rendered, can we add {{ status.test_run.id }}  to this line similar to env view above?
It just feels more complete to have testrun_id also displayed even though it could  be one testrun per env. what do you think?

chaws: Good point! Pushed new updates that address your comment.

amrohassaan: Awseome, thanks @chaws.

PR #648, opened 2020-02-04
ci: tasks: avoid multiple submissions

chaws:
Because of the lab outage this morning, failed attempts to `submit` tasks are kept in the workers to be retried in one hour, which is the exact time for SQS to re-queue them (due to visibility timeout is also 1 hour). Several tasks got picked by multiple workers, thus submitting exact same tasks multiple times to labs.

This PR makes sure to not submit tasks if the testjob has already been submitted.


mwasilew:
This looks good. Could you also add a unit test to check this feature?

chaws:
@mwasilew I added a test that avoid multiple submissions, it covers outage cases as well

PR #649, opened 2020-02-10
Fix multiple submissions

chaws:
Enables SQUAD to add a suffix to queue names so that when deployed with SQS, it can use FIFO queues (FIFO queues MUST end with '.fifo' =/ ). 

We tested this in staging and it seemed to work


chaws:
@mwasilew changes in this PR are already running in staging, along with changes from in that PR: https://github.com/Linaro/qa-reports.linaro.org/pull/33

PR #650, opened 2020-02-14
Clear testjob failure

chaws:
If a test job submission fails, the error message is saved in `failure` attribute. But in a successful retry afterwards, the failure message was still going to show up.

This PR clears `failure` whenever a test job submission was successful.

Also, I changed the criteria do display the successful and error icons in the test job page. People would get confused to see the image below:

Screenshot from 2020-02-13 21-39-50

The two test jobs at the bottom had their submission failed in the first attempt, but they were successful later on, but the error icon would still be displayed, which is confusing.

### DIFF
--- a/squad/ci/tasks.py
+++ b/squad/ci/tasks.py
@@ -43,6 +43,7 @@ def submit(self, job_id):
     if not test_job.submitted:
         try:
             test_job.backend.submit(test_job)
+            test_job.failure = None
             test_job.save()
         except SubmissionIssue as issue:
             logger.error("submitting job %s to %s: %s" % (test_job.id, test_job.backend.name, str(issue)))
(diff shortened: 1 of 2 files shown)


mwasilew:
Overall looks good. How do we signal failure other than the job submission?

chaws:
The other exceptions are:
* during fetch (protocol issues): https://github.com/Linaro/squad/blob/master/squad/ci/tasks.py#L34
* during processing results: https://github.com/Linaro/squad/blob/master/squad/ci/models.py#L123
* failed lava tests: https://github.com/Linaro/squad/blob/master/squad/ci/backend/lava.py#L399
* plugins might also set it during post processing of the testjob

Hmm, I didn't think of the other errors, which makes sense to show the 'error' icon using the failure flag. I'll revert that part

PR #660, opened 2020-03-04
Add submit test results example with results

### LINKED ISSUE #656: submit api should probably accept json
The submit results API (documented at https://squad.readthedocs.io/en/latest/intro.html#result-submit-ref-label) is unnecessarily difficult to use with python's requests library for two reasons.

1. It seems to require multipart/form-data, which requests() implements with "files=".
2. It accepts a json-based testrun, but it throws the payload away.

The behavior can be demonstrated with the program below, which submits a test result using json, followed by the same data using multipart/form-data (files=).

The first submission shows up at https://staging-qa-reports.linaro.org/api/testruns/1114287/.
The second submission shows up at https://staging-qa-reports.linaro.org/api/testruns/1114288/.

It took me a while to figure this out, so I hope we can at least update the documentation with a requests() example, and investigate supporting the more native json-based payload.

```py
import json
import os
import requests
import time

url = "https://staging-qa-reports.linaro.org/api/submit/lkft/linux-stable-rc-5.5-oe/v5.5/i386"

metadata = {
    "job_id": f"{time.time()}",
    "make_kernelversion": "5.5.0",
}
log = "make output"
tests_file = {
    "build/build-arm64-gcc-9": "pass",  # or "fail"
}
headers = {
    "Auth-Token": os.environ["QA_TOKEN"],
}

result = requests.post(
    url,
    headers=headers,
    json={"metadata": metadata, "log": log, "tests_file": tests_file},
)

if not result.ok:
    print(f"Error submitting to qa-reports: {result.reason}: {result.text}")


metadata = {
    "job_id": f"{time.time()}",
    "make_kernelversion": "5.5.0",
}
result = requests.post(
    url,
    headers=headers,
    files={
        "tests": json.dumps(tests_file),
        "metadata": json.dumps(metadata),
        "log": json.dumps(log),
    },
)

if not result.ok:
    print(f"Error submitting to qa-reports: {result.reason}: {result.text}")
```

chaws:
This PR adds an example of submitting test results through the api using Python's requests library. Also, it fixes a bug that happened when a float value was passed as job_id.

Closes #656

### DIFF
--- a/doc/intro.rst
+++ b/doc/intro.rst
@@ -98,6 +98,26 @@ Example with test data as regular ``POST`` parameters::
         --form attachment=@/path/to/extra-info.txt \
         https://squad.example.com/api/submit/my-group/my-project/x.y.z/my-ci-env
 
+Example with test data using Python's requests library:
+
+.. code:: python
+    import json
+    import requests
+
+    tests = json.dumps({"test1": "pass", "test2": "fail"})
+    metrics = json.dumps({"metric1": 21, "metric2": 4})
+    metadata = json.dumps({"foo": "bar", "baz": "qux"})
+    log = 'log text ...'
+
+    headers = {"Auth-Token": 'xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx'}
+    url = 'https://squad.example.com/api/submit/my-group/my-project/x.y.z/my-ci-env'
+    data = {"metadata": metadata, "log": log, "tests": tests_file}
+
+    result = requests.post(url, headers=headers, data=data)
+    if not result.ok:
+        print(f"Error submitting to qa-reports: {result.reason}: {result.text}")
+
+
 Since test results should always come from automation systems, the API
 is the only way to submit results into the system. Even manual testing
 should be automated with a driver program that asks for user input, and
--- a/doc/intro.rst
+++ b/doc/intro.rst
@@ -77,7 +77,7 @@ the format of the data files.
 Example with test data as file uploads::
 
     $ curl \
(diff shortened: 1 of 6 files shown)


chaws:
Thanks for the feedback @danrue! I think having it more secure by default is a good practice as well, less chance of hard-coding stuff in.

I also changed the other examples that used `xxxx` elsewhere in docs.

doc/intro.rst
```diff
@@ -98,6 +98,26 @@ Example with test data as regular ``POST`` parameters::
...
 
+Example with test data using Python's requests library:
+
+.. code:: python
+    import json
+    import requests
+
+    tests = json.dumps({"test1": "pass", "test2": "fail"})
+    metrics = json.dumps({"metric1": 21, "metric2": 4})
+    metadata = json.dumps({"foo": "bar", "baz": "qux"})
```
mwasilew: I think metadata needs some keys we consider 'mandatory'

chaws: @mwasilew I reviewed the code that handles metadata, and no fields are actually mandatory. The only auto-filling we do is regarding job_id: https://github.com/Linaro/squad/blob/master/squad/core/tasks/__init__.py#L146

What we do have is `SPECIAL_FIELDS`: https://github.com/Linaro/squad/blob/master/squad/core/tasks/__init__.py#L117, but I don't think it's necessary to add them all to the examples. What do you think?

chaws:
@mwasilew I've rebased it with master.

doc/intro.rst
mwasilew: OK, let's not use special fields.

PR #661, opened 2020-03-06
Environments in project settings UI

chaws:
This enables handling environments in project settings UI tab. Here's a snapshot of how it looks like:
Screenshot from 2020-03-06 12-17-19

### DIFF
--- a/squad/frontend/project_settings.py
+++ b/squad/frontend/project_settings.py
@@ -3,7 +3,7 @@ from django.conf.urls import url
 from django.shortcuts import render, redirect, reverse
 from django.utils.translation import ugettext_lazy as N_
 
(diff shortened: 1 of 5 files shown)


squad/frontend/project_settings.py
```diff
@@ -74,6 +80,25 @@ def thresholds(request, group_slug, project_slug):
     return render(request, 'squad/project_settings/thresholds.jinja2', context)
 
 
+@auth_write
+def environments(request, group_slug, project_slug):
+    EnvironmentFormSet = forms.inlineformset_factory(Project, Environment, form=EnvironmentForm, extra=1)
```
mwasilew: do we want admins to have ability to add environments from UI?

PR #666, opened 2020-03-13
Add tests status summary to Testrun api

amrohassaan:
core: api: add tests status totals to testrun endpoint to be readily
available for client calls.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,11 +1,11 @@
 -r requirements.txt
 babel
-django_extensions
 flake8
 ipython
 pickleshare
 Werkzeug
 django-debug-toolbar
+django_extensions
 pytest
 pytest-django
 pytest-cov
--- a/squad/api/rest.py
+++ b/squad/api/rest.py
@@ -28,9 +28,12 @@ from squad.core.models import (
 )
 from squad.core.tasks import prepare_report, update_delayed_report
 from squad.ci.models import Backend, TestJob
+from squad.compat import drf_basename
 from django.http import HttpResponse
 from django.urls import reverse
 from django import forms
+from rest_framework_extensions.routers import ExtendedDefaultRouter
+from rest_framework_extensions.mixins import NestedViewSetMixin
 from rest_framework import routers, serializers, viewsets, status
 from rest_framework.decorators import action
 from rest_framework.exceptions import NotFound
@@ -241,13 +244,16 @@ class API(routers.APIRootView):
 
     * Client interaction is enabled with <a href="/api/schema/">/api/schema</a>
     URL.
+
+    * Testrun statuses are available at:
+        * `/testruns/<testrun_id>/status/`
     """
 
     def get_view_name(self):
         return "API"
 
 
-class APIRouter(routers.DefaultRouter):
+class APIRouter(ExtendedDefaultRouter):
 
     APIRootView = API
 
@@ -812,6 +818,30 @@ class HyperlinkedTestsIdentityField(serializers.HyperlinkedIdentityField):
             return None
 
 
+class StatusFilter(filters.FilterSet):
+    class Meta:
+        model = Status
+        fields = {'suite': ['exact', 'isnull'],
+                  'metrics_summary': ['gt', 'lt'],
+                  'tests_pass': ['gt', 'lt'],
+                  'tests_fail': ['gt', 'lt'],
+                  'tests_xfail': ['gt', 'lt'],
+                  'tests_skip': ['gt', 'lt'],
+                  'has_metrics': ['exact'], }
+
+
+class StatusSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Status
+        exclude = ('test_run',)
+
+
+class StatusViewSet(NestedViewSetMixin, viewsets.ReadOnlyModelViewSet):
+    queryset = Status.objects
+    serializer_class = StatusSerializer
+    filterset_class = StatusFilter
+
+
 class TestRunSerializer(serializers.HyperlinkedModelSerializer):
 
     id = serializers.IntegerField(read_only=True)
@@ -1112,7 +1142,12 @@ router.register(r'groups', GroupViewSet)
 router.register(r'projects', ProjectViewSet)
 router.register(r'builds', BuildViewSet)
 router.register(r'testjobs', TestJobViewSet)
-router.register(r'testruns', TestRunViewSet)
+router.register(r'testruns', TestRunViewSet).register(
+    r'status',
+    StatusViewSet,
+    parents_query_lookups=['test_run_id'],
+    **drf_basename('testrun-status'),
+)
 router.register(r'tests', TestViewSet)
 router.register(r'suites', SuiteViewSet)
 router.register(r'environments', EnvironmentViewSet)
--- a/squad/compat.py
+++ b/squad/compat.py
@@ -1,6 +1,9 @@
 """
 SQUAD compatibity file
 """
+from rest_framework_extensions import __version__ as DRFE_VERSION_STR
+
+DRFE_VERSION = [int(n) for n in DRFE_VERSION_STR.split(".")]
 
 # Handles compatibility for django_restframework_filters
 try:
@@ -12,3 +15,15 @@ try:
     from rest_framework_filters.backends import ComplexFilterBackend # noqa
 except ImportError:
     from squad.api.filters import ComplexFilterBackend # noqa
+
+
+def drf_basename(name):
+    """
+    Handles compatibility with different versions of djangorestframework, in
+    terms of the deprecation of `base_name` when registering ViewSets on DRF >=
+    3.10.
+    """
+    if DRFE_VERSION >= [0, 6]:
+        return {"basename": name}
+    else:
+        return {"base_name": name}
--- a/test/api/test_rest.py
+++ b/test/api/test_rest.py
@@ -3,7 +3,7 @@ import json
 from test.mock import patch
 from django.utils import timezone
 from squad.core import models
-from squad.core.tasks import UpdateProjectStatus, ReceiveTestRun
+from squad.core.tasks import UpdateProjectStatus, ReceiveTestRun, RecordTestRunStatus, ParseTestRunData
 from squad.ci import models as ci_models
 from rest_framework.authtoken.models import Token
 from rest_framework.reverse import reverse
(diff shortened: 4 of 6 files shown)


squad/api/rest.py
```diff
@@ -818,11 +818,28 @@ class TestRunSerializer(serializers.HyperlinkedModelSerializer):
     log_file = serializers.HyperlinkedIdentityField(view_name='testrun-log-file')
     tests = HyperlinkedTestsIdentityField(view_name='testrun-tests')
     metrics = HyperlinkedMetricsIdentityField(view_name='testrun-metrics')
+    status_summary = serializers.SerializerMethodField(read_only=True)
 
     class Meta:
         model = TestRun
         fields = '__all__'
 
+    def get_status_summary(self, obj):
```
mwasilew: IMHO this is a bad idea. The totals should be calculated for the model and the API should only expose them. With the proposed implementation, calculation will happen with each API call. Since TestRun object doesn't change after it's created, there is no need to do that.

mwasilew:
> You don't need to recalculate it. That status is already present in TestRun.status object. See https://github.com/Linaro/squad/blob/master/squad/api/rest.py

@chaws I don't see TestRun.status anywhere. Could you point me to the proper line?

chaws:
@mwasilew I think the line didn't go thru, but that's the one I was referring to: https://github.com/Linaro/squad/blob/master/squad/api/rest.py#L641

And this is the line where we already use TestRun.status to decide whether or nor `tests/metrics` field is displayed or not: https://github.com/Linaro/squad/blob/master/squad/api/rest.py#L804

For testrun case, I think it would look something like:

```python
    @action(detail=True, methods=['get'], suffix='status')
    def status(self, request, pk=None):
        try:
            status = self.get_object().status.all()[0]
            serializer = StatusSerializer(status, many=False, context={'request': request})
            return Response(serializer.data)
        except Status.DoesNotExist:
            raise NotFound()
```

mwasilew:
> @mwasilew I think the line didn't go thru, but that's the one I was referring to: https://github.com/Linaro/squad/blob/master/squad/api/rest.py#L641

this is ProjectStatus. It keeps summary data for the whole build. In case of TestRun we need totals for TestRun only.

> 
> And this is the line where we already use TestRun.status to decide whether or nor `tests/metrics` field is displayed or not: https://github.com/Linaro/squad/blob/master/squad/api/rest.py#L804
> 

this is artificial. There is no such field in TestRun object. In this code status is created directly from the JSON file containing all tests.

> For testrun case, I think it would look something like:
> 
> ```python
>     @action(detail=True, methods=['get'], suffix='status')
>     def status(self, request, pk=None):
>         try:
>             status = self.get_object().status.all()[0]

this won't work, there is no field 'status'
https://github.com/Linaro/squad/blob/master/squad/core/models.py#L591

>             serializer = StatusSerializer(status, many=False, context={'request': request})
>             return Response(serializer.data)
>         except Status.DoesNotExist:
>             raise NotFound()
> ```

amrohassaan:
> You don't need to recalculate it. That status is already present in TestRun.status object. See https://github.com/Linaro/squad/blob/master/squad/api/rest.py

@chaws the only place i saw this calculated is in [here](https://github.com/Linaro/squad/blob/7f862c9c4a548ede55d31b016d8d71ab3dd115ed/squad/core/tasks/__init__.py#L370) which actually saves it to db as well. I cannot reuse this code. 
The only status i see in rest.py is the projectstatus which tallies the totals for but for the entire build using the TestSummary.
I guess you mean to use this: https://github.com/Linaro/squad/blob/7f862c9c4a548ede55d31b016d8d71ab3dd115ed/squad/core/models.py#L880
I will try that now.

amrohassaan:
We are already prefetching the TestRun status https://github.com/Linaro/squad/blob/7f862c9c4a548ede55d31b016d8d71ab3dd115ed/squad/api/rest.py#L890
Somehow I didn't notice this last night :/
Fixing.

chaws:
@mwasilew @amrohassaan sorry, poor choice of words! 

TestRun has a foreign key in Status. There's one Status for each Suite that ran in a TestRun, then having `suite=None` in Status means overall summary of tests and metrics for that TestRun. 

@amrohassaan maybe we can change `Prefetch("status")` to `Prefetch('statuses')`, because although we're querying only one Status, the relation between these 2 models is a one-to-many.

squad/api/rest.py
```diff
@@ -799,16 +811,23 @@ def get_url(self, *args):
         testrun = args[0]
         statuses = testrun.status.all()
         if len(statuses) > 0:
-            tr_status = testrun.status.all()[0]
-            num_tests = (
-                tr_status.tests_pass + tr_status.tests_fail + tr_status.tests_skip + tr_status.tests_xfail
-            )
-            if num_tests > 0:
-                return super().get_url(*args)
+            for s in statuses:
```
chaws: I think it's fine querying only the first `status` like it used to be. Then I guess it makes sense to do:
```python
queryset = TestRun.objects.prefetch_related(	
        Prefetch("status", queryset=Status.objects.filter(suite=None))	
).order_by("-id")
```
instead of the new prefetch you implemented:

```python
queryset = TestRun.objects.prefetch_related("status").order_by("-id")
```

One less loop in the code :)

squad/api/rest.py
```diff
@@ -945,6 +964,12 @@ def metrics(self, request, pk=None):
         serializer = MetricSerializer(page, many=True, context={'request': request})
         return paginator.get_paginated_response(serializer.data)
 
+    @action(detail=True, methods=['get'], suffix='status')
+    def status(self, request, pk=None):
+        status = self.get_object().status.all()
```
chaws: Here is better to display only the status which, `suite=None`, I pulled your PR and it's showing all statuses by default.

Screenshot from 2020-03-18 09-58-22

amrohassaan: the reason i changed it to show all suites is what @mwasilew said last Monday when he said it would be better to query/filter by suite_id in some cases! Milosz?

mwasilew: The filter is wrong. By default you should be using get(suite=None). 'suite' should be an additional parameter of this endpoint and only if suite is present it should be used to filter the proper data. This is exactly what I said.

amrohassaan: @chaws @mwasilew reset and repushed.

squad/api/rest.py
```diff
@@ -784,29 +795,36 @@ class EnvironmentViewSet(ModelViewSet):
...
     def get_url(self, *args):
         testrun = args[0]
         statuses = testrun.status.all()
         if len(statuses) > 0:
             if testrun.status.all()[0].has_metrics:
                 return super().get_url(*args)
-        else:
-            return None
+            else:
+                return None
```
mwasilew: hmm, doesn't return anything if len(statuses) == 0 ?

squad/api/rest.py
```diff
@@ -784,29 +795,36 @@ class EnvironmentViewSet(ModelViewSet):
...
-            num_tests = (
-                tr_status.tests_pass + tr_status.tests_fail + tr_status.tests_skip + tr_status.tests_xfail
-            )
+            num_tests = (tr_status.tests_pass + tr_status.tests_fail + tr_status.tests_skip + tr_status.tests_xfail)
             if num_tests > 0:
                 return super().get_url(*args)
-        else:
-            return None
+            else:
+                return None
```
mwasilew: ditto

squad/api/rest.py
mwasilew: how do I pass additional parameters to status() ?

squad/api/rest.py
amrohassaan: @mwasilew my mistake ..added extra indentation.

amrohassaan:
@mwasilew, @chaws 
This will display all the list of statuses by default and will filter by suite is. eg: testrun/#/status/?suite=#
Only caveat is the custom code to filter the list won't filter for suite=null, rather will display all the list.
This still missing fixing the test. Please review first and let me know.

chaws:
It's fine for me.

squad/api/rest.py
```diff
@@ -784,31 +785,60 @@ class EnvironmentViewSet(ModelViewSet):
...
+                    num_tests = (s.tests_pass + s.tests_fail + s.tests_skip + s.tests_xfail)
+                    if num_tests > 0:
+                        return super().get_url(*args)
+                    else:
+                        return None
         else:
             return None
 
 
+class StatusFilteredListSerializer(serializers.ListSerializer):
```
mwasilew: maybe we should use drf-extensions instead? This implementation looks a bit overcomplicated to me. drf-extensions: https://github.com/chibisov/drf-extensions

dev-docker
```diff
@@ -11,7 +11,7 @@ mkdir -p "$datadir"
 dockerfile="$datadir"/../Dockerfile.dev
 (
   sed -e '1,/WORKDIR/!d' "$basedir"/Dockerfile
-  echo 'RUN apt-get install -q2 snakefood flake8 git python3-wheel python3-django-extensions ipython3 rabbitmq-server chromium nodejs ruby-foreman procps && \'
+  echo 'RUN apt-get install -q2 snakefood flake8 git python3-wheel python3-django-extensions python3-djangorestframework-extensions ipython3 rabbitmq-server chromium nodejs ruby-foreman procps && \'
```
chaws, line 14: Dependencies here are just when running tests in docker. What you added should not be a dev thing, which you did correctly above adding to Dockerfile. Adding here is redundant

requirements-dev.txt
```diff
@@ -1,11 +1,11 @@
 -r requirements.txt
 babel
-django_extensions
 flake8
 ipython
 pickleshare
 Werkzeug
 django-debug-toolbar
+drf-extensions
```
chaws, line 8: I think you wanted to keep 'django_extensions' here, and leave 'drf-extensions' in requirements.txt

squad/compat.py
```diff
@@ -1,6 +1,11 @@
 """
 SQUAD compatibity file
 """
+from django.utils.version import get_version
+from rest_framework_extensions import __version__ as DRFE_VERSION_STR
+
+DJANGO_VERSION = get_version()
```
chaws, line 7: I ran `grep -rn DJANGO_VERSION squad/` and found no use of `DJANGO_VERSION`

chaws: The same for `get_version`, imported above

test/api/test_rest.py
```diff
@@ -522,6 +522,17 @@ def test_testruns_metrics(self):
         data = self.hit('/api/testruns/%d/metrics/' % self.testrun.id)
         self.assertEqual(list, type(data['results']))
 
+    def test_testruns_status(self):
```
chaws, line 525: Not to be too picky, but could you please add a test where you test Status' filtering?

chaws:
All good now, merging

PR #668, opened 2020-03-19
core: models, rest, add env to MetricThreshold

amrohassaan:
Add environment to MetricThreshold to replace project as FK. This is the
first commit and individual pr of a set of 3. Migrating the data will take place in 2nd PR and dropping the project will be in last PR. This was requested by @mwasilew to avoid any downtime when rolling this change to production.
Travis my report errors in subsequent PRs since all 3 commits will be contained in different PR each.


mwasilew:
@amrohassaan 1st PR should only create the field but not use it. IMHO most of this patch should be moved to the last PR in the series.

amrohassaan:
> @amrohassaan 1st PR should only create the field but not use it. IMHO most of this patch should be moved to the last PR in the series.
I was thinking the same honestly while doing this but wasn't quite sure. Will do this.

mwasilew:
I was wrong, most of the code should go to the middle patch in the series. As mentioned on IRC, it should work like this:
1. add field but not use it
2. start using it
3. remove old field

amrohassaan:
As mentioned on IRC, following this [reference](https://gist.github.com/majackson/493c3d6d4476914ca9da63f84247407b) on a local fresh db, the [first point](https://gist.github.com/majackson/493c3d6d4476914ca9da63f84247407b#adding-a-nullable-field-or-a-new-table) which breaks the first PR into two, works fine. ie: running migration before rolling out changes to model. That said, I still think we can do this in max of 3 PRs.

chaws:
LGTM, merging it now and rolling it out later today

PR #669, opened 2020-03-19
core: migrations, migrate data MetricThreshold

amrohassaan:
Migrate data from project to env. This is the second commit and individual pr of a set of 3.
The final commit/pr will remove project.
This won't pass travis check. Unless [First PR](https://github.com/Linaro/squad/pull/668) is run first.


squad/core/migrations/0127_metric_thresholds_migrate_data.py
```diff
@@ -0,0 +1,33 @@
+# Generated by Django 2.2.6 on 2020-02-26 14:42
+
+from django.db import migrations
+
+
+def populate_env_in_thresholds_and_migrate_olddata(apps, schema_editor):
+    Threshold = apps.get_model('core', 'MetricThreshold')
+    Environment = apps.get_model('core', 'Environment')
+    existing_thresholds = Threshold.objects.all().iterator()
```
chaws: Do you really need the `.iterator()`? I think up to `.all()` is OK

squad/core/models.py
```diff
@@ -1058,20 +1058,31 @@ def get_fixes(self):
...
+            thresholds = MetricThreshold.objects.filter(
+                environment__in=environments_qs.values_list('id', flat=True)
+            )
+            thresholds_names = thresholds.values_list('name', flat=True)
             for metric in Metric.objects.filter(
                     Q(test_run__in=test_runs) | Q(suite__in=suites),
-                    name__in=self.build.project.metricthreshold_set.values_list('name', flat=True)):
-                for threshold in self.build.project.metricthreshold_set.all():
+                    name__in=thresholds_names):
+                for threshold in thresholds:
```
chaws: OK, This is a bit confusing when adding environments, but I think we can solve that changing very little code:

Although you added environment to MetricThreshold, you're still filtering metrics that exceeded in a project's threshold.

What you need to do is check if the Metric's TestRun's environment matches the environment in the threshold. something like that:
```
for metric in Metric.objects.filter(...): # this filtering is OK
    for threshold in thresholds:
        if metric.test_run.environment_id != threshold.environment_id:
            continue # this threshold doesn't apply to this metric 
```

But this will imply that you should have `thresholds` an iterator of objects, instead of just a list of metric threshold names.

Does it make sense?

squad/core/models.py
```diff
@@ -1058,20 +1058,31 @@ def get_fixes(self):
     def get_exceeded_thresholds(self):
         # Return a list of all (threshold, metric) objects for those
         # thresholds that were exceeded by corresponding metrics.
+        environments_qs = self.build.project.environments.all()
         thresholds_exceeded = []
         if self.has_metrics:
             test_runs = self.build.test_runs.all()
             suites = Suite.objects.filter(test__test_run__build=self.build)
+            thresholds = MetricThreshold.objects.filter(
+                environment__in=environments_qs.values_list('id', flat=True)
```
chaws: You're making two queries here:
1. querying all environments for this project
2. gathering all ids and passing it as `environment_id IN (...)` in the metricThreshold query

You can make it a single, an faster, query by doing:

```python
thresholds = MetricThreshold.objects.filter(environment__in=self.build.project.environments.all())
```

squad/core/models.py
```diff
@@ -1058,20 +1058,31 @@ def get_fixes(self):
...
                     Q(test_run__in=test_runs) | Q(suite__in=suites),
-                    name__in=self.build.project.metricthreshold_set.values_list('name', flat=True)):
-                for threshold in self.build.project.metricthreshold_set.all():
+                    name__in=thresholds_names):
+                for threshold in thresholds:
                     if threshold.is_higher_better:
                         if metric.result < threshold.value:
-                            thresholds_exceeded.append((threshold, metric))
+                            thresholds_exceeded.append((
+                                environments_qs.get(id=threshold.environment_id),
```
chaws: Do you really need to have environment added to this tuple? I you can simply fetch it using `threshold.environment` later on

test/core/test_project_status.py
```diff
@@ -288,13 +288,13 @@ def test_get_exceeded_thresholds(self):
         testrun.metrics.create(name='metric1', suite=self.suite, result=5)
         status = ProjectStatus.create_or_update(build)
         MetricThreshold.objects.create(project=self.project,
+                                       environment=self.environment,
```
chaws: I think there can have a test with two metricthresholds. This means two environments therefore two testruns and metrics. This way you can test that a metric exceeded in one environment but didn't for another environment.

squad/core/templates/squad/notification/diff.txt.jinja2
```diff
@@ -34,9 +34,10 @@ Failures
 Threshold(s) exceeded
 ---------------------
 {% for threshold_metric in thresholds %}
-Metric name: {{threshold_metric[1].name}}
-Metric result: {{threshold_metric[1].result}}
-Threshold value: {{threshold_metric[0].value}}
+Environment: {{threshold_metric[0].slug}}
```
chaws: Continuing my comment above, here instead of having `threshold_metric[0].slug` you can have `threshold_metric[2].environment.slug`, this is, removing `environment` from the tuple, the indexes would change accordingly.

amrohassaan:
@chaws all fixed.

squad/core/models.py
```diff
@@ -1059,19 +1060,25 @@ def get_exceeded_thresholds(self):
...
+                for threshold in thresholds:
+                    if metric.test_run.environment_id != threshold.environment_id:
+                        continue
                     if threshold.is_higher_better:
                         if metric.result < threshold.value:
                             thresholds_exceeded.append((threshold, metric))
-                    else:
-                        if metric.result > threshold.value:
-                            thresholds_exceeded.append((threshold, metric))
+                        else:
```
chaws: I think you indented this `else` by mistake. It used to the the else for `if threshold.is_higher_better`, but now it's else'ing `if metric.result < threshold.value`

test/core/test_project_status.py
```diff
@@ -281,20 +283,34 @@ def test_cache_fixes_update(self):
...
+
         status = ProjectStatus.create_or_update(build)
         MetricThreshold.objects.create(project=self.project,
-                                       name='metric1', value=4,
+                                       environment=self.environment,
+                                       name='/metric2', value=4,
                                        is_higher_better=True)
-
         thresholds = status.get_exceeded_thresholds()
+        self.assertEqual(len(thresholds), 0)
```
chaws: This was supposed to be 1, no? Since metric threshold has `value=4, is_higher_better=True, name='/metric2'` and there's a testrun with a metric like `name='metric2', suite='/', result=2` 

Does it make sense?

squad/core/models.py
amrohassaan: Makes sense, fixed this. The semantics is bit confusing there for a sec.

test/core/test_project_status.py
amrohassaan: @chaws I've changed this to conform with what @terceiro explained about the threshold. Two cases, one with higher_better true then false.

chaws:
Looking good. Merging.

PR #670, opened 2020-03-19
core: model,migration:delete project from threshold

amrohassaan:
Delete project from MetricThreshold. This is the final commit and
individual pr of a set of 3.
This won't pass travis check. Unless [first PR](https://github.com/Linaro/squad/pull/668) & [second PR](https://github.com/Linaro/squad/pull/669) rolled out first.


squad/core/migrations/0128_metric_thresholds_remove_proj_col.py
```diff
@@ -0,0 +1,28 @@
...
+
+from django.db import migrations, models
+import django.db.models.deletion
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('core', '0099_metricthreshold'),
+        ('core', '0127_metric_thresholds_migrate_data'),
```
chaws: I think the reason for having multiple dependencies is if those are from different modules, since both of these are from `core`, the second dependency will probably have a cascading dependency up to `0099_metricthreshold`. I think it's safe to remove the first dependency here `0099...`

amrohassaan: @chaws true..i tested this yesterday. I will remove this. Just added it initially to be on the safe side since i read it on a resource online saying constraints in db would stick around unless your add this. But seems it was an outdated post.

chaws:
Looks good, I'll merge it and make a release tomorrow

PR #671, opened 2020-03-20
api: rest: show testrun id in test endpoint

amrohassaan:
rest.py: show testrun id conditionally in api depending on calling
endpoint. Testrun id shows in test main ep while not from nested Test ep. AFAIK, the main hold back from having testrun id show in tests main endpoint was regression in db queries. Tests on this PR shows 0 regressions.

### DIFF
--- a/squad/api/rest.py
+++ b/squad/api/rest.py
@@ -849,13 +849,22 @@ class SuiteViewSet(viewsets.ModelViewSet):
 
 
 class TestSerializer(serializers.ModelSerializer):
+
+    def __init__(self, *args, **kwargs):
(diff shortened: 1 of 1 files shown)


amrohassaan:
@mwasilew FYI for eval, this is the detour I talked to you about yesterday.

mwasilew:
Looks OK. Did you check how it impact the number of queries?

amrohassaan:
> Looks OK. Did you check how it impact the number of queries?

Using Django Debug Toolbar. I see same exact number of queries as before. Test also reported nothing.

PR #672, opened 2020-03-23
Avoid sending emails when body >1MB

chaws:
This morning we got an error saying that one email got too big (>10MB) and wasn't able to deliver to recipients.

This patch checks if email content is >1MB and send a link to a build's email directly from the api. This email is a DelayedReport, and will be deleted in 1 day after its creation.

### DIFF
--- a/squad/core/notification.py
+++ b/squad/core/notification.py
@@ -2,6 +2,7 @@ from collections import OrderedDict
 from squad.mail import Message
 from django.conf import settings
 import django.template
+import logging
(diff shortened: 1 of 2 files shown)


mwasilew:
Maybe in case of too big email notification the DelayedReport object should live longer (a week?).

chaws:
@mwasilew makes sense, yeah. I've changed the link to that when accessed, will create a report with `data_retention_period=7`

PR #675, opened 2020-04-01
core: models fix metricthreshold unqiue contraint

amrohassaan:
This is to fix unique together constraint in MetricThreshold model. Env should replace project to avoid duplicate ids when inserting two envs for a project.


amrohassaan:
PUshing new one and closing this

chaws:
Looks good, merging

Issue #674, opened 2020-04-01
testjob is being duplicated


chaws:
This morning there were a lot of errors regarding duplicated test jobs.
Here are some examples (i don't know which project and builds those belong):
```
Failed to fetch test_job(2677999): "There is already a test run with job_id 1977575"
Failed to fetch test_job(2674408): "There is already a test run with job_id 1977391"
```

PR #677, opened 2020-04-02
Reduce notification tasks

chaws:
This patch limits the number of `notification_timeout` tasks to a single one.
    
Before, every new testrun for a given build would call up the task `maybe_notify_project_status` and therefore a new task for `notification_timeout` would be added to the queue.

Problem is that if a project has hundreds of testruns, this means hundreds of the exact same call to `notification_timeout`. Because that task forces sending notification, it'll invalidate all subsequent attempts to send the same notification.


squad/core/tasks/notification.py
```diff
@@ -9,13 +9,30 @@
...
     build = projectstatus.build
     project = build.project
 
-    timeout = project.notification_timeout
-    if timeout:
-        notification_timeout.apply_async(args=[status_id], countdown=timeout)
+    if projectstatus.notified_on_timeout is None and project.notification_timeout:
+        notification_timeout.apply_async(args=[status_id], countdown=project.notification_timeout)
+        projectstatus.notified_on_timeout = False
+        projectstatus.save()
```
terceiro, line 35: there is a race condition between fetching `projectstatus` from the DB and getting `notified_on_timeout is None`, and saving `notified_on_timeout = False`. you probably want to wrap this  in an atomic transation.

chaws: You're right. I think you suggested an atomic transaction on something else also related to tasks. I'll try to come up with it so that this issue can be addressed as well: https://github.com/Linaro/squad/issues/674

chaws:
@terceiro I managed to get this condition tested using threads and [TransactionTestCase](https://docs.djangoproject.com/en/3.0/topics/testing/tools/#transactiontestcase). It turns out that django already has [select_for_update](https://docs.djangoproject.com/en/3.0/ref/models/querysets/#select-for-update) that needs to be wrapped around a [transaction.atomic](https://docs.djangoproject.com/en/3.0/topics/db/transactions/#controlling-transactions-explicitly) statement.

Only problem is that tests fail for sqlite because it doesn't support select_for_update. [Documentation](https://docs.djangoproject.com/en/3.0/ref/databases/#database-is-locked-errors) says we should increase timeout, but that didn't work at all. I ended up creating a tag to the test that has a race condition so we disable it when running on sqlite.

What do you think of that?

squad/manage.py
```diff
@@ -7,6 +7,10 @@ def main():
     if len(sys.argv) > 1 and sys.argv[1] == 'test':
         os.environ.setdefault("DJANGO_SETTINGS_MODULE", "test.settings")
         testing = True
+
+        # skip tests that would raise exception in sqlite
+        if os.getenv('DATABASE') is None or os.getenv('DATABASE') == '':
```
terceiro, line 12: you can set DATABASE to explicitly use sqlite. 

I suggest using a variable called, say, EXCLUDE_TAG, and use it's value. so this can be fully controlled from e.g. the travis config

.travis.yml
```diff
@@ -21,7 +21,7 @@ env:
   - SCRIPT="./scripts/test-docker"
   - SCRIPT="./manage.py compilemessages" INSTALLPIP=yes
   - SCRIPT="docker run -v $(pwd):/squad --workdir /squad fsfe/reuse reuse lint"
-  - SCRIPT="python3 -m pytest" INSTALLPIP=yes
+  - SCRIPT="python3 -m pytest" DATABASE=ENGINE=django.db.backends.postgresql_psycopg2:NAME=squad:USER=postgres:PASSWORD=squad:HOST=127.0.0.1:PORT=5432 INSTALLPIP=yes
```
terceiro, line 24: why are you only changing this test, but not the others? (sqlite is the default after all)

chaws: This one uses pytest, and thought it would be easier to run pytest against postgres than find out how to exclude that test in pytest

squad/manage.py
chaws: I wanted `./manage.py test` to run against postgres by default, since that's our production's. OTOH testing in sqlite3, like we also do is interesting so that SQUAD can have a "testing" db.

I understood `EXCLUDE_TAG`, let me work on that.

PR #679, opened 2020-04-07
frontend: add badge for build

mwasilew:
The concept of project badge was extended to build. With this commit
each build can produce a badge that can be included in your build
system.

### DIFF
--- a/squad/frontend/views.py
+++ b/squad/frontend/views.py
@@ -101,19 +101,7 @@ def project_home(request, group_slug, project_slug):
     return render(request, 'squad/project.jinja2', context)
 
 
-@auth
-def project_badge(request, group_slug, project_slug):
-    project = request.project
-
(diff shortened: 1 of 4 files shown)


squad/frontend/views.py
```diff
@@ -176,6 +164,49 @@ def project_badge(request, group_slug, project_slug):
     return HttpResponse(badge, content_type="image/svg+xml")
 
 
+def __badge_text(request):
```
chaws, line 167: Did you mean "__badge_style()"?

chaws:
Looking good! I finally looked at the code that generates badges, was curious for a while :)

squad/frontend/views.py
mwasilew: yes, we can call it that. I'll change it.

PR #681, opened 2020-04-07
doc: fix python example for submissions

chaws:
Fix a line break that prevented python example from being rendered in docs


chaws:
@mwasilew after looking at https://github.com/Linaro/squad/commit/8a4ef7edb00916a49dcbec6dbbf83f1ccd1e2fb4, I realized that "job_id" is mandatory for submissions but not for testruns generated within squad (https://github.com/Linaro/squad/commit/61f72150aec42ac43b8adc1a0ee07b62a60894ef). I got confused there! I mentioned it in docs as well

Issue #673, opened 2020-03-25
Add metrics top-level endpoint identical to Tests


amrohassaan:
For efficiency in extracting and reporting, metrics should have a dedicated top-level endpoint similar to Tests. This became very evident when using the client. One shouldn't have to query Testruns to get to metrics. This will help maintaining abstract design and light weight on client as well.

mwasilew:
Metrics have stricter requirements for access control. Please keep that in mind when writing the code.

PR #685, opened 2020-04-12
core: rest: add top-level metric endpoint

amrohassaan:
Add top level metric ep similar to tests.


chaws:
I fetched your branch locally and it worked just fine :)
Only request I make is to add a test for this in `test/api/test_rest.py`

amrohassaan:
Tests in.

amrohassaan:
Closes issue #673

chaws:
Looks good to me!

PR #687, opened 2020-04-13
core: history: check for empty suite metadata

chaws:
Fix this error:

```
AttributeError at /lkft/linux-stable-rc-5.5-oe/tests/linux-log-parser/check-kernel-warning-1360505
'NoneType' object has no attribute 'instructions_to_reproduce'
```


chaws:
Merging, I'll update production to fix those

Issue #503, opened 2019-04-01
force not working with report API


danrue:
The following example url should generate a new report every time, but it does not. 'force' does seem to work with /email, but not with /report.

https://qa-reports.linaro.org/api/builds/14837/report/?template=15&force=true

mwasilew:
@danrue I'm trying to reproduce it and I can't. What makes you think that the force doesn't work with /report?

mwasilew:
We agreed that the fix for this will be creating 'last_updated' field in the Report object.

danrue:
Hit this again today. It's not high priority, but would be nice to have it addressed eventually.

chaws:
@danrue does this still happen?

danrue:
hi @chaws - yes it's still a problem.

If you hit the url given above, it will provide a link to a report. The report itself will have a created_at field.

If you hit the link again, asking for a new report to be generated, it will give the same url, with the same created_at field.

How does the user know if the report was actually re-generated, or not?

mwasilew:
@chaws I guess the proper solution here is to update created_at (not sure if this is possible) and delete status_code so when new report is created these can indicate that 'force' worked.

danrue:
Previously it was suggested to create 'last_updated'. Another suggestion that's a little easier for the client/programmer is to create a new report url when force is given. That's what I would have expected as a user, rather than modifying reports.

mwasilew:
Creating new report is probably the easiest solution to implement.

chaws:
Ok, I opened a PR that changes it so that every call with `force=true` actually does create a new report. It will get you a new link to every F5 on that page.

Oh, it's also worth mentioning that if one hits that `force=true` many times, the regular `/report` (with no force) will always point to the most updated report.

PR #689, opened 2020-04-14
api: rest: fix force-created reports

chaws:
For every call to `/api/builds/id/report/?force=true` a new report will be created.

Close https://github.com/Linaro/squad/issues/503


mwasilew:
I think 'wait' parameter meaning is reversed. When 'wait' is true the call returns immediately, when 'wait' is false it waits for the report to be generated. This isn't very intuitive to me. Could you also document what 'wait' does?

chaws:
You're right, it's the opposite. I'll fix it

PR #691, opened 2020-04-17
[WIP] doc: A couple of typo fixes and requests for clarification

pfalcon:
1. Fix a couple of genuine typos.
2. Clarify a few points which I think I could myself (need confirmation from maintainers).
3. Pinpoint even more points which could provide a bit more information - these would need input from maintainers.

### DIFF
--- a/doc/intro.rst
+++ b/doc/intro.rst
@@ -28,7 +28,7 @@ Suite, which is a basically used to group and analyze results together.
 Every test suite must be associated with exactly one Environment, which
 describes the environment in which the tests were executed, such as
 hardware platform, hardware configuration, OS, build settings (e.g.
-regular compilers vcs optimized compilers), etc. Results are always
+regular compilers vs optimized compilers), etc. Results are always
 organized by environments, so we can compare apples to apples.
 
 Projects can have subscriptions, which are either users or manually-entered
--- a/doc/intro.rst
+++ b/doc/intro.rst
@@ -31,7 +31,7 @@ hardware platform, hardware configuration, OS, build settings (e.g.
 regular compilers vs optimized compilers), etc. Results are always
 organized by environments, so we can compare apples to apples.
 
-Projects can have subscriptions, which are either users or manually-entered
+Projects can have subscriptions, which are either users registered in the system or manually-entered
 email addreses that should be notified about important events such as changing
 test results. ProjectStatus records the most recent build of a project, against
 which future results should be compared in search for important events to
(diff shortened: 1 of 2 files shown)


pfalcon:
@mwasilew, @chaws: Here's what I accumulated in my local copy from so far rather cursory reading of the docs. I decided to share as WIP to seek whether this activity is useful (trying to clarify more points, hopefully fixing typos is w/o questions). Let me know.

doc/intro.rst
```diff
@@ -176,15 +176,15 @@ storing test log in the Test object, passed JSON file can look as follows:
 
 Both forms are supported. In case log entry is missing or simple JSON
 format is used, logs for each Test object are empty. They can be filled
-in using plugins.
+in using plugins [at later time??].
```
chaws, line 179: @pfalcon yes! After results are fetched, [enabled plugins](https://github.com/Linaro/squadplugins) will post-process tests seeking for extra logs. One example is a built-in plugin `linux-log-parser`, which searches job log for linux kernel oopsies, then create tests for each specific type of oops (kernel-panic, warning, bug, etc)

doc/intro.rst
```diff
@@ -224,7 +224,9 @@ strings. The following fields are recognized:
 * ``job_status``: string identifying the status of the project. SQUAD
   makes no judgement about its value.
 * ``job_url``: URL pointing to the original test run.
-* ``resubmit_url``: URL that can be used to resubmit the test run.
+* ``resubmit_url``: URL that can be used to resubmit the test run. [As in: this is
+  API-level URL, which, being accessed, would trigged a resubmit (auth is an obvious concern),
+  or this is UI-level page, which user can open, login if needed and press "Resubmit" button?]
```
chaws, line 229: This is actually the URL that was used to resubmit a job. For instance, if you visit a build's testjob tab, you get a list of testjobs and a blue button next to it that allow you to resubmit that job. Then the resubmit-url is that new url, mostly with a new job id in it.

pfalcon:
@chaws: Thanks for your comments! Will try to integrate them into next version of the patch. (But currently switching to other tasks, so that may take some time.)

doc/intro.rst
mwasilew: I propose to change it like this:
"They can be filled in using plugins in a post-processing step. This happens after all results are collected in SQUAD"

doc/intro.rst
```diff
@@ -176,15 +176,15 @@ storing test log in the Test object, passed JSON file can look as follows:
...
+in using plugins [at later time??].
 
 Metrics
 ~~~~~~~
 
 Metrics must be posted as JSON, encoded in UTF-8. The JSON data must be
 a hash (an object, strictly speaking). Metric names go in the keys, and
 values must be either a single number, or an array of numbers. In the
-case of an array of numbers, then their mean will be used as the metric
+case of an array of numbers, then their mean [or average? elaborate] will be used as the metric
```
mwasilew, line 187: It's mean: https://docs.python.org/3/library/statistics.html#statistics.mean

PR #705, opened 2020-04-27
ci: fix Backend.fetch() implementation

mwasilew:
In case the job is still in the queue or is in progress, fetch() raises
TemporaryFetchIssue from within block that catches FetchIssue. Since
TemporaryFetchIssue derives from FetchIssue, the exception is
immediately consumed and fetch_attempts counter is increased. This leads
to maxing out fetch_counter for jobs that stay in the queue for a long
time. This patch removes the exception and instead returns from the
fetch() function without an error. This doesn't result in increasing
fetch_attempt counter.


chaws:
Really good catch!!

PR #706, opened 2020-04-27
ci: models: handle race condition without wait

chaws:


mwasilew:
Let's test it in staging before rolling out to production.

chaws:
Indeed. I was monitoring the queue until late yesterday, some tasks to over 1h to complete. I'm going to try to submit the same type of jobs in staging and see what happens

Issue #709, opened 2020-04-29
Seemingly incorrect number of projects per group is shown


pfalcon:
At https://qa-reports.linaro.org/ , I see that my group is shown as:

> Paul Sokolovsky's group
Paul Sokolovsky's qa-reports group, mostly for testing.
4 projects

However, going inside it, https://qa-reports.linaro.org/pfalcon/, there're only 2 projects (and none archived).

pfalcon:
I decided to submit this, before creating new projects in there. Let me also provide screenshots of the current state:

![Screenshot at 2020-04-29 14-11-50](https://user-images.githubusercontent.com/500451/80589757-7898f500-8a23-11ea-89cb-e2e61b8999b9.png)

---

![Screenshot at 2020-04-29 14-12-14](https://user-images.githubusercontent.com/500451/80589768-7fc00300-8a23-11ea-9acb-7ff5a0c4b608.png)

pfalcon:
After creating 3rd project, the count on front page shows "6". With 4th, it shows "8".

mwasilew:
@pfalcon this is weird. I can see 4 projects in your group: zephyr-net, manual-test, zephyr-jerryscript, zephyr-micropython. However main page only shows '2' next to your name. In other words I have the opposite situation. Let me check what might be going on.

pfalcon:
> I can see 4 projects in your group: zephyr-net, manual-test, zephyr-jerryscript, zephyr-micropython. 

Yep, just created 2 more (apparently we posted comments in about the same time).

mwasilew:
Ah, in this case my numbers look OK when I'm logged in. When I'm not logged in I'm getting the same problem.

pfalcon:
> Ah, in this case my numbers look OK when I'm logged in. When I'm not logged in I'm getting the same problem.

For me the situation is, with current 4 actual projects in https://qa-reports.linaro.org/pfalcon/ :

* When logged in, with existing the login session (for which the original report was submitted): count of front page is 8.
* Logged out: 8
* Private browser window: 8
* Logged in again: 8

mwasilew:
For some reason admin users get different numbers than 'ordinary' users. I'll try to figure out why.

pfalcon:
Thanks. This is definitely a low-priority issue, just wanted to record it.

mwasilew:
I found the issue, it's here (somewhere): https://github.com/Linaro/squad/blob/master/squad/core/models.py#L49

[...]

PR #713, opened 2020-05-01
core: fix project counter display on index page

mwasilew:
Fixes: #709

### DIFF
--- a/squad/core/models.py
+++ b/squad/core/models.py
@@ -10,7 +10,7 @@ from django.db import transaction
 from django.db.models import Q, Count, Sum, Case, When, F, Value
 from django.db.models.functions import Concat
 from django.db.models.query import prefetch_related_objects
-from django.contrib.auth.models import User
+from django.contrib.auth.models import User, AnonymousUser
 from django.conf import settings
 from squad.mail import Message
 from django.forms.fields import URLField as FormURLField
@@ -46,9 +46,11 @@ class GroupManager(models.Manager):
         projects = Project.objects.accessible_to(user)
         project_ids = [p.id for p in projects]
         group_ids = set([p.group_id for p in projects])
+        if not isinstance(user, AnonymousUser):
+            group_ids = group_ids | set([g.id for g in Group.objects.filter(members__in=[user])])
         return self.filter(
-            Q(id__in=group_ids) | Q(members__id=user.id)
-        ).order_by('slug').annotate(
+            id__in=group_ids
+        ).distinct().order_by('slug').annotate(
             project_count=Sum(
                 Case(
                     When(projects__id__in=project_ids, then=1),


chaws:
I think we have another minor bug related to this:

```python
1. class ProjectManager(models.Manager):
2.     def accessible_to(self, user):
3.        if user.is_superuser or user.is_staff:
4.             return self.all()
5.         else:
6.             groups = Group.objects.filter(members__id=user.id).only('id')
7.             return self.filter(Q(group__in=groups) | Q(is_public=True))
```

Take line 6 `Group.objects.filter(members__id=user.id)`:
* for user = AnonymousUser you get a query like 

```sql
SELECT "core_group"."id" 
FROM "core_group" LEFT OUTER JOIN "core_groupmember" ON (
        "core_group"."id" = "core_groupmember"."group_id")
WHERE "core_groupmember"."user_id" IS NULL ORDER BY "core_group"."slug" ASC
```

* whereas for user = User (not admin nor staff) the query would have `IS NULL` replaced by something like `ID IN (...)`

Because of `LEFT OUTER JOIN` behavior, it'll being groups that don't have any members related to it, which I think is wrong!

mwasilew:
@chaws I think it's OK as long as the project is public. Am I missing something?

chaws:
@mwasilew I did some more digging:

Let:
```python
# group without members
gwom = Group.objects.filter(members__id__isnull=True).first()
public_project_gwom = gwom.projects.filter(is_public=True).first()
private_project_gwom = gwom.projects.filter(is_public=False).first()

# group with memebers
gwm = Group.objects.filter(members__id__isnull=False).first()
public_project_gwm = gwm.projects.filter(is_public=True).first()
private_project_gwm = gwm.projects.filter(is_public=False).first()
```

I tested the following conditions:

| group-project/user   | Anonymous User                             | Member User                                | Not-a-member User                              |
|----------------------|--------------------------------------------|--------------------------------------------|------------------------------------------------|
| gwom                 | sees group: yes<br>sees project count: all | N/A                                        | sees group: yes*<br>sees project count: public |
| gwm                  | sees group: no                             | sees group: yes<br>sees project count: all | sees group: yes*<br>sees project count: public |
| public_project_gwom  | sees project: yes                          | N/A                                        | sees project: yes                              |
| private_project_gwom | sees project: no                           | N/A                                        | sees project: no                               |
| public_project_gwm   | sees project: yes                          | sees project: yes                          | sees project: yes                              |
| private_project_gwm  | sees project: no                           | sees project: yes                          | sees project: no                               |

The `yes*` means that group is only visible because there's at least one public project within.

It's just weird that an Anonymous user can still see a group's home page with a list of all projects (private ones included!), whereas a user that's logged in cannot. We can discuss that in the future though.

squad/core/models.py
```diff
@@ -46,9 +46,11 @@ def accessible_to(self, user):
         projects = Project.objects.accessible_to(user)
         project_ids = [p.id for p in projects]
         group_ids = set([p.group_id for p in projects])
+        if not isinstance(user, AnonymousUser):
+            group_ids = group_ids | set([g.id for g in Group.objects.filter(members__in=[user])])
         return self.filter(
-            Q(id__in=group_ids) | Q(members__id=user.id)
-        ).order_by('slug').annotate(
+            id__in=group_ids
+        ).distinct().order_by('slug').annotate(
```
chaws, line 53: I tried it locally, `distinct()` is unnecessary as you already removed the join `members__id=user.id`.

chaws:
It looks good, tricky bug to catch. I'm merging this now, we can work on a new PR to fix access issues I mentioned above

PR #721, opened 2020-05-07
ci: lava: add flag allowing to accept results from failed jobs

mwasilew:
If LAVA job fails with infrastructure error but before that produces
valid results, these results can be pulled to SQUAD. It is possible
using CI_LAVA_IGNORE_INFRA_ERRORS flag. It can be set in either backend
or a project.


chaws:
LGTM, i'll just wait for tests to finish then i can merge it

Issue #723, opened 2020-05-08
UI: Can't subscribe to project notifications from project, d'oh


pfalcon:
Steps to reproduce:

1. Be at the project page, e.g. https://qa-reports.linaro.org/pfalcon/zephyr-net/
2. Want to subscribe to project (email) notifications.
3. Seeing no obvious "bell" icon straight on the page itself, proceed to "Project settings" tab.
4. Wander endless to still find nothing.
5. Be perplexed, perhaps assume that SQUAD doesn't support notifications.
6. By a pure chance, go to a user Settings, and find "Subscriptions" tab there.

Don't get me wrong - such a tab, listing all subscriptions and allowing to unsubscribe from a particular one, definitely should be there. But the most obvious place to subscribe to a project is a project itself. The current subscription interface with that gigantic dropdown is rather unwieldy and probably even took more effort to code.

If this is acted upon, I really recommend to consider making the "subscribe" action easily accessible, based on the argument presented in https://lists.lavasoftware.org/pipermail/lava-users/2020-April/002380.html - that there would be a noticeable share of users who are interested to look at the test results in 2 cases: a) during initial project setup; b) when tests fail. To achieve b), they would need to subscribe to notifications, so that should be "at fingertips" action.

Specifically, I might suggest to include "Subscription" tab in Project settings, where all settings are collected, but also add a quick-access straight from the project page(s), which can be in the form of a tab heading or a graphic icon (a bell or similar). In both cases, it should lead to a page with additional subscription options and confirmation button, not lead to an immediate action with some "defaults".

Thanks.

pfalcon:
And yeah, I'm actually keep hitting this issue again and again.

mwasilew:
I totally agree, subscribe button will be a nice addition to the project page.

pfalcon:
Thanks! It's really hard to complain when your bugs are looked into, and such quickly, but for history, there's a concern that the implemented and linked solution (#726) addresses the concern only partially: https://github.com/Linaro/squad/pull/726#issuecomment-627944793

mwasilew:
@pfalcon you can have something 'quickly' or nothing for a pretty long time. IMHO former is better.

pfalcon:
[...]

PR #726, opened 2020-05-11
Subscribe ui fixes

mwasilew:
Fixes: #723

### DIFF
--- a/squad/frontend/templates/squad/_subscribe.jinja2
+++ b/squad/frontend/templates/squad/_subscribe.jinja2
@@ -1,0 +1,17 @@
+    {% if user.is_authenticated %}
+  		{% if project.is_subscribed(user) %}
+		<form class="pull-right" action="{{url("settings-subscription-remove-post")}}" method="POST">
+		{% else %}
+  		<form class="pull-right" action="{{url("settings-subscriptions")}}" method="POST">
+  		{% endif %}
+    		{{ csrf_input }}
+    		<input type="hidden" name="notification-strategy" value="all"/>
+    		<input type="hidden" name="email" value="{{ user.email }}"/>
+    		<input type="hidden" name="subscription" value="{{ project.id }}"/>
+    		{% if project.is_subscribed(user) %}
+    		<button class="btn btn-default" type="submit"><i class='fa fa-bell-slash'>&nbsp;</i></button>
+    		{% else %}
+    		<button class="btn btn-default" type="submit"><i class='fa fa-bell'>&nbsp;</i></button>
+    		{% endif %}
+  		</form>
+  	{% endif %}
--- a/squad/core/models.py
+++ b/squad/core/models.py
@@ -275,6 +275,11 @@ class Project(models.Model, DisplayName):
     def writable_by(self, user):
         return self.group.writable_by(user)
 
+    def is_subscribed(self, user):
+        if self.subscriptions.filter(user=user):
+            return True
+        return False
+
     @property
     def full_name(self):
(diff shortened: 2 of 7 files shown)


squad/frontend/templates/squad/project.jinja2
```diff
@@ -25,3 +25,7 @@
 
 {% endwith %}
 {% endblock %}
+
+{% block javascript %}
+<script type="module" src='{{static("squad/project.js")}}'></script>
```
chaws, line 30: This file doesn't exist. Did you mean to include in this PR?

squad/frontend/templates/squad/_subscribe.jinja2
```diff
@@ -0,0 +1,17 @@
...
+  		<form class="pull-right" action="{{url("settings-subscriptions")}}" method="POST">
+  		{% endif %}
+    		{{ csrf_input }}
+    		<input type="hidden" name="notification-strategy" value="all"/>
+    		<input type="hidden" name="email" value="{{ user.email }}"/>
+    		<input type="hidden" name="subscription" value="{{ project.id }}"/>
+    		{% if project.is_subscribed(user) %}
+    		<button class="btn btn-default" type="submit"><i class='fa fa-bell-slash'>&nbsp;</i></button>
+    		{% else %}
+    		<button class="btn btn-default" type="submit"><i class='fa fa-bell'>&nbsp;</i></button>
```
chaws, line 14: Nice! Looks really cool

chaws:
Looks good to me. I'll wait for comments about `project.js` file then I can merge it

squad/frontend/templates/squad/project.jinja2
mwasilew: ah, that's leftover from AJAX implementation that I abandoned. Will remove.

chaws:
Tests failed, I checked and it seems like Flake8 is now catching use of variables that are "hard to read": https://www.python.org/dev/peps/pep-0008/#names-to-avoid

Never heard of that before...

mwasilew:
yeah, and it's in the test I didn't touch in this patch. Let me try to fix it.

pfalcon:
Guys, I really appreciate a quick turnaround on the reports. But sometimes "quick" means "less attention to detail". And in #723, I specifically mentioned:

> Specifically, I might suggest to include "Subscription" tab in Project settings, where all settings are collected, but also add a quick-access straight from the project page(s), which can be in the form of a tab heading or a graphic icon (a bell or similar). **In both cases, it should lead to a page with additional subscription options and confirmation button, not lead to an immediate action with some "defaults".**

While, code like: https://github.com/Linaro/squad/pull/726/commits/fc0a4dea733ac7305bfb06b5046b5c496c77bb20#diff-b639dad4741e889c2d1ae340bc082f55R11 (i.e. POST + `&lt;input type="hidden" name="notification-strategy"`) makes me believe that a button was added which silently subscribes a user to "all" notifications.

This is a good improvement in one area (access to subscriptions), but arguably only a regression in another area - customizability of subscriptions. Because now a user would have lesser chance to discover that SQUAD allows different types ("strategies") of notifications, and could easily grow a misconception that SQUAD always sends unconditional notifications regardless of test status.

And some users are not interested in "all" notifications. Moreover, it's LAVA's problem that it offers only as "all" notifications, and SQUAD is advertized as a solution to that. It would be only sad that the way to leverage that becomes only harder to discover and leverage.

So again, I appreciate quick response, but IMHO, a more detailed implementation, even if longer timeframe, would be beneficial. Thanks!

Issue #708, opened 2020-04-28
Expose previous build in build status


jscook2345:
I see that there is a status assigned to a build (ex: https://qa-reports.linaro.org/api/builds/35450/status/), but it does not list what the previous build was. I think that should be included.

jscook2345:
@mwasilew @chaws I think this is a higher priority issue.

mwasilew:
Thanks @jscook2345. I already started working on this. It requires a fair bit of changes to the models.

danrue:
@mwasilew since we're implementing a linked list here... just in case it's possible, a "next build" field would also be helpful so that you could traverse builds in either direction (forward or backward).

mwasilew:
@danrue I didn't think about that :) I guess it's possible. The problems start once you try to delete Build objects. But it's an issue for single and double link list. I'll think about that. Thanks!

mwasilew:
@danrue this isn't a linked list (single or double). There is at least one use case that makes link list impossible. We have a test case that checks whether builds always use previous _finished_ build as a baseline. It can happen that finished build is followed by unfinished one. When 3rd build comes it has to have the same baseline as the already unfinished build:

build1 (finished)
build2 (unfinished) - baseline: build1
build3 (unfinished) - baseline: build1

Later build3 can switch baseline to build2 but at least for some time we have to allow for one-to-many relation for baseline. This excludes linked list.

chaws:
@mwasilew in the use case you pointed, which build should be build1's next_build?

mwasilew:
@chaws both build2 and build3. Any better idea?

mwasilew:
@jscook2345 I might be reading your request incorrectly. Are you looking for the reference to the baseline (build that was used to calculate regressions and fixes) or the actual previous build (in whatever state it is)?

jscook2345:
@mwasilew 

The baseline. I would use the previous build for looking over regressions and fixes as you say.

PR #734, opened 2020-05-18
WIP: Project status baseline

mwasilew:
Fixes: #708


squad/core/models.py
```diff
@@ -940,6 +940,8 @@ class ProjectStatus(models.Model, TestSummaryBase):
     last notification.
     """
     build = models.OneToOneField('Build', related_name='status', on_delete=models.CASCADE)
+    baseline = models.OneToOneField('Build', related_name='next_status', on_delete=models.SET_NULL, null=True, blank=True)
```
chaws: IMHO, I think `next_build` field is redundant:

For example, let there be builds b1, b2, b3 and we want to lay them out in a way that exists today (a build baseline is always the most recent finished one):
* b1.baseline = None
* b2.baseline = b1, this will make b1.next_status = b2
* b3.baseline = b2, this will make b2.next_status = b3
* if b2 gets removed, b3.baseline = b1, making b1.next_status = b3

Does that make sense or am I missing something here?

Also, can two builds have a common baseline?

squad/core/migrations/0130_project_status_baseline_next.py
```diff
@@ -0,0 +1,19 @@
...
+
+    dependencies = [
+        ('core', '0129_projectstatus_nullable_notified_on_timeout'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='projectstatus',
+            name='baseline',
+            field=models.ForeignKey(blank=True, null=True, on_delete=django.db.models.deletion.SET_NULL, related_name='next_statuses', to='core.Build'),
```
chaws, line 17: For some reason here we get `related_name='next_statuses'` whereas in the model it's `related_name='next_status'`

squad/core/models.py
mwasilew: If b2 gets deleted, ProjectStatus in b3 might have wrong numbers (as baseline changed). So my idea is to pin the baseline so we don't get this issue. 

Yes, you can have 2 builds with common baseline
 * build1 - finished
 * build2 - not finished (baseline build1)
 * build3 - not finished (baseline build1)

If build3 finishes before build2 (can happen) than without pinning it's baseline we might have numbers wrong, same as with deleting use case.

squad/core/migrations/0130_project_status_baseline_next.py
mwasilew: typo, thanks for spotting.

chaws:
I think you can squash https://github.com/Linaro/squad/pull/734/commits/23fcc6c2ae2e192b3a23271eb1c69ac770e4cace and https://github.com/Linaro/squad/pull/734/commits/f4b503156ebee6c423f1b984a1749522becb5680

`next_build` is added in the first commit, but gets removed in the former. Also the migration in the first commit refers to a model field in the second commit

mwasilew:
Depending on how we want to roll it out I'll split the changes into different commits/pull requests. I guess let's try to review it now and find any 'grave' issues. After everything looks OKish I'll send new PRs.

squad/core/migrations/0130_project_status_baseline_next.py
mwasilew: Fixed. Also migration/model had field type mismatch

squad/core/models.py
```diff
@@ -990,11 +990,15 @@ def create_or_update(cls, build):
         regressions = None
         fixes = None
 
-        previous_build = Build.objects.filter(
-            status__finished=True,
-            datetime__lt=build.datetime,
-            project=build.project,
-        ).order_by('datetime').last()
+        previous_status = ProjectStatus.objects.filter(baseline=build).last()
```
chaws, line 993: I was thinking about this, and it's easier to expose my point with a use case:

b1 is any finished build
b2.status.baseline = b1
b3.status.baseline = b1
b4.status.baseline = b3

When processing `create_or_update` for status of b3, `ProjectStatus.objects.filter(baseline=b3).last()` would retrieve status of b4, which is ahead of b3. Am I missing something here?

I know that b4 wouldn't probably exist by the time b3 is being processed, so that query would return None most of the time.

mwasilew: Maybe you're right. I need to think about that. Maybe I did the comparison wrong way around.

mwasilew: I did this the wrong way around. Should be fixed now.

squad/core/management/commands/compute_project_statuses.py
```diff
@@ -0,0 +1,88 @@
...
+        return datetime.strptime(date, "%Y-%m-%d")
+    except ValueError:
+        msg = "'{0}' is not a valid date.".format(
+            date)
+        raise argparse.ArgumentTypeError(msg)
+
+
+class Command(BaseCommand):
+
+    help = """Compute metric summaries per build per environment"""
```
amrohassaan, line 26: same help msg as computer_build_summaries.py ?

mwasilew: copy/paste error. Thanks for spotting.

squad/core/management/commands/compute_build_summaries.py
```diff
@@ -55,6 +50,11 @@ def add_arguments(self, parser):
             help="End date for project status updates (default: today, format: YYYY-MM-DD)."
         )
 
+    def __progress__(self, show):
+        if show:
+            self.stdout.write(".", ending="")
+            self.stdout._out.flush()
```
chaws, line 56: just fyi, you can do `print('.', end='', flush=True)` for that...

squad/core/management/commands/compute_project_statuses.py
```diff
@@ -0,0 +1,88 @@
...
+            logger.info('Filtering builds from project "%s"' % (project_name))
+            builds = builds.filter(project=project)
+
+        logger.info('Computing metrics summary for %d builds' % (builds.count()))
+        if show_progress:
+            logger.info('Showing progress, one dot means one processed build')
+
+        for build in builds.all():
+            self.__progress__(show_progress)
+            ProjectStatus.create_or_update(build)
```
chaws, line 84: I foresee 1% of buggy results with this approach. Example:

b1 (finished=True)
b2 (finished=False, previous_build=b1)
b3 (finished=False, previous_build=b1)

Let's say `b3` received all TestJobs and it got finished before `b2`. ProjectStatus will be stored with regressions/fixes calculated against `b1`. After sometime, `b2` will also have finished=True. So when running this command, `b3.baseline=None`, therefore `previous_build=b2` and ProjectStatus will be recalculated against `b2` instead of `b1`.

I'm not sure of the % of builds that fall in this scenario, but it's nice to be aware of. I guess there's no way of preventing that...

chaws:
It's looking good! I think it's a start for selecting different builds as baseline!

One thing I wanted to ask: could you please add tests using `status.baseline` and `build.next_statuses` ?

squad/core/management/commands/compute_project_statuses.py
mwasilew: IMHO this is OK. Since we'll have direct link to the baseline it will be clear how the regressions were calculated so I don't consider this a bug.

squad/core/management/commands/compute_build_summaries.py
mwasilew: It was like this before and it didn't work. I guess it's because django uses additional caching.

squad/core/management/commands/compute_project_statuses.py
chaws: Ok

mwasilew:
@chaws I added a test for preserving baseline.

chaws:
Thanks for adding those tests!

mwasilew:
Oups, didn't we mean to split it into 2 PRs?

PR #736, opened 2020-05-22
ci: improve TestJob.cancel()

mwasilew:
When test job is not yet submitted cancel() marks it as submitted and
fetched and sets failure message.

### DIFF
--- a/squad/ci/models.py
+++ b/squad/ci/models.py
@@ -257,6 +257,12 @@ class TestJob(models.Model):
     def cancel(self):
         if self.job_id is not None:
             return self.backend.get_implementation().cancel(self)
+        else:
+            self.fetched = True
+            self.submitted = True
+            self.job_status = "Canceled"
+            self.failure = "Cancelled before submission"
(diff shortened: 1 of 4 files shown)


squad/ci/models.py
```diff
@@ -257,6 +257,12 @@ def force_resubmit(self):
     def cancel(self):
         if self.job_id is not None:
             return self.backend.get_implementation().cancel(self)
+        else:
```
chaws, line 260: What do you think about making this block of code available even if job_id exists? This way job_status would turn to "Cancelled" right away.

mwasilew: I don't think it's a good idea. You can imagine a situation (assuming LAVA backend) when you add a job with watch_job API and then try to cancel it but you don't have enough privileges. You can view the job, but not alter it. In such case job_status would be incorrect. I think it only applies to jobs that haven't been yet submitted, thus no job_id.

chaws:
LGTM

Issue #725, opened 2020-05-08
UI: Build page: Switching among suites*envs/suites/envs views resets "failures only" checkbox


pfalcon:
Steps to reproduce:

1. Go to a page like https://qa-reports.linaro.org/pfalcon/zephyr-net/build/c6edfa09/
2. Uncheck "failures only" checkbox.
3. In the nearby dropdown, choose "by suites".
4. Watch that "failures only" checkbox was automatically switched to be set again, so you don't see any useful information that you expected, and need to do extra handwork again. This (re)setting happens on each view switch.

Expected behavior: on initial page open, "failures only" should be set, and afterwards, while switching views, the checkbox should preserve last value set by user.

mwasilew:
Thanks @pfalcon, nice catch!

chaws:
Closed in #743

PR #743, opened 2020-05-29
frontend: build: fix missing "failures only" filter

chaws:
When users change the layout type in build page, "failures only" checkbox would always reset to checked. This patch preserves its value as users toggle between layout types


chaws:
Fixes https://github.com/Linaro/squad/issues/725

Issue #738, opened 2020-05-26
Server Error on builds comparion when picking n/a tests


amrohassaan:
from builds view picking two builds to compare and then checking the n/a checkbooks renders a server error.

mwasilew:
```RAW_URI = '/_/comparebuilds/?target=v5.6.14-122-g0215fac231ee&project=lkft%2Flinux-stable-rc-5.6-oe&transitions=None&baseline=v5.6.14-121-g8f40203f4915&page=None&transitions=pass%3Afail&transitions=pass%3An%2Fa&transitions=fail%3Apass'```

Problem is with ```transitions=None```. It's either backend that needs to be adjusted, or frontend should pass ```None:None``` or thereabout.

chaws:
It was in the backend: transitions are passed to query string every time, and when page is loaded I had to clean the url from "transitions", bug was that transitions was assigned to None, instead of being removed from the querystring. Fixed now :)

chaws:
Fixed in #744

PR #744, opened 2020-05-29
frontend: comparison: fix transitions

chaws:
There was a bug tha caused 500 errors when new transitions were selected. This was due to how we update querystring parameteres. This patch fixes that.

### DIFF
--- a/squad/frontend/comparison.py
+++ b/squad/frontend/comparison.py
@@ -39,6 +39,8 @@ def __get_transitions(request):
 
     if len(marked_transitions) > 0:
         for t in marked_transitions:
+            if t is None or t == 'None':
+                continue
             _from, _to = t.split(':')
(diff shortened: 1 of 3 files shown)


chaws:
Fixes: https://github.com/Linaro/squad/issues/738

squad/frontend/comparison.py
```diff
@@ -39,6 +39,8 @@ def __get_transitions(request):
 
     if len(marked_transitions) > 0:
         for t in marked_transitions:
+            if t is None:
```
mwasilew, line 42: It's serialized, so should be sth like: 
```
if t is None or t == "None":
```
Right now it throws exception.

chaws: nice spot, just pushed a fix

amrohassaan:
LGTM.

Issue #727, opened 2020-05-12
timeout when filtering tests by suite


chaws:
This https://qa-reports.linaro.org/api/tests/?suite=7672 times out, when directly in the database it returns right away.

I started debugging it and got the query:
```sql
SELECT "core_test"."id",
       "core_test"."test_run_id",
       "core_test"."suite_id",
       "core_test"."metadata_id",
       "core_test"."name",
       "core_test"."result",
       "core_test"."log",
       "core_test"."has_known_issues"
  FROM "core_test"
 INNER JOIN "core_testrun"
    ON ("core_test"."test_run_id" = "core_testrun"."id")
 INNER JOIN "core_build"
    ON ("core_testrun"."build_id" = "core_build"."id")
 WHERE ("core_build"."project_id" IN (SELECT V0."id" FROM "core_project" V0 WHERE (V0."group_id" IN (SELECT U0."id" FROM "core_group" U0 LEFT OUTER JOIN "core_groupmember" U1 ON (U0."id" = U1."group_id") WHERE U1."user_id" IS NULL) OR V0."is_public" = true)) AND "core_test"."suite_id" = 5049)
 ORDER BY "core_test"."id" ASC
 LIMIT 51
```

We need to get rid of those joins and check only if the suite is accessible.

chaws:
After this is done, we should implement `Suite.tests()` in squad-client

chaws:
I'm closing this for now `Suite.tests()` in squad-client will work fine because of https://github.com/Linaro/squad/pull/854

PR #745, opened 2020-05-29
api: rest: disable permission check for Test objects

chaws:
Test objects are a particular case in Squad mainly because queries tend to timeout. This patch removes a check if test objects are accessible to users. We need to figure out a better way of doing it that doesn't time out queries

Fix https://github.com/Linaro/squad/issues/727

### DIFF
--- a/squad/api/rest.py
+++ b/squad/api/rest.py
@@ -1044,7 +1044,8 @@ class TestSerializer(serializers.ModelSerializer):
         exclude = ['metadata']
 
 
-class TestViewSet(ModelViewSet):
+# TODO: Figure a way out on how to check permissions of Test objects
+class TestViewSet(viewsets.ModelViewSet):
 
     queryset = Test.objects.prefetch_related('suite', 'known_issues').all()
     project_lookup_key = 'test_run__build__project__in'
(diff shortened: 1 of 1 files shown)


mwasilew:
I understand the need for a fix but I don't think it's a good idea on it's own. If you want to go this path we need to disable 'tests' and 'metrics' endpoints. BTW metrics should use the same logic. So this should be part of a bigger change in the API, not a standalone patch.

chaws:
@mwasilew I agree. The thing with tests is that it's the atomic unit of SQUAD, so any other filters would have to check it's association to the test object and check if user has permission to it. For example, if a user filters tests by suite, he/she should have permission to read suite, then the project, then the group. If filtering by testrun, permissions would check for testrun, build, project and group. We need a smart and generic way of doing it.

chaws:
@mwasilew I did the same for metrics. Do you think that's OK or should we find the best solution now?

mwasilew:
Maybe we should nest the endpoint. This way you only check permission to the Build or TestRun object?

chaws:
I'm closing this for now, as the original idea for this PR was to implement https://github.com/Linaro/squad-client/pull/46, but that has been solved here https://github.com/Linaro/squad/commit/536f31a9a6ee3858516514cfdb27dc394269db2a and already updated in squad-client

Issue #739, opened 2020-05-26
create a native api endpoint to expose test comparison 


amrohassaan:
It would be ideal if we enable users to use squad client to call test comparing class in squad to pass two arbitrary builds to compare for regressions. This will help avoid the confusion of projectstatus sometimes not having the latest regressions calculated accurately in case testrun get fetched later after the status has already been calculated.

chaws:
Closed in #749

PR #749, opened 2020-06-01
api: rest, endpoint to compare two builds

amrohassaan:
expose an endpoint to compare two builds within same project and report
regressions and fixes.

### DIFF
--- a/squad/api/rest.py
+++ b/squad/api/rest.py
@@ -27,6 +27,7 @@ from squad.core.models import (
     Status
 )
 from squad.core.tasks import prepare_report, update_delayed_report
+from squad.core.comparison import TestComparison
 from squad.ci.models import Backend, TestJob
 from squad.compat import drf_basename
 from django.http import HttpResponse
@@ -467,6 +468,10 @@ class ProjectViewSet(viewsets.ModelViewSet):
      * `api/projects/<id>/unsubscribe` POST
 
         Unsubscribe from project notifications
+
+    * `api/projects/<id>/compare_builds` POST/GET
+
+        Compares two builds and report regressions/fixes
     """
     queryset = Project.objects
     serializer_class = ProjectSerializer
@@ -571,6 +576,29 @@ class ProjectViewSet(viewsets.ModelViewSet):
         data = {"email": subscriber_email}
         return Response(data, status=status.HTTP_200_OK)
 
+    @action(detail=True, methods=['get', 'post'], suffix='compare_builds')
+    def compare_builds(self, request, pk=None):
+        builds_to_compare = {f: request.GET.get(f, request.POST.get(f, None)) for f in ['baseline', 'to_compare']}
+        if all(builds_to_compare.values()):
+            try:
+                int(builds_to_compare['baseline'])
+                int(builds_to_compare['to_compare'])
+                baseline = self.get_object().builds.get(pk=builds_to_compare['baseline'])
+                to_compare = self.get_object().builds.get(pk=builds_to_compare['to_compare'])
+                if not baseline.status.finished or not to_compare.status.finished:
+                    raise serializers.ValidationError("Cannot report regressions/fixes on a non-finished builds")
+            except Build.DoesNotExist:
+                raise NotFound()
+            except ValueError:
+                raise serializers.ValidationError("builds IDs must be integer")
+        else:
+            raise serializers.ValidationError("Invalid args provided. 'baseline' and 'to_compare' build ids must NOT be empty")
+
+        if baseline and to_compare:
+            comparison = TestComparison(baseline, to_compare)
+            serializer = BuildsComparisonSerializer(comparison)
+            return Response(serializer.data)
+
 
 class ProjectStatusSerializer(serializers.HyperlinkedModelSerializer):
 
@@ -671,6 +699,16 @@ class DelayedReportViewSet(viewsets.ReadOnlyModelViewSet):
     filterset_class = DelayedReportFilter
     filter_class = filterset_class  # TODO: remove when django-filters 1.x is not supported anymore
     ordering_fields = ('id', 'created_at')
+
+
+class BuildsComparisonSerializer(serializers.BaseSerializer):
+    def to_representation(self, comparison):
+        ret = {}
+        if comparison.regressions is not None:
+            ret['regressions'] = comparison.regressions_grouped_by_suite
+        if comparison.fixes is not None:
+            ret['fixes'] = comparison.fixes_grouped_by_suite
+        return ret
 
 
 class BuildSerializer(serializers.HyperlinkedModelSerializer):
--- a/test/api/test_rest.py
+++ b/test/api/test_rest.py
@@ -24,15 +24,18 @@ class RestApiTest(APITestCase):
         t3 = timezone.make_aware(datetime.datetime(2018, 10, 3, 1, 0, 0))
         self.build3 = self.project.builds.create(version='3', datetime=t3)
         t4 = timezone.make_aware(datetime.datetime(2018, 10, 4, 1, 0, 0))
-        self.build4 = self.project.builds.create(version='4', datetime=t4)
+        self.build4 = self.project.builds.create(version='v4', datetime=t4)
         t5 = timezone.make_aware(datetime.datetime(2018, 10, 5, 1, 0, 0))
         self.build5 = self.project.builds.create(version='5', datetime=t5)
-        self.environment = self.project.environments.create(slug='myenv')
+        t6 = timezone.make_aware(datetime.datetime(2018, 10, 6, 1, 0, 0))
+        self.build6 = self.project.builds.create(version='v6', datetime=t6)
+        self.environment = self.project.environments.create(slug='myenv', expected_test_runs=1)
         self.environment_a = self.project.environments.create(slug='env-a')
         self.testrun = self.build.test_runs.create(environment=self.environment, build=self.build)
         self.testrun2 = self.build2.test_runs.create(environment=self.environment, build=self.build2)
         self.testrun3 = self.build3.test_runs.create(environment=self.environment, build=self.build3)
         self.testrun4 = self.build4.test_runs.create(environment=self.environment, build=self.build4, completed=True)
+        self.testrun6 = self.build6.test_runs.create(environment=self.environment, build=self.build6, completed=True)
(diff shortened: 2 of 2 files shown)


amrohassaan:
@mwasilew, @chaws this is a quick PoC and not full/proper implementation yet. Comments on direction plz.

squad/api/rest.py
```diff
@@ -571,6 +572,25 @@ def unsubscribe(self, request, pk=None):
         data = {"email": subscriber_email}
         return Response(data, status=status.HTTP_200_OK)
 
+    @action(detail=True, methods=['get', 'post'], suffix='compare_builds', permission_classes=[AllowAny])
```
mwasilew, line 575: I think users with permissions to the project should only be allowed.

squad/api/rest.py
```diff
@@ -571,6 +572,25 @@ def unsubscribe(self, request, pk=None):
...
+    def compare_builds(self, request, pk=None):
+        builds_to_compare = {
+            "first": request.data.get("first", None) if request.method == 'POST' else request.query_params.get("first", None),
```
mwasilew, line 578: baseline? This is the name which we use in ProjectStatus so I think it would be consistent to use here as well.

squad/api/rest.py
```diff
@@ -571,6 +572,25 @@ def unsubscribe(self, request, pk=None):
...
+            "second": request.data.get("second", None) if request.method == 'POST' else request.query_params.get("second", None)
+        }
+        if any(builds_to_compare.values()) and all(builds_to_compare.values()):
+            try:
+                builds = self.get_object().builds.filter(version__in=builds_to_compare.values()).order_by('-created_at')
```
mwasilew, line 583: I think this is wrong. You should instantiate the builds separately as the baseline might come after the build to compare. This is a silly usecase but IMHO we should allow it.

Do you expect version strings to be sent to this method? IMHO it would be better to expect IDs.

squad/api/rest.py
amrohassaan: Agree.

squad/api/rest.py
amrohassaan: @mwasilew I agree it should be ids not versions. I understand the corner case you talk about but what if users wants to compare two specific builds regardless of a baseline existence? Is there such a case even? or should we only compare the passed baseline arg only not a newer one exists?

mwasilew: I don't think it matters which builds you compare. Baseline is 'first' build passed to TestComparison. The code you proposed uses oldest as 'first'. My request it so give user opportunity to select the build which he considers baseline and not make the selection 'secretly'.

amrohassaan:
Fixed as per the review req.

squad/api/rest.py
```diff
@@ -571,6 +572,32 @@ def unsubscribe(self, request, pk=None):
...
+            "to_compare": request.data.get("to_compare", None) if request.method == 'POST' else request.query_params.get("to_compare", None)
+        }
+        if any(builds_to_compare.values()) and all(builds_to_compare.values()):
+            try:
+                int(builds_to_compare['baseline'])
+                int(builds_to_compare['to_compare'])
+                baseline = self.get_object().builds.get(pk=builds_to_compare['baseline'])
+                if not baseline.status.finished:
+                    raise serializers.ValidationError("Cannot report regressions/fixes on a non-finished baseline build")
+                to_compare = self.get_object().builds.get(pk=builds_to_compare['to_compare'])
```
mwasilew, line 588: I think to_compare should also be finished.

mwasilew:
LGTM. I'm not sure if comparing unfinished build makes sense, but otherwise it's OK. Tests failed because of breakage in celery. Check #750 on how to fix it.

squad/api/rest.py
```diff
@@ -571,6 +572,32 @@ def unsubscribe(self, request, pk=None):
         data = {"email": subscriber_email}
         return Response(data, status=status.HTTP_200_OK)
 
+    @action(detail=True, methods=['get', 'post'], suffix='compare_builds')
+    def compare_builds(self, request, pk=None):
+        builds_to_compare = {
+            "baseline": request.data.get("baseline", None) if request.method == 'POST' else request.query_params.get("baseline", None),
```
chaws, line 578: I think could read:
```python
builds_to_compare = {f: int(request.GET.get(f, request.POST.get(f, None))) for f in ['baseline', 'to_compare']}
```

squad/api/rest.py
chaws: I agree with Milosz

squad/api/rest.py
```diff
@@ -571,6 +572,32 @@ def unsubscribe(self, request, pk=None):
...
+            "to_compare": request.data.get("to_compare", None) if request.method == 'POST' else request.query_params.get("to_compare", None)
+        }
+        if any(builds_to_compare.values()) and all(builds_to_compare.values()):
```
chaws, line 581: The `any()` doesn't enforce that all values in the dictionary are filled, but `all` does, I think it's safe to remove the `any` then

chaws:
It looks good to me too. I'd just work a bit more on the serialization of regressions and fixes given that TestComparison uses a 2-dimensional dictionary where first key is test name and second-level key is a tuple with build and env objects

mwasilew:
Some tests would also be required.

squad/api/rest.py
amrohassaan: ah..was wondering too. thanks

squad/api/rest.py
amrohassaan: @chaws the problem is all() yields true if both values are None as well and when combined with any(), which will returns false in this case (both empty) then this corner is handled properly. Think we should just abort in case both values are empty?

squad/api/rest.py
amrohassaan: @chaws ..this is neat :+1:

squad/api/rest.py
amrohassaan: I was wrong. I thought if endpoint was called without any args at all, then `builds_to_compare` will be an empty dict without even keys. Totally got myself confused :x:

amrohassaan:
@chaws for serializer improvment, i removed the json.dumps for clarity and now it shows a list of test name which is more readable in UI. Also used the existing `regressions_grouped_by_suite` and `fixes_grouped_by_suite` for a friendlier presentation. Not sure if we want it this way tho, thoughts?

squad/api/rest.py
```diff
@@ -571,6 +572,29 @@ def unsubscribe(self, request, pk=None):
...
+                int(builds_to_compare['baseline'])
+                int(builds_to_compare['to_compare'])
+                baseline = self.get_object().builds.get(pk=builds_to_compare['baseline'])
+                to_compare = self.get_object().builds.get(pk=builds_to_compare['to_compare'])
+                if not baseline.status.finished and not to_compare.status.finished:
+                    raise serializers.ValidationError("Cannot report regressions/fixes on a non-finished builds")
+            except Build.DoesNotExist:
+                raise NotFound()
+            except ValueError:
+                raise serializers.ValidationError("builds ids must be integer")
```
mwasilew, line 589: Build IDs must be integers.

squad/api/rest.py
```diff
@@ -571,6 +572,29 @@ def unsubscribe(self, request, pk=None):
...
+        else:
+            raise serializers.ValidationError("Invalid args provided. 'baseline' and 'to_compare' build ids must NOT be empty")
```
mwasilew, line 591: s/ids/IDs/

test/api/test_rest.py
```diff
@@ -632,3 +635,16 @@ def test_known_issues_filter_by_environment(self):
...
+    def test_project_compare_builds_with_finished_status_and_regressions(self):
+        foo_suite, _ = self.project.suites.get_or_create(slug='foo')
+        self.testrun4.tests.get_or_create(suite=foo_suite, name='dummy', result=True)
+        self.testrun6.tests.get_or_create(suite=foo_suite, name='dummy', result=False)
+        UpdateProjectStatus()(self.testrun4)
+        UpdateProjectStatus()(self.testrun6)
+        data = self.hit('/api/projects/%d/compare_builds/?baseline=%d&to_compare=%d' % (self.project.id, self.build4.id, self.build6.id))
+        self.assertEqual(1, len(data['regressions']['myenv']['foo']))
+
+    def test_project_compare_builds_with_non_finished_status(self):
```
mwasilew, line 648: Please also add a test that tries to use versions instead of IDs for selecting a build.

mwasilew:
> @chaws for serializer improvment, i removed the json.dumps for clarity and now it shows a list of test name which is more readable in UI. Also used the existing `regressions_grouped_by_suite` and `fixes_grouped_by_suite` for a friendlier presentation. Not sure if we want it this way tho, thoughts?
> @mwasilew tests added.

IMHO _by_suite methods fit here nicely.

test/api/test_rest.py
amrohassaan: @mwasilew sure..make sense

amrohassaan:
Fixes #739

squad/api/rest.py
```diff
@@ -571,6 +572,29 @@ def unsubscribe(self, request, pk=None):
...
+    @action(detail=True, methods=['get', 'post'], suffix='compare_builds')
+    def compare_builds(self, request, pk=None):
+        builds_to_compare = {f: request.GET.get(f, request.POST.get(f, None)) for f in ['baseline', 'to_compare']}
+        if all(builds_to_compare.values()):
+            try:
+                int(builds_to_compare['baseline'])
+                int(builds_to_compare['to_compare'])
+                baseline = self.get_object().builds.get(pk=builds_to_compare['baseline'])
+                to_compare = self.get_object().builds.get(pk=builds_to_compare['to_compare'])
+                if not baseline.status.finished and not to_compare.status.finished:
```
chaws, line 584: If `baseline.status.finished == True`, the "if" will evaluate to False and you're going to be discarding the check for `to_compare.status.finished`. The operator should be "or" instead of "and".

amrohassaan: Good catch @chaws. I will fix this.

test/api/test_rest.py
chaws: @mwasilew did you mean that the api should support both id and version for picking builds for comparison?

mwasilew: no, just ID. The test should exercise the code that checks proper input.

chaws:
LGTM

PR #750, opened 2020-06-03
ci: fix get_listener_url() in case of REST

mwasilew:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,6 +15,7 @@ RUN apt-get update -q=2 && \
         python3-djangorestframework \
         python3-djangorestframework-filters \
         python3-djangorestframework-extensions \
+        python3-future \
         python3-gunicorn \
         python3-jinja2 \
         python3-markdown \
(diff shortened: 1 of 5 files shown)


mwasilew:
I fixed the tests and covered the docker-compose use case. IMHO it can be merged now.

Dockerfile
```diff
@@ -15,6 +15,7 @@ RUN apt-get update -q=2 && \
         python3-djangorestframework \
         python3-djangorestframework-filters \
         python3-djangorestframework-extensions \
+		python3-future \
```
chaws, line 18: I think this indentation is tabbed

requirements.txt
```diff
@@ -8,6 +8,7 @@ django_filter>=2.0
 djangorestframework>=3.9.2
 djangorestframework-filters>=1.0.0.dev0
 drf-extensions
+future
```
chaws, line 11: I knew this fix would be fixed in the future :P

chaws:
Did you rebase against master? I see the commit about removing CODEOWNERS file in this PR as well, but they have different commit hashes if compared to the one in master.

chaws:
LGTM

Issue #680, opened 2020-04-07
Official docker image


mr-sour:
Lava has official images that are versioned and released each month. It would be great if squad also had a matching official image to along with that. I currently have to build and manage this image myself which is kind of a hassle when I goto update as I have to build the whole thing again and keep track of my own image.

https://hub.docker.com/u/linaro

https://git.lavasoftware.org/lava/lava/container_registry

chaws:
Hi @mr-sour , I'm currently working on migrating our deployment to containers, so I'll probably have an official docker image in the following weeks here: https://hub.docker.com/r/qareports/squad/

We've tried taking https://hub.docker.com/u/squad/ but that's already taken :(

chaws:
Just a follow up: the official docker image for SQUAD will be hosted at https://hub.docker.com/repository/docker/squadproject/squad

I'll keep the image version in sync with squad version in pypi, also I'll sync the dockerhub with this repo.

chaws:
@mr-sour official SQUAD images will be hosted at https://hub.docker.com/r/squadproject/squad. Feedbacks are welcome!

I'm closing this now, thanks for opening this ticket

PR #751, opened 2020-06-04
Dockerfile: improve Dockerfile

chaws:
This Dockerfile installs squad from a locally built package which prevents dev stuff from being packed into the image, and also packs squadplugins within.

The patch also updates docker-compose.yml so that all running containers use a shared database.


chaws:
Fix https://github.com/Linaro/squad/issues/680

PR #759, opened 2020-06-11
frontend: add prev & next in build page

### LINKED ISSUE #746: UI: add 'next' and 'previous' build buttons
UI should contain buttons/links to navigate builds without using build list. This feature is meant for single build view only.

amrohassaan:
Fixes #746 adding a previous and next navigation buttons in build page.

### DIFF
--- a/squad/frontend/templatetags/squad.py
+++ b/squad/frontend/templatetags/squad.py
@@ -10,7 +10,7 @@ from hashlib import md5
 from markdown import markdown as to_markdown
 
 from squad import version
-from squad.core.models import Test
+from squad.core.models import Test, Build
 from squad.core.utils import format_metadata
 from squad.jinja2 import register_global_function, register_filter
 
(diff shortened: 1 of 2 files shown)


amrohassaan:
Things that are still missing: 
1. Not found builds (in case we reach last build and click next still) returns None and 404 page. 
2. the prev and next url append '?' to url..need to know why and fix.
3. Still missing tests

squad/frontend/templatetags/squad.py
```diff
@@ -91,6 +92,20 @@ def build_url(build):
     return reverse("build", args=(build.project.group.slug, build.project.slug, build.version))
 
 
+@register_global_function
+def previous_build_url(build):
+    prev_bld_status = ProjectStatus.objects.filter(build__project=build.project, build__datetime__lt=build.datetime).last()
```
amrohassaan, line 97: I chose to go thru the status here rather than the build just in case more info is needed.

chaws: Any particular reason for `build__project=build.project` instead of `build=build`?

chaws:
It looks good to me, I don't know why "?" keeps appearing in the url though

squad/frontend/templatetags/squad.py
amrohassaan: oh wow, forgot am dealing with a db record already :man_facepalming: lol..thanks @chaw :)

amrohassaan:
> for `first()` and `last()` to work properly in this case, I think you will need to specify an explicit ordering in the query (e.g. by datetime)

Makes total sense @terceiro, thanks :)

chaws:
@amrohassaan any updates?

amrohassaan:
> @amrohassaan any updates?

@chaws yes..will finish this today.

squad/frontend/templatetags/squad.py
```diff
@@ -95,6 +95,35 @@ def build_url(build):
...
+        pass
+    if next_build:
+        return build_url(next_build)
+    else:
+        return build_url(build)
+
+
+@register_global_function
+def back_to_latest_build_url(build):
+    return build_url(Build.objects.filter(project=build.project).last())
```
mwasilew, line 124: this might get a bit slow when there are a lot of builds. But I think it's OK to take this risk.

chaws:
LGTM

PR #761, opened 2020-06-17
frontend: fix crash when paginator page is not found

mwasilew:


chaws:
LGTM

PR #762, opened 2020-06-17
frontend: fix URL reverse for test names with /

mwasilew:
When test name contains / it's now converted on the fly to $ to make
regexes work properly. Down side is that $ can't be used in the test
names any more. This should be mentioned in the docs.


amrohassaan:
LGTM.

PR #763, opened 2020-06-18
frontend: fix filtering box in build view

mwasilew:


chaws:
LGTM

mwasilew:
This patch improves things but it's not complete. We're not filtering in envs/suites table contents. I didn't find a good solution for that yet.

chaws:
I think it already handles that. Because `match` will match the whole tag, if you type the environment name it'll display only tags with that environment filtered.

mwasilew:
My point is that the whole environment table is displayed even if user wants to filter on the test across envs. So I think we'll need to enable that somehow.

PR #764, opened 2020-06-18
frontend: add pagination to testjobs view

mwasilew:


chaws:
LGTM, i think it would be nice to filter testjobs, but that's not the scope of this patch

mwasilew:
I'm thinking about filtering with XHR. But that's longer to do

PR #765, opened 2020-06-23
api: rest: make project/<id>/test_results faster

chaws:
This improves query to test_results and also makes UI a bit faster


mwasilew:
It's about twice as fast as it used to be. Also the change in JS makes the page load even if the call to test_results fails. Nice job!

PR #766, opened 2020-06-23
Fix timeout when deleting builds

chaws:
The fix consists in not loading a list of objects to be deleted when Build gets deleted. Some builds have thousands of TestRuns with thousands of tests attached. Loading a list of objects is just not practical.

### DIFF
--- a/squad/admin.py
+++ b/squad/admin.py
@@ -1,0 +1,28 @@
+from django.contrib.admin import ModelAdmin
+from django.utils.translation import ugettext as _
+
+
+class NoDeleteListingModelAdmin(ModelAdmin):
+    """
+    ModelAdmin without list of objects being deleted
+    """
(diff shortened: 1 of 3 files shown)


mwasilew:
I guess this also applies to Project and Group. It might be applicable to CI Backends (they will list all jobs they deal with).

chaws:
@mwasilew good idea. I applied it for Group, Project and TestRun. I think CI backends are OK because they would only list testjobs, and I think the big problem here is loading a list of test objects.

mwasilew:
Having 1M jobs belonging to a backend won't be possible to display. So I'd rather have it there as well.

chaws:
Ah, forgot that! Just pushed an update that enables BackendAdmin to not load object lists when deleting as well.

PR #767, opened 2020-06-23
frontend: improve group page load times

chaws:
This patch improves load times for group page from 1s to 0.2s and lower number of queries from 51 to 7.


amrohassaan:
LGTM

PR #768, opened 2020-06-24
Improve load times for project and builds page

chaws:
This reduces a few queries in project and builds pages:

before:
project home page: 15 queries take ~90ms, page takes ~500ms
builds page: 14 queries takes ~10ms, page takes ~300ms

after

project home page: 8 queries take ~8ms, page takes ~200ms, about 50% faster
builds page: 12 queries takes ~8ms, page takes ~250ms, about 30% faster


amrohassaan:
Great response time indeed :+1:

PR #776, opened 2020-07-02
frontend: test_history: improve load time

chaws:
This does improve load time but increases the number of queries depending on how many test runs. Although it seems silly, doing so will prevent production from spending too much time joining tests x test_runs x builds

### DIFF
--- a/squad/core/history.py
+++ b/squad/core/history.py
@@ -1,25 +1,32 @@
 from collections import OrderedDict
 from django.core.paginator import Paginator
+from django.db.models import Prefetch
+from django.db.models.query import prefetch_related_objects
 
-
-from squad.core.utils import parse_name
-from squad.core.models import Test
+from squad.core.utils import parse_name, split_list
+from squad.core.models import Test, SuiteMetadata, TestRun, KnownIssue
 
 
 class TestResult(object):
 
     __test__ = False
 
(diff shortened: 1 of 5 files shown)


squad/core/history.py
```diff
@@ -28,39 +35,56 @@ class TestHistory(object):
...
     __test__ = False
 
+    def __get_tests__(self, builds, metadata):
+        test_runs = []
+        for build in builds:
+            for test_run in build.test_runs.all():
+                test_runs.append(test_run)
+
+        tests = []
+        for test_runs_batch in split_list(test_runs, chunk_size=100):
```
mwasilew, line 45: maybe it would be even faster to do sth like this:
```
test_run_ids = []
for build in builds:
    test_run_ids += build.test_runs.all().values_list('id', flat=True)

tests = Test.objects.filter(test_run__id__in=test_run_ids)
```
I didn't test this code

chaws: That's exactly what's happening in `prefetch_related_objects(test_runs_batch, Prefetch('tests',`, it translates to something like:

```sql
SELECT "core_test"."id", ...
  FROM "core_test"
 WHERE ("core_test"."metadata_id" = 149002 AND "core_test"."test_run_id" IN (955445, 955501, 955813, ...))
```

PR #778, opened 2020-07-02
frontend: compare: improve compare load times

chaws:
OK This time I took a bit more time reading the code and removing unnecessary queries. Another change in this patch is that instead of sending parallel ajax requests to the backend, it's done one-by-one.

Sending all in parallel via ajax kinda killed performance. The first ajax request as lightning fast but the rest just hung there. The one-by-one approach looked much better in my local env. Also, given that that page is a bit long depending on the number of projects, loading projects one-by-one fills the page accordingly.

### DIFF
--- a/squad/api/rest.py
+++ b/squad/api/rest.py
@@ -28,6 +28,7 @@ from squad.core.models import (
 )
 from squad.core.tasks import prepare_report, update_delayed_report
 from squad.core.comparison import TestComparison
+from squad.core.utils import parse_name
 from squad.ci.models import Backend, TestJob
 from squad.compat import drf_basename
 from django.http import HttpResponse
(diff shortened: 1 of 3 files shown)


squad/api/rest.py
```diff
@@ -347,56 +348,46 @@ class Meta:
...
+class LatestTestResultsSerializer(serializers.BaseSerializer):
+    def to_representation(self, build):
+        metadata = self.context.get('metadata')
+        project_environments = self.context.get('environments')
+        suite = self.context.get('suite')
 
+        test_runs = {tr.id: tr.environment for tr in build.test_runs.all()}
+        tests = Test.objects.filter(
+            test_run_id__in=test_runs.keys(),
+            metadata=metadata,
```
amrohassaan, line 360: So what if a test doesn't have metadata? as mentioned [here](https://github.com/Linaro/squad-client/issues/56#issuecomment-650328768), metadata is optional for tests.

amrohassaan:
:+1:

PR #785, opened 2020-07-07
Add migrations to create auth group and add users

### LINKED ISSUE #740: Add object-level permissions when adding members to groups
We're now using more of REST api, therefore object-level permissions are checked for api endpoints.

Instead doing that manually, the idea of this issue is to create a group of permissions (e.g. "results submitter") and add that to a user that's being added to a group

amrohassaan:
Fixes #740 add migration that automates creating a group with permissions to enable
result submitting and add all users to this group.


amrohassaan:
will update this pr so any newly added user is added to the 'result_submitters' group automatically.

squad/core/migrations/0131_create_resultsubmitters_auth_group_and_add_users.py
```diff
@@ -0,0 +1,60 @@
...
+def split_on_upper_and_make_lower(name):
+    split = re.findall('[A-Z][^A-Z]*', name)
+    return [w.lower() for w in split]
+
+
+def create_result_submitters_group_and_add_users(apps, schema_editor):
+    Group = apps.get_model('auth', 'Group')
+    Permission = apps.get_model('auth', 'Permission')
+    ContentType = apps.get_model('contenttypes', 'ContentType')
+    result_sumbitters, created = Group.objects.get_or_create(name='result_submitters')
```
mwasilew, line 28: please change the name of the group to "squad"

mwasilew:
I also would like to see docs about that. We're adding too many features without mentioning them in the docs.

squad/core/migrations/0131_create_resultsubmitters_auth_group_and_add_users.py
amrohassaan: @mwasilew ok, but why? isn't result_submitters bit more meaningful? or you mean that since this will apply to authenticated users, then squad is a more suitable name?

amrohassaan:
> I also would like to see docs about that. We're adding too many features without mentioning them in the docs.

totally missed this. Will do.

squad/core/migrations/0131_create_resultsubmitters_auth_group_and_add_users.py
mwasilew: IMHO result_submitters is misleading. Creating Annotation is not 'submitting results' for example. The 'squad' group would include all basic django permissions you need to work with SQUAD.

amrohassaan: yup..makes sense.

PR #787, opened 2020-07-10
Fix download of test run attachments

terceiro:


mwasilew:
I think travis is lost because 2 commits in this PR are not on the source branch. I tested locally and it looks OK.

PR #801, opened 2020-07-29
api: rest: add filterset_class attribute to StatusViewSet

chaws:
This fixes the issue where "/api/testruns/2997715/status/?suite=54959" wouldn't filter correctly.


amrohassaan:
:+1:

PR #802, opened 2020-07-29
plugins: linux_log_parser: create metadata

chaws:
This was the only spot in squad that wasn't creating a suitemetadata for tests.

I'm thinking about creating a migration to fill up all empty suitemetadata, but doing so would take a couple of hours (I've tested in testing db)


mwasilew:
I think we don't need a migration. Mgmt command should be sufficient. It can run in the background without interrupting normal operations.

Issue #781, opened 2020-07-03
Number of tests summary in build view


chaws:
Have a summary of number of pass/fail/total tests in suites/envrionments boxes in build view.

Also make boxes closed by default with a "expand all / collapse all" option at the top.

PR #805, opened 2020-07-31
Add totals test results build view

chaws:
Close https://github.com/Linaro/squad/issues/781

Here's a snapshot of what's looking like:

Screenshot from 2020-07-31 19-35-30

### DIFF
--- a/squad/frontend/static/main.css
+++ b/squad/frontend/static/main.css
@@ -378,11 +378,16 @@ table.test-results-box thead tr:first-child {
 a.toggle-test-results {
     display: inline;
     margin-left: 5px;
+    margin-right: 5px;
     background-color: #777;
-    padding: 2px;
-    padding-left: 5px;
-    padding-right: 5px;
+    padding: 3px;
+    padding-left: 6px;
+    padding-right: 3px;
     color: white;
+}
+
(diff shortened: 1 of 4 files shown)


mwasilew:
LGTM. Would it be possible to also add summaries to suites x envs view?

chaws:
Hum, I thought about it but didn't know where to place it in the layout. Under suite name?

chaws:
@mwasilew I've added to suites x environments as you asked. It looks OK in some cases but kinda odd in cases were there are multiple statuses per environment.

Regular (most cases):
Screenshot from 2020-08-03 23-09-43


Cases when there are multiple statuses per env. Notice the boot suite:

Screenshot from 2020-08-03 23-10-41

mwasilew:
@chaws yeah, we should group the statuses when there is more than one and provide the 'confidence score'. But that's not yet implemented. @danrue what do you think? Does this look OK for you, or it's too early to introduce? We could go with first set of patches (for suites and envs views only) and leave this bit for later (after confidence score is implemented).

chaws:
@mwasilew I removed the changed in suite x environments layout for now and will be landing an new PR with that soon.

Issue #798, opened 2020-07-28
Reorder home page


chaws:
Anders pointed something interesting: the home page is mixed with personal groups (with '~' at the beginning) and non-personal groups. This issue is to display non-personal projects before personal-projects

PR #806, opened 2020-08-03
frontend: home: separate groups by active, inactive and usernamespaces

chaws:
Closes https://github.com/Linaro/squad/issues/798

This will split groups into three categories:
- active groups
- inactive groups (don't know if the term is too strong)
- user groups (groups starting with '~')

and it looks like this:

Screenshot from 2020-08-03 17-12-50

### DIFF
--- a/squad/frontend/templates/squad/index.jinja2
+++ b/squad/frontend/templates/squad/index.jinja2
@@ -1,34 +1,58 @@
 {% extends "squad/base.jinja2" %}
 
 {% block content %}
-<h2 class="page-header well">{{ _('All groups') }}</h2>
 
 {% if user.is_authenticated %}
 <div style='margin-top: 2em; margin-bottom: 2em;'>
     <a class='btn btn-default' href="{{url("new-group")}}">{{ _('New group') }}</a>
 </div>
-
 {% endif %}
-<div class='highlight-row'>
-{% for group in groups %}
-<div class="row row-bordered project">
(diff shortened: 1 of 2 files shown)


mwasilew:
@chaws I'm not sure this is a proper way. I would go with 2 sets: user groups (personal spaces) and groups. Then allow for sorting by 'last updated' or 'alphabetically'. This should address the issue. 'last updated' sort should be default so groups with recent data would show on top. What do you think?

chaws:
@mwasilew I agree. Then sorting would be applied to both sets

chaws:
This is how it's looking like:

Screenshot from 2020-08-04 15-44-38

squad/frontend/templates/squad/index.jinja2
```diff
@@ -1,34 +1,58 @@
...
+        <form action="/" method="GET">
+            <label class="form-check-label" for="order">{{ _('Order by') }}:</label>
+            <input
+              name="order"
+              value="most_active"
+              type="radio"
+              onchange="$(this.form).submit()"
+              {{ 'checked' if request.GET.get('order') != 'by_name' else ''}}
+            />
+            <span>{{ _('most active') }}</span>&nbsp;
```
mwasilew, line 25: "last updated"?

chaws: Ops, you mentioned that in the comment above, didn't catch it :)

Updated it now

PR #807, opened 2020-08-04
frontend: test_results: add overall summary in suite x environments layout

chaws:
This is how it looks like:

Screenshot from 2020-08-04 13-40-15


mwasilew:
@danrue, @nareshkamboju is this something we should rollout or you would rather see the multiple results squashed into single item?

jscook2345:
@chaws Who do you think should review this to get some traction?

chaws:
@jscook2345 this was requested a while ago, I remember Dan saying this might not be a fit when viewing suites x environments. Then I don't remember if someone else was against it. Perhaps getting your, Naresh's and Daniel's approval would do it.

This is how it currently looks like (notice the arrows, you need to click them to see results):

Screenshot from 2021-04-14 06-46-12

jscook2345:
@mrchapp @nareshkamboju @roxell

Can you please review this and leave feedback?

mrchapp:
I'll defer to @nareshkamboju.

chaws:
The old way seems good enough : )

PR #809, opened 2020-08-05
settings: really configure celery max retries when queue server is out of reach

chaws:
If MQ server is out of reach, workers will keep retrying to reconnect to it endlessly.
The setting was there but wasn't really working.
This patch really enforces it


amrohassaan:
:+1:

Issue #202, opened 2018-01-10
record submission details


mwasilew:
Imported from https://projects.linaro.org/browse/QA-1495
Record the token, ip address and similar data to recognize from the data comes from.

mwasilew:
@terceiro do you think using https://github.com/justquick/django-activity-stream is a good idea to implement this feature?

terceiro:
@mwasilew I don't think we need that level of detail. this issue is about something way simpler. but also we never really had the need for this, so maybe it's not even worth implementing it

Issue #473, opened 2019-02-06
Log the user who used the API with his token


axelnxp:
It may be useful to be able to log who used the API, based on his token.
It would give a trace to admin if anything goes wrong.

mwasilew:
I missed this ticket. We used to have a similar one (#202) but decided to drop it in favour of more urgent features. Do you have a use case in which this is important?

axelnxp:
Nothing is urging but it would be nice to have that kind of feature for admin purpose.

PR #811, opened 2020-08-14
api: log user actions as LogEntry

mwasilew:
When user creates/changes/deletes an object using API new LogEntry
object is created. LogEntry objects are available in the admin view.

Fixes: #202, #473

### DIFF
--- a/squad/api/views.py
+++ b/squad/api/views.py
@@ -18,6 +18,10 @@ from squad.core.tasks import ReceiveTestRun
 from squad.core.tasks import exceptions
 
 
+from squad.core.utils import log_addition
+from squad.core.utils import log_change
+
+
 logger = logging.getLogger()
 
 
@@ -47,7 +51,11 @@ def create_build(request, group_slug, project_slug, version):
         fields['patch_id'] = patch_id
 
     create_build = CreateBuild(project)
-    create_build(version=version, **fields)
+    new_build, created = create_build(version=version, **fields)
+    if created:
+        log_addition(request, new_build, "Build created")
+    else:
+        log_change(request, new_build, "Build updated")
     return HttpResponse('', status=201)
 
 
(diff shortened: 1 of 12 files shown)


mwasilew:
I'm still planning to ad history view for each object. The submitted code works but there are no tests for creating log entries yet.

squad/core/utils.py
```diff
@@ -130,3 +131,37 @@ def split_list(_list, chunk_size=1):
         chunks.append(_list[:chunk_size])
         _list = _list[chunk_size:]
     return chunks
+
+
+def _log_entry(request, object, message, flag):
+    from django.contrib.auth.models import AnonymousUser
+    from django.contrib.contenttypes.models import ContentType
+    from squad.http import auth_user_from_request
```
stevanradakovic, line 139: Is there an import loop? Why not put these on the top?

squad/api/views.py
```diff
@@ -92,7 +100,8 @@ def add_test_run(request, group_slug, project_slug, version, environment_slug):
     receive = ReceiveTestRun(project)
 
     try:
-        receive(**test_run_data)
+        testrun = receive(**test_run_data)
+        log_addition(request, testrun, "Test Run created")
```
stevanradakovic, line 104: I figure you can't move these to core/tasks/__init__ because there's no request available there (it'd be nice because of the code repetition though). 
But can't you return a tuple of (testrun, build) if the build was created in ReceiveTestRun? It would be sensible afais.

stevanradakovic (changes requested):
LGTM, couple of comments. Didn't test the UI, I'll trust you :)

squad/core/utils.py
mwasilew: yes, there is. I didn't figure out how to avoid it in some other way.

squad/api/views.py
mwasilew: maybe this is an option. Let me take a look

mwasilew: I fixed that.

PR #812, opened 2020-08-17
ci: lava: ensure project settings take priority

mwasilew:
When resubmitting a test job automatically only global backend settings
were used. This patch uses Project settings as preferrence and only
defaults to global backend settings when Project doesn't define a
variable

### DIFF
--- a/squad/ci/backend/lava.py
+++ b/squad/ci/backend/lava.py
@@ -515,7 +515,11 @@ class Backend(BaseBackend):
 
     def __resubmit_job__(self, test_job, metadata):
         infra_messages_re_list = []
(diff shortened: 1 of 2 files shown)


chaws:
Looking good. One question though: down in `__resubmit_job__` there's another setting check for "CI_LAVA_SEND_ADMIN_EMAIL", do we want to prioritize that setting as well?

mwasilew:
yeah, I know. I guess the proper patch should be moving this implementation to self.settings so project settings are always more important. My point here was to fix the issue asap and I'll do the 'proper' fix later. Does this sound OK?

chaws:
Agreed! I think we should refactor that settings part a bit, specifically in parse_results. I'm merging this

PR #815, opened 2020-08-19
ci: backend: Use project settings in LAVA backend

mwasilew:
This patch ensures all settings used in backend implementation come from
Project (1st choice) or Backend (2nd choice).


squad/ci/backend/lava.py
```diff
@@ -424,23 +424,22 @@ def __get_publisher_event_socket__(self):
...
-            clone_measurements_to_tests = self.__resolve_setting__(ps, 'CI_LAVA_CLONE_MEASUREMENTS', False)
-            ignore_infra_errors = self.__resolve_setting__(ps, 'CI_LAVA_WORK_AROUND_INFRA_ERRORS', False)
-        else:
-            handle_lava_suite = self.settings.get('CI_LAVA_HANDLE_SUITE', False)
-            handle_lava_boot = self.settings.get('CI_LAVA_HANDLE_BOOT', False)
-            clone_measurements_to_tests = self.settings.get('CI_LAVA_CLONE_MEASUREMENTS', False)
-            ignore_infra_errors = self.settings.get('CI_LAVA_WORK_AROUND_INFRA_ERRORS', False)
+        project_settings = None
+        if getattr(test_job, 'target', None) is not None:
+            project_settings = test_job.target.project_settings
```
chaws, line 438: I think this can be done without multiple calls to `yaml.safe_load`:

```python
def __resolve_setting__(...):
...
    tmp_setting = project_settings.get(setting_name)
```

then

```python
def __parse_results__(...):
    project_settings = yaml.safe_load(test_job.target.project_settings) or {}
```

Also I checked TestJob model and `target` is mandatory, perhaps this `getattr(test_job, 'target')` check was due to legacy code.

mwasilew: some tests still fail if I try to do test_job.target and it's not set. I guess tests should be fixed then :)

mwasilew: As for the multiple calls I wanted to have it in one place rather than multiple. It's not only __parse_results__ that need the project settings.

mwasilew:
I changed it a bit. Now for each use there is just one all to yaml.load.

Issue #814, opened 2020-08-19
testjobs tab listing 25 jobs bump to 50 or 100


nareshkamboju:
testjobs tab listing only 25 jobs bump to 50 or 100.
Each build testing around 300 jobs and In many cases
I have to browse the missing jobs results from the "testjobs".

For easy browsing please increase the number of entries from 25 to 50 or 100.
or provide a drop down box to increase number of jobs in a page like LAVA do.

nareshkamboju:
reference link,
https://qa-reports.linaro.org/lkft/linux-stable-rc-5.8-oe/build/v5.8.1/testjobs/

mwasilew:
@nareshkamboju this hits database hard and we're currently investigating ways of improving the performance. I wouldn't expect this feature to happen any time soon.

chaws:
@mwasilew maybe order testjobs by not-fetched first?

mwasilew:
@chaws proper solution IMHO would be to add filtering options to the list, so you could filter by any parameter. At the moment it's low on my priority list. If @amrohassaan can get to it, he can propose something.

mwasilew:
@nareshkamboju how about this:
![testjob_filters](https://user-images.githubusercontent.com/618901/91174883-cecb2b80-e6d7-11ea-9474-f97331eba9db.png)
I'm planning to hide the filter section by default and only leave labels showing current filter. Will this solve your problem?

nareshkamboju:
@mwasilew The proposed solutions looking good to me.
Thank you +1

PR #819, opened 2020-08-25
frontend: add filter to testjobs view

mwasilew:
Fixes: #814


amrohassaan:
LGTM @mwasilew, If a clear fields button can be added would be great to reset form instead of opening and having to clear them manually..but not a biggie.

PR #820, opened 2020-08-25
core/frontend: fix side effects of removing test name column

chaws:
This patch fixes remaining locations affected by removing test name column


stevanradakovic:
Looks good, merging.

PR #821, opened 2020-08-26
Frontend,  add knownissues

### LINKED ISSUE #784: UI for KnownIssues
There currently is no UI for displaying KnownIssue objects. IMHO it would make sense to add the tab to project view showing all KnownIssues associated with project's environments. Only active issues should be displayed by default but inactive (aka historical) ones should also be somehow available. Checkbox/button should be used to show them.

amrohassaan:
Add knownissues tab to project/builds view.
Fixes #784.

### DIFF
--- a/squad/frontend/templates/squad/knownissues.jinja2
+++ b/squad/frontend/templates/squad/knownissues.jinja2
@@ -1,0 +1,131 @@
+{% extends "squad/base.jinja2" %}
+
+{% block content %}
+
+<h2>{{ _('Known Issues') }}</h2>
+
+{% with items=results %}
+{% include "squad/_pagination.jinja2" %}
+{% endwith %}
+
+<div>
+    <div class='row row-bordered'>
+        <div class='col-md-12 col-sm-12 filter'>
+          <a id="searchLink"><button type='button' class='btn btn-primary fa fa-search'></button></a>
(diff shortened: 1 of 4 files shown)


amrohassaan:
Just noticed there is an older dummy commit with 'fix' message will remove and squash when i am done adding the filter.

amrohassaan:
@mwasilew @chaws filtering added. ready for review now.

squad/frontend/templates/squad/knownissues.jinja2
```diff
@@ -0,0 +1,139 @@
...
+        </div>
+        <div id="notes" style="display: none;">
+          <p><strong>{{ _('Notes:') }}</strong></p>
+          <div id="notes_inner">
+            <pre></pre>
+          </div>
+        </div>
+        <div id="active" style="display: none;">
+          <p><strong>{{ _('Active:') }}</strong></p>
+          <div id="active_inner">
```
mwasilew, line 69: It would look better if you displayed FA tick or x instead of 'true' or 'false' string in the modal dialog.

mwasilew (changes requested):
Please also update commit message on change to project-nav.jinja2

squad/frontend/templates/squad/knownissues.jinja2
amrohassaan: @mwasilew this is not fixed.

mwasilew (approved):

PR #823, opened 2020-08-27
Allow equal sign and dicts in django settings

stevanradakovic:
This will allow having i.e. django db options as part of the
database settings. Example {'options': '-c statement_timeout=5000'}.

### DIFF
--- a/squad/settings.py
+++ b/squad/settings.py
@@ -14,7 +14,10 @@ from celery.schedules import crontab
 from django.conf import global_settings
 from email.utils import parseaddr
 from glob import glob
+import contextlib
+import json
 import os
+import re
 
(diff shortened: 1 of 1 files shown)


stevanradakovic:
Ignore everything but the settings.py. I pulled some changes from master as well but this should not break testing.

squad/settings.py
```diff
@@ -170,7 +172,12 @@
 }
 database_config = os.getenv('DATABASE')
 if database_config:
-    db_from_env = dict(x.split('=') for x in database_config.split(':'))
+    db_from_env = dict()
+    for item in database_config.split(';'):
```
chaws, line 176: could you make this split by `:` instead of `;`?
This way current setups wouldn't need to change their environment variables if they didn't need to pass extra settings

PR #824, opened 2020-09-01
Simplify test run query in metrics

stevanradakovic:


stevanradakovic:
@chaws did I just introduce the limit of 10 test runs back in with this pr? damn...

chaws:
What? weird... did you get the most recent version of testing branch? I removed that commit last week

PR #825, opened 2020-09-02
Use WHERE IN instead of INNER JOIN for test run query in metrics.

stevanradakovic:
We need to get all the test runs in a project. It takes far less time
if we do not use join on the testrun table.


chaws:
Looks good to me. I'll merge as soon as tests are finished.

PR #827, opened 2020-09-02
Remove tests file

chaws:
This patch removes tests_file and metrics_file columns from testrun table.

We'll probably need some sort of vacuum to free up old tests_file and metrics_file


mwasilew:
I didn't read the patch but where do we store the submitted files now?

chaws:
@mwasilew that's the catch. We wouldn't anymore :/

For most cases, `tests_file` and `metrics_file` come from a `json.dumps` on results and metrics provided by the lava backend, later they get processed.

This patch would completely remove those fields and would generate them on-the-fly when needed.

There's a pg_toast in db for testrun table that takes 100G, it stores all blob fields (tests_file, metrics_file, log_file, etc). My guess is that half of that is due to tests_file.

Do you think we shouldn't remove these fields?

mwasilew:
These are the files that are originally submitted. If not in db I would like to still have them somewhere. We can keep them on the disk or in S3. This would require some abstraction on how to write/read such files, but that shouldn't be too complicated. IIRC @stevanradakovic already did something similar for LAVA log files.

chaws:
@mwasilew if we want to got down that path we'd better store all files (tests, metrics, logs, attachments) in an abstracted file server like s3. This would probably be the easiest way to shrink the db to less than 100G (5x smaller)

I guess we could just stick with removing name column for now and think about this file server later. This should decrease database growth rate.

mwasilew:
I came across this library: https://django-storages.readthedocs.io. We could move the TextField to FileField. All files can be stored in S3 (or some other backend). By default it would be local filesystem. In order to make migration with zero-downtime we will have to change the field names unfortunately. How does this sound?

chaws:
@mwasilew Does this abstract all file handling functions through a django model? It looks very neat!
It would be cheaper to store files in S3 ($0.023/Gb-month against $0.115/Gb-month in RDS)

We would need a way to determine file names and directories given that there's almost 2 million test runs.

mwasilew:
Yes, I think it abstracts storage for all FileField objects. The migration to copy stuff from TextField to FileField is on us. I'll work on that. I'll pull your patches and add my on top.

chaws:
Ack

chaws:
Closing this for now, as it will be replaced by moving tests file an other file objects into S3 by https://github.com/Linaro/squad/pull/850 and https://github.com/Linaro/squad/pull/849

PR #830, opened 2020-09-03
frontent: build: include test jobs with no status in Test jobs tab

chaws:
Users complained that the "Test jobs" tab in the build view showed zero jobs, but when clicked, it showed a list of jobs.

This happened due to the way that tab was being counted. It didn't count jobs which "job_status=null".

This patch changes it so that it includes all test jobs despite of their status

### DIFF
--- a/squad/ci/templatetags/filter_jobs.py
+++ b/squad/ci/templatetags/filter_jobs.py
@@ -1,8 +1,12 @@
-from django.db.models import Count
+from collections import defaultdict
 from squad.jinja2 import register_filter
 
 
 @register_filter
 def filter_jobs(build):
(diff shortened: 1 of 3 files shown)


chaws:
Just for the record: https://code.djangoproject.com/ticket/15183

Using Django's default `Count` all null values are ignored.

mwasilew:
I guess it would be better to make sure that job_status is never null and force backends to fill it in when creating a job.

chaws:
@mwasilew what to initialized with,  "Created"?

mwasilew:
@chaws yeah, some constant defined in the squad.ci.backend.models

PR #832, opened 2020-09-07
Lava xml requests transport

mwasilew:
It's a potential fix for issues with pods hanging on pending requests to LAVA.

### DIFF
--- a/squad/ci/backend/lava.py
+++ b/squad/ci/backend/lava.py
@@ -2,6 +2,7 @@ import json
 import re
 import requests
 import ssl
+import socket
 import traceback
 import yaml
 import xmlrpc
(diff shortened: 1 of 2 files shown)


mwasilew:
Hmm, it doesn't look OK. Github picked older version of the patch. Please don't merge yet before this issue is resolved.

mwasilew:
I think I pushed to wrong branch. Should be fixed now.

chaws:
LGTM, I think testing timeouts is a bit tricky, but those should be OK

mwasilew:
tests only check what happens if actual timeout happens. I didn't write any tests that trigger the timeout in requests. Let's see if this helps getting rid of the strange suspended pods.

chaws:
I'm keeping an eye one the queue length since then. I'll roll this out to prod today and keep an eye as well. I'm also considering change auto scale metrics from cpu usage to number of tasks in the queue.

PR #834, opened 2020-09-09
Create composite index in testrun table.

stevanradakovic:
Create composite index between id and environment_id in core_testrun.
This way the query required for metrics page will use index only scan
instead of bitmap hash scan and will be much faster.


chaws:
LGTM, merging and updating testing with it

PR #836, opened 2020-09-09
Use SuiteMetadata.name in favor of Test.name

chaws:
I'm bringing this commit over from testing branch. I tested it navigating through the frontend and refetched jobs across lkft/android projects to make sure plugins were exercised as well. Still feedback is welcome as this patch does change code significantly.

This PR only stops using Test.name field, let's merge this prior to actually removing DB column: https://github.com/Linaro/squad/pull/863


squad/api/rest.py
```diff
@@ -199,7 +199,7 @@ class TestFilter(filters.FilterSet):
 
     class Meta:
         model = Test
-        fields = {'name': ['exact', 'in', 'startswith', 'contains', 'icontains'],
+        fields = {'metadata__name': ['exact', 'in', 'startswith', 'contains', 'icontains'],
```
terceiro, line 202: does this break existing queries that are using `name`? is is possible to have `name`  as an alias to `metadata__name`?

squad/core/models.py
```diff
@@ -761,6 +760,12 @@ class Test(models.Model):
     def __str__(self):
         return self.name
 
+    @property
+    def name(self):
+        if self.metadata is not None:
+            return self.metadata.name
+        return 'missing test name'
```
terceiro, line 767: it's probably a matter of taste, but I would prefer 'missing test name` to be the fallback and not the default, i.e.

```python
if self.metadata is None:
    return 'missing test name'
return self.metadata.name
```

squad/core/models.py
```diff
@@ -810,9 +815,6 @@ def history(self):
         self.__history__ = Test.History(since, count, last_different)
         return self.__history__
 
-    class Meta:
-        ordering = ['name']
```
terceiro, line 814: this will cause all the places where tests were ordered by name to not be anymore. maybe this will have annoying effects. or maybe ordering on `metadata__name` is too expensive for us to even consider?

squad/frontend/tests.py
```diff
@@ -85,15 +85,15 @@ def __get_page_filter__(self, page, per_page):
...
+            if key not in stats:
+                stats[key] = {'pass': 0, 'fail': 0, 'xfail': 0, 'skip': 0, 'ids': []}
+            stats[key][test.status] += 1
+            stats[key]['ids'].append(test.id)
 
         def keyfunc(item):
-            name = item[0]
             statuses = item[1]
-            return tuple((-statuses[k] for k in ['fail', 'xfail', 'skip', 'pass'])) + (name,)
+            return tuple((-statuses[k] for k in ['fail', 'xfail', 'skip', 'pass']))
```
terceiro, line 96: same comment as above, with this we lose the ordering by name.

squad/frontend/tests.py
```diff
@@ -117,7 +117,6 @@ def get(cls, build, page, search, per_page=50):
             id__in=table.__get_page_filter__(page, per_page)
         ).prefetch_related(
             Prefetch('test_run', queryset=TestRun.objects.only('environment')),
-            'suite',
```
terceiro, line 120: can you ellaborate why `suite` is being removed? is it not used for anything?

squad/frontend/views.py
```diff
@@ -446,11 +446,10 @@ def test_run_suite_tests(request, group_slug, project_slug, build_version, testr
     )
 
     all_tests = context['status'].tests.prefetch_related(
-        'suite',
         'metadata',
         'known_issues',
         'suite__metadata'
-    ).order_by(Case(When(result=False, then=0), When(result=True, then=2), default=1), 'name')
+    ).order_by(Case(When(result=False, then=0), When(result=True, then=2), default=1), 'metadata__name')
```
terceiro, line 452: here you are sorting on `metadata__name`! :)

squad/api/rest.py
chaws: It did break yeah, I looked it up in `django_filters` and made this work
```python
name = filters.CharFilter(lookup_expr='icontains', field_name='metadata__name')
```
I think `icontains` will cover `['exact', 'in', 'startswith', 'contains', 'icontains']` and we'll still preserve name queries like `/api/tests/?name=test_receives_metadata_file`

squad/core/models.py
chaws: Makes sense, I think yours is better to look at :)

squad/frontend/tests.py
chaws: ack

squad/frontend/tests.py
chaws: If you are already traversing a model, that model will be automatically prefetched. Ex:
```python
.prefetch_related('suite__metadata')
```
because there's a prefetch of `suite__metadata`, `suite` is automatically prefetched, so no need to explicitly prefetch it

squad/frontend/views.py
chaws: I guess this one was the correct one!

chaws:
Thanks for reviewing @terceiro , sorry I took some time to apply your comments, had some days off these past two weeks. I've updated this PR.

chaws:
@mwasilew I tried making `name` nullable in the model, but the migration generated a field removal command. I don't think Django likes a non-nullable field to become nullable after there's data in the table.

mwasilew:
That isn't great. Let me check with SQL directly.

mwasilew:
It works with SQL. Maybe just write the migration by hand. The caveat is, there is no going back :)
```
squad=> ALTER TABLE core_test ALTER COLUMN name DROP NOT NULL;
ALTER TABLE
squad=> \d+ core_test
                                                             Table "public.core_test"
      Column      |          Type          | Collation | Nullable |                Default                | Storage  | Stats target | Description 
------------------+------------------------+-----------+----------+---------------------------------------+----------+--------------+-------------
 id               | integer                |           | not null | nextval('core_test_id_seq'::regclass) | plain    |              | 
 result           | boolean                |           |          |                                       | plain    |              | 
 suite_id         | integer                |           | not null |                                       | plain    |              | 
 test_run_id      | integer                |           | not null |                                       | plain    |              | 
 name             | character varying(256) |           |          |                                       | extended |              | 
 log              | text                   |           |          |                                       | extended |              | 
 metadata_id      | integer                |           |          |                                       | plain    |              | 
 has_known_issues | boolean                |           |          |                                       | plain    |              | 
Indexes:
    "core_test_pkey" PRIMARY KEY, btree (id)
    "core_test_1d0cc99a" btree (suite_id)
    "core_test_ba18909e" btree (test_run_id)
    "core_test_metadata_id_644c977a" btree (metadata_id)
    "core_test_name_3fd93ba6" btree (name)
    "core_test_name_3fd93ba6_like" btree (name varchar_pattern_ops)
Foreign-key constraints:
    "core_test_metadata_id_644c977a_fk_core_suitemetadata_id" FOREIGN KEY (metadata_id) REFERENCES core_suitemetadata(id) DEFERRABLE INITIALLY DEFERRED
    "core_test_suite_id_bc51b240_fk_core_suite_id" FOREIGN KEY (suite_id) REFERENCES core_suite(id) DEFERRABLE INITIALLY DEFERRED
    "core_test_test_run_id_fc4d1e19_fk_core_testrun_id" FOREIGN KEY (test_run_id) REFERENCES core_testrun(id) DEFERRABLE INITIALLY DEFERRED
Referenced by:
    TABLE "core_test_known_issues" CONSTRAINT "core_test_known_issues_test_id_7281bf77_fk_core_test_id" FOREIGN KEY (test_id) REFERENCES core_test(id) DEFERRABLE INITIALLY DEFERRED
squad=>

PR #841, opened 2020-09-15
frontend, rest: add dropdown to download attachments

amrohassaan:


squad/frontend/templates/squad/_test_results_suitebox.jinja2
```diff
@@ -34,17 +34,64 @@
...
-                    <td>
-                        {% for status in statuses %}
-                        <a href="{{testrun_suite_tests_url(project.group, project, build, status)}}">{% include "squad/_test_results_summary.jinja2" %}</a><br />
-                        {% endfor %}
+                    <td class="col-md-4 col-sm-4">{{ environment }}</td>
+                {% for status in statuses %}
+                    <td class="col-md-4 col-sm-4">
+                        <a href="{{testrun_suite_tests_url(project.group, project, build, status)}}">{% include "squad/_test_results_summary.jinja2" %}</a>&nbsp;
                     </td>
+                    {% if 'boot' != suite.slug %}
```
mwasilew, line 46: Why is this condition here? Is 'boot' any different than any other suite?

squad/frontend/views.py
```diff
@@ -509,6 +509,7 @@ def __download__(filename, data, content_type=None):
         if content_type is None:
             content_type = 'application/octet-stream'
     response = HttpResponse(data, content_type=content_type)
+    response['Content-Disposition'] = 'attachment; filename="%s"' % filename
```
mwasilew, line 512: Please don't do that. This was intentionally removed.

mwasilew (changes requested):

mwasilew:
squad_attachments

Sth went wrong with the table.

amrohassaan:
@mwasilew can you take another quick look please.

squad/frontend/views.py
```diff
@@ -571,6 +571,12 @@ def build_attachment(request, group_slug, project_slug, build_version, testrun,
     return attachment(request, group_slug, project_slug, build_version, testrun, None, None, filename)
 
 
+@auth
+def download_attachment(request, group_slug, project_slug, build_version, attachment_id):
```
mwasilew, line 575: this is only slightly different from attachment() function. Maybe try patching it instead of creating a new one?

mwasilew (commented):
One of the requests from Android team was to be able to predict the URL of the attachment. Is there any way we can do that?

squad/frontend/views.py
amrohassaan: @mwasilew agree..changed this a bit.

amrohassaan:
> One of the requests from Android team was to be able to predict the URL of the attachment. Is there any way we can do that?

@mwasilew right now the url for attachment will always going to be: 
http://qa-reports.linaro.org/group/project/build/buildversion/attachment/attachment id
does this work?

mwasilew:
> > One of the requests from Android team was to be able to predict the URL of the attachment. Is there any way we can do that?
> 
> @mwasilew right now the url for attachment will always going to be:
> http://qa-reports.linaro.org/group/project/build/buildversion/attachment/attachment id
> does this work?

Not really as you can't predict the ID beforehand. But let's worry about that later. Maybe we can do some smart redirect based on the file name.

squad/frontend/views.py
```diff
@@ -557,6 +557,7 @@ def test_details_metadata(request, group_slug, project_slug, build_version, test
     return __download__(filename, test_run.metadata_file)
 
 
+@auth
```
mwasilew, line 560: why 2 decorators?

squad/frontend/templatetags/squad.py
```diff
@@ -140,6 +140,11 @@ def build_section_url(build, name):
     return reverse(name, args=(build.project.group.slug, build.project.slug, build.version))
 
 
+@register_global_function
+def download_build_attachments_url(group_slug, project_slug, build_version, testrun, filename, name):
+    return reverse(name, args=(group_slug, project_slug, build_version, testrun, filename))
```
mwasilew, line 145: Is it used with different name than 'build_attachments' ?

mwasilew (changes requested):
I also don't see 'attachments' column which makes the dropdown come out of the blue.

squad/frontend/views.py
amrohassaan: oops that was a typo.

amrohassaan:
@mw

> I also don't see 'attachments' column which makes the dropdown come out of the blue.

Since not all projects have attachments, adding the attachment column introduced lots of hassle in the template just to show 'n/a' for those rows that have attachments and the dropdown for those that have attachments. I opted for just showing the dropdown with tooltip explaining what is it rather than wasting lots space or having to change the context for the view to pass on bool flag saying if any has attachments on or not so we can use this to add the attachments column. IOW, in most cases the attachment column shows 'n/a' which didn't justify the complexity to determine when to show it or not.

squad/frontend/templatetags/squad.py
amrohassaan: I don't get what you mean here @mwasilew

mwasilew: why do you pass 'name' variable when it's always the same?

mwasilew:
> > I also don't see 'attachments' column which makes the dropdown come out of the blue.
> 
> Since not all projects have attachments, adding the attachment column introduced lots of hassle in the template just to show 'n/a' for those rows that don't have attachments and the dropdown for those that have attachments. I opted for just showing the dropdown with tooltip explaining what is it rather than wasting lots space or having to change the context for the view to pass on bool flag saying if any has attachments on or not so we can use this to add the attachments column. IOW, in most cases the attachment column shows 'n/a' which didn't justify the complexity to determine when to show it or not.

Fair enough. However the empty dropdown isn't much better. Users won't know what it's for. Tooltip helps but IMHO it's not enough. We need some permanent indication this is for downloading attachments.

squad/frontend/templatetags/squad.py
amrohassaan: changed this one @mwasilew

amrohassaan:
> > > I also don't see 'attachments' column which makes the dropdown come out of the blue.
> > 
> > 
> > Since not all projects have attachments, adding the attachment column introduced lots of hassle in the template just to show 'n/a' for those rows that don't have attachments and the dropdown for those that have attachments. I opted for just showing the dropdown with tooltip explaining what is it rather than wasting lots space or having to change the context for the view to pass on bool flag saying if any has attachments on or not so we can use this to add the attachments column. IOW, in most cases the attachment column shows 'n/a' which didn't justify the complexity to determine when to show it or not.
> 
> Fair enough. However the empty dropdown isn't much better. Users won't know what it's for. Tooltip helps but IMHO it's not enough. We need some permanent indication this is for downloading attachments.
Added a placeholder option in the select control instead of the tooltip.

mwasilew (approved):

PR #844, opened 2020-09-16
core: make metric name 256 characteres long instead of 100

chaws:
This fixes `DataError('value too long for type character varying(100)` we've been having this morning


chaws:
I'm merging and checking if it works well in testing, then I'm planning to move it to prod right after

PR #848, opened 2020-09-22
core, frontend add group 'privileged' access level

amrohassaan:
Add 4th access level for testjob submitters and privileged users
to have a level less than admin and more than results submitters.


amrohassaan:
I think this needs to be split into two prs: one for model change and migration and another to change client code.

mwasilew:
In this case I think it's OK to have just one PR. The migration will be very quick and I don't see how it would affect the service availability.

mwasilew:
LGTM.

PR #849, opened 2020-09-24
Tests file storage

mwasilew:
Add model fields that allow to store attachments outside of DB


chaws:
LGTM, it just needs migrations dependencies updated/rebased against master.

chaws:
I'm merging this, and will fix migrations on a new commit

mwasilew:
It still doesn't work as we agreed to save files to both db and storage. Do
you plan to add this feature separately?

chaws:
@mwasilew looks like TF is taking up a some of your time and I wanted to help a little, I thought about merging this alone and going with your original idea just to get things moving. I hope you don't mind :)

mwasilew:
Sure thing, but I think it's better to have the files saved in both places
to help with migration. Could you prepare a patch for that?

PR #850, opened 2020-09-24
Tests file storage implementation

mwasilew:
Implementation of storing attachments files outside of database.


squad/core/tasks/__init__.py
```diff
@@ -151,16 +153,17 @@ def __call__(self, version, environment_slug, metadata_file=None, metrics_file=N
...
         testrun = build.test_runs.create(
             environment=environment,
-            tests_file=tests_file,
-            metrics_file=metrics_file,
-            log_file=log_file,
             metadata_file=metadata_file,
             completed=completed,
             **metadata_fields
         )
+        testrun.tests_file_storage.save("testrun/%s/tests_file" % testrun.pk, File(StringIO(tests_file)))
```
chaws, line 160: Do you want to keep 
```python
tests_file=tests_file,
metrics_file=metrics_file,	
log_file=log_file,
```
?

This way we can rollout this patch and have data being saved to DB and S3 at the same time

squad/core/tasks/__init__.py
```diff
@@ -151,16 +153,17 @@ def __call__(self, version, environment_slug, metadata_file=None, metrics_file=N
...
+        testrun.metrics_file_storage.save("testrun/%s/metrics_file" % testrun.pk, File(StringIO(metrics_file)))
+        testrun.log_file_storage.save("testrun/%s/log_file" % testrun.pk, File(StringIO(log_file)))
 
         for f, data in attachments.items():
-            testrun.attachments.create(filename=f, data=data, length=len(data))
+            attachment = testrun.attachments.create(filename=f, length=len(data))
```
chaws, line 165: Same here

Dockerfile
```diff
@@ -53,8 +53,8 @@ COPY . /squad-build/
 ENV SQUAD_STATIC_DIR=/app/static
 
 RUN cd /squad-build && ./scripts/git-build && \
-    pip3 install --no-dependencies ./dist/squad*.whl squad-linaro-plugins sentry-sdk==0.14.3 && \
-    pip3 install --no-dependencies ./dist/squad*.whl zipp importlib-metadata asgiref django-bootstrap3 && \
+    pip3 install --no-dependencies ./dist/squad*.whl squad-linaro-plugins sentry-sdk==0.14.3 zipp importlib-metadata asgiref django-bootstrap3 django-storages && \
```
chaws, line 57: Can you make it `django-storages==1.9`? This will support `python3-django` (Django 1.11)

chaws:
Is https://github.com/Linaro/squad/pull/849 included here or am I missing something?

Do you think you can leave both ways (tests_file and tests_file_storage, etc) for now ? This way we would roll these changes out and leave it running for a week or so. That's about the time it will take to move all DB files to S3, when we run the second time we'll be sure that all files will be available in S3. Then it'll be safe to remove tests/metrics/logs attachments from DB and use only storage.

Also, could you also create `MEDIA_ROOT = 'storage'` in `settings.py`? Running this locally is saving testruns and attachments to squad root folder, it would be nice to have it saved in a folder like `storage`

squad/core/tasks/__init__.py
mwasilew: the idea is that we merge #849 first. Roll it out and migrate the data. Only then merge #850 which adds implementation for using the files from the storage. Lastly I will prepare a patch removing the old fields.

Dockerfile
mwasilew: OK

mwasilew:
> Is #849 included here or am I missing something?
> 
> Do you think you can leave both ways (tests_file and tests_file_storage, etc) for now ? This way we would roll these changes out and leave it running for a week or so. That's about the time it will take to move all DB files to S3, when we run the second time we'll be sure that all files will be available in S3. Then it'll be safe to remove tests/metrics/logs attachments from DB and use only storage.
> 

This is why #849 exists.

> Also, could you also create `MEDIA_ROOT = 'storage'` in `settings.py`? Running this locally is saving testruns and attachments to squad root folder, it would be nice to have it saved in a folder like `storage`

Yes, good point.

chaws:
@mwasilew I wonder if we should make saving testrun files and attachments optional and enable it only in qa-reports setup. This would leave other setups as is. What do you think?

mwasilew:
@chaws do you mean to save nothing by default? Than the 'attachment' feature becomes meaningless. IMHO this is more work to re-write all UI/API than saving to MEDIA_ROOT by default.

chaws:
yeah, you got a really good point.

I'm almost about to send a PR using storage and db along, I just stumbled across a dependency issue: when we install `pip3 install boto3` its dependency tree forces upgrading `urllib3`, which breaks compatibility with `python3-requests` installed via `apt` (which is `requests==1.24`, a bit old). I [downgraded](https://github.com/Linaro/squad/commit/64eaf8b1af2dccfbea11245dc45dd0fda84f66a4) `boto3==1.15`, which uses same urllib3 as `python3-requests` but then I'm facing an auth issue I didn't see when using current boto3. My guess is there's a missing setup for this version of boto3. I'll keep you posted

squad/core/tasks/__init__.py
```diff
@@ -151,16 +153,17 @@ def __call__(self, version, environment_slug, metadata_file=None, metrics_file=N
...
+            attachment.storage.save(f, File(BytesIO(data)))
```
chaws, line 166: This is creating files in the root of storage. I fixed it locally, will send PR soon

chaws:
This PR inspired a new one https://github.com/Linaro/squad/pull/913, which is taking place

PR #853, opened 2020-09-30
frontend: static, export metrics graphs as picture

### LINKED ISSUE #775: export graphs as pictures
Right now it's not possible to save benchmark graphs. This ticket is a request to have ability to export graphs as pictures to be included in some external reporting.

amrohassaan:
Fixes #775, add download button with permalink to each chart in project metrics.

### DIFF
--- a/squad/frontend/static/squad/controllers/charts.js
+++ b/squad/frontend/static/squad/controllers/charts.js
@@ -299,6 +299,7 @@ function ChartPanel($http, DATA) {
             }
 
             var ctx = document.createElement('canvas')
+            ctx.setAttribute('id', this.metric.name)
             target.appendChild(ctx)
 
             var scatterChart = new Chart(ctx, {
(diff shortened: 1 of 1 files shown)


squad/frontend/static/squad/controllers/charts.js
```diff
@@ -719,6 +720,13 @@ function ChartsController($scope, $http, $location, $compile, ChartPanel, DATA)
                 if (! metric.dynamic) {
                     metric.drawn = true
                 }
+                var canvas_id = document.getElementById(metric.name).id
+                var button_id = canvas_id + "-anchor"
+                var canvas_div = document.getElementById(metric.name+'-div').children[0]
+                var download_elem = "<a id='"+ button_id + "' ng-click='downloadChartImage(\""+ canvas_id + "\""+
```
amrohassaan, line 726: @mwasilew I am not sure we can have a permalink pre-created since the href is populated once the 'download' button is clicked first time...hmm..will look into this again later tomorrow.

mwasilew: @amrohassaan then we need an API to do that. The idea is to allow squad client downloading images for the purpose of reporting.

mwasilew:
As agreed, there is no easy way to call JS from client. Let's merge it and find a better long term solution later.

PR #854, opened 2020-10-02
api: rest: add tests endpoint to suite view

chaws:
Really implements https://github.com/Linaro/squad/blob/1.22/squad/api/rest.py#L1085

### DIFF
--- a/squad/api/rest.py
+++ b/squad/api/rest.py
@@ -1071,9 +1071,9 @@ class SuiteViewSet(viewsets.ModelViewSet):
     """
     Additional actions:
 
-     * `api/suites/<id>/testnames` GET
+     * `api/suites/<id>/tests` GET
 
-        Returns list of all test names belonging to this suite
+        Returns list of all test belonging to this suite
     """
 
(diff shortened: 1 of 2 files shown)


squad/api/rest.py
```diff
@@ -1071,20 +1071,24 @@ class SuiteViewSet(viewsets.ModelViewSet):
...
     queryset = Suite.objects.all()
     serializer_class = SuiteSerializer
     filterset_class = SuiteFilter
     filter_class = filterset_class  # TODO: remove when django-filters 1.x is not supported anymore
 
-    @action(detail=True, methods=['get'], suffix='testnames')
-    def testnames(self, request, pk=None):
-        # In [4]: [t.name for t in SuiteMetadata.objects.filter(kind='test', suite='testsuite1')]
-        pass
+    @action(detail=True, methods=['get'], suffix='tests')
```
mwasilew, line 1084: IIRC the idea behind 'testnames' endpoint was to just present the list of names. This would be much faster than retrieving list of Test objects. I'm OK with adding both.

PR #857, opened 2020-10-06
core, models: add unit field to Metric model

### LINKED ISSUE #851: Store Metric unit in database
Currently Metric objects only store a number. It would be good to also include an optional field for storing a unit. In case of LAVA units can be pulled from the results.

amrohassaan:
Fixes #851.


mwasilew:
I think this patch is incomplete. Currently there is nothing in the code that uses this field. I can update lava backend but IMHO this change should contain addition to our results parsing as well.

amrohassaan:
@mwasilew agree. I will go thru the code to see where it needs to be added/considered.

amrohassaan:
This still has lots of work/fixes to be done. Touches many areas of code and tests...WIP.

amrohassaan:
@chaws @mwasilew i 'think' this can be reviewed now.

squad/core/tasks/__init__.py
```diff
@@ -86,18 +86,22 @@ def __validate_metrics(self, metrics_file):
...
-                    value = float(value)
-                except ValueError:
-                    raise exceptions.InvalidMetricsData.value(value)
-            if type(value) not in [int, float, list]:
-                raise exceptions.InvalidMetricsData.value(value)
-            if type(value) is list:
-                for item in value:
-                    if type(item) not in [int, float]:
+        for metric, value_dict in metrics.items():
+            for key, value in value_dict.items():
```
chaws, line 90: I think you don't need this loop given there's just "unit" and "value" keys in "value_dict".

chaws:
LGTM. I see that all tests related to metrics have "unit" now, but it would be nice to have a test for backward compatibility using metric with no unit. This will probably happen with legacy scripts (e.g. squad-client submit) sending metrics with no unit, I mean before we upgrade it.

squad/core/tasks/__init__.py
amrohassaan: @chaws , yup I have changed this and made the validation of metrics and the parsing handle both, the legacy metric with no unit and the new one. Also added the test you for this.

PR #862, opened 2020-10-07
requirements: add support to use AWS resources

chaws:
This patch makes squad able to communicate to AWS, particularly using django-storages, which allows storing files in DB while still being managed as a regular django model.

@mwasilew I chose a slightly older (1.9, while current one is 1.10) version of django-storages which still supports Django 1.11, which is the one we use in our docker images. This solves the problem we discussed here https://github.com/Linaro/qa-reports.linaro.org/issues/63#issuecomment-697325729

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -53,8 +53,19 @@ COPY . /squad-build/
 ENV SQUAD_STATIC_DIR=/app/static
 
 RUN cd /squad-build && ./scripts/git-build && \
-    pip3 install --no-dependencies ./dist/squad*.whl squad-linaro-plugins sentry-sdk==0.14.3 && \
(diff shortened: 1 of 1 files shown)


mwasilew:
Do we need boto3 and django-storages in requirements.txt? I guess it's only required for the containers we deploy not for anything else.

chaws:
@mwasilew you're right! Currently we don't have a way to create a squad image only for our use. I'll leave django-storages and boto3 in Dockerfile only.

chaws:
I overlooked @mwasilew 's PR https://github.com/Linaro/squad/pull/850/commits/67ac59e5f3704596dcde57bec96d43bf36587f2d#diff-3254677a7917c6c01f55212f86c57fbfR56 which already includes `boto3` and `django-storages` to Dockerfile. I'm closing this PR

PR #863, opened 2020-10-08
Remove test name column

chaws:
After https://github.com/Linaro/squad/pull/836 gets merged, this one simply contains the migration to really remove name column from DB


chaws:
I accidentally added the test name removal in another migration, prior to this one. I don't know why Django included that, since it was auto-generated. Anyways, this shouldn't impact migrations in production.

chaws:
All tests have passed now, I'm merging this myself because I plan to roll it out today

PR #868, opened 2020-10-13
settings: add celery result backend

mwasilew:
In order to make CTS plugin jobs run faster celery needs a results
backend. This is required to use celery chords.


chaws:
LGTM

Just for the record: This will apply 8 migrations to the DB

mwasilew:
Yes it will. I'm a bit worried that celery might store a lot of data in the DB. Let's monitor it after next rollout.

chaws:
@mwasilew ack, let's keep an eye on that. We can always delete old data if that's necessary

mwasilew:
Results are only necessary for chords to work. So after the postprocessing task is finished, the data isn't needed any more and can be deleted.

Issue #864, opened 2020-10-09
Broken links in "Failures" section of notification emails


pfalcon:
Just noticed that the HTML part of SQUAD monitoring emails (forwarded/attached) has broken links in "Failures" section. To show it (in plain text):

------------
Failures

disco_l475_iot1:
lava/job (log)
lava/lava-test-monitor (log)
lava/lava-test-monitor-retry (log)
------------

Links to individual testsuites and "(log)" entries lead to 404 (e.g.
https://qa-reports.linaro.org/lite/zephyr-micropython/build/0bdd8820-1344/testrun/2164649).

An HTML part of sample notification email is attached (gzip'ed as github doesn't allow otherwise). Let me know if further info/data is needed.

[squad-email.html.gz](https://github.com/Linaro/squad/files/5354245/squad-email.html.gz)

mwasilew:
@amrohassaan  could you take a look? It should be as simple as updating the default email notification templates in the database (I hope)

PR #874, opened 2020-10-20
core: templates, notification. Fix broken links

amrohassaan:
Fix broken testrun logs links in html notification.
Fixes: #864


mwasilew:
LGTM

PR #880, opened 2020-10-22
core: models: increase size of password in PatchSource

chaws:
After https://github.com/Linaro/squad/pull/877, the encrypted pass had increased in size, this patch increases size of the field.


chaws:
Reminder: re-create patchsource for gerrit TF

PR #882, opened 2020-10-22
frontend: views: avoid loading attachment data

chaws:
This commit https://github.com/Linaro/squad/commit/6ced55dd78fd61dd26827d8df24c1e990da7eb0f has introduced attachment downloads, it also makes the frontend prefetch attachments blindly, which contains `data` field where all data resides. This PR avoids loading this data, only attachment name is needed.


mwasilew:
LGTM, let's wait for the tests to complete.

Issue #876, opened 2020-10-20
SuiteMetadata matching query does not exist.


mwasilew:
It happens in some cases for projects with tradefed plugin:
https://qa-reports.linaro.org/android-lkft/5.4-gki-aosp-master-db845c/build/5.4.72-b34934d40c49/testrun/3332171/suite/cts-lkft$armeabi-v7a.CtsOsTestCases/test/android.os.cts.AutoRevokeTest.testAutoRevoke_userWhitelisting/history/
```
File "/usr/lib/python3/dist-packages/django/core/handlers/exception.py" in inner
  41.             response = get_response(request)
File "/usr/lib/python3/dist-packages/django/core/handlers/base.py" in _get_response
  187.                 response = self.process_exception_by_middleware(e, request)
File "/usr/lib/python3/dist-packages/django/core/handlers/base.py" in _get_response
  185.                 response = wrapped_callback(request, *callback_args, **callback_kwargs)
File "/usr/local/lib/python3.7/dist-packages/squad/http.py" in auth_wrapper
  94.         return func(*args, **kwargs)
File "/usr/local/lib/python3.7/dist-packages/squad/frontend/tests.py" in test_history
  204.         history = TestHistory(project, full_test_name, top=top, page=page)
File "/usr/local/lib/python3.7/dist-packages/squad/core/history.py" in __init__
  78.         metadata = SuiteMetadata.objects.get(kind='test', suite=suite_slug, name=test_name)
File "/usr/lib/python3/dist-packages/django/db/models/manager.py" in manager_method
  85.                 return getattr(self.get_queryset(), name)(*args, **kwargs)
File "/usr/lib/python3/dist-packages/django/db/models/query.py" in get
  380.                 self.model._meta.object_name

Exception Type: DoesNotExist at /android-lkft/5.4-gki-aosp-master-db845c/build/5.4.72-b34934d40c49/testrun/3332171/suite/cts-lkft$armeabi-v7a.CtsOsTestCases/test/android.os.cts.AutoRevokeTest.testAutoRevoke_userWhitelisting/history/
Exception Value: SuiteMetadata matching query does not exist.
```

chaws:
The link now opens fine, but environment columns are still not present.

@amrohassaan any idea why?

also, I let https://github.com/Linaro/squad/pull/889 running... it should take a day or two due to query complexity

amrohassaan:
[...]

PR #887, opened 2020-10-28
frontend, views: fix suite parsing in test details

amrohassaan:
This was missing which caused test details view to fail if suite slug contained '$'. Exposed by #876


chaws:
Tests aren't complete yet (travis' slow today!), since this change is very minimal I'm merging it right away.

NOTE: it would be nice to have a test for situations like this, I opened this issue https://github.com/Linaro/squad/issues/888 so we can try to target it there

PR #889, opened 2020-10-28
commands: add script to clean tradefed bad entries

amrohassaan:
clean bad suite in suitemetadata generated from tradefed


chaws:
Merging this, let's run in staging then if it's safe run in production

mwasilew:
Hmm, why was this merged with failed travis build? I know that the build failed because of infra issues, but still this should not have happened!

PR #891, opened 2020-10-29
Use storage

chaws:
Inspired from https://github.com/Linaro/squad/pull/850

This PR makes use of storage fields without removing legacy ones. The intention is to start saving data to both DB and S3 so that when migrate legacy data to S3 we'll be sure that no data will be lost.

Next step will be replacing db data by S3 data.

### DIFF
--- a/squad/core/management/commands/migrate_attachments.py
+++ b/squad/core/management/commands/migrate_attachments.py
@@ -1,8 +1,58 @@
-from io import BytesIO, StringIO
-from django.core.files import File
+import math
+import multiprocessing
+import threading
+
+from django.core.files.base import ContentFile
 from django.core.management.base import BaseCommand
+from django.db.models import Q
 
 from squad.core.models import Attachment, TestRun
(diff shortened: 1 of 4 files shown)


mwasilew:
+1. Please merger if the tests complete today.

chaws:
Please don't merge it yet!

chaws:
I think the script to move files to storage is not doing it correctly, combining DB offset and filtering might skip processing some testruns, i'll send a fix today

chaws:
I think I got it right this time.

Migration script takes 1000 testruns at a time per thread. Then it iterates over its attachments to also save it to S3.
Running this locally I got up to 400 testruns per second, but it'll depend heavily on `tests_file` and attachments.

Occasionally I ran out of memory, I don't know why Django won't free memory though.

mwasilew:
LGTM. Should I merge it now?

chaws:
@mwasilew yes, please! Once merged, i'll fire start up exporting files in staging

PR #895, opened 2020-11-03
ci: work around LAVA REST error

mwasilew:
When retrieving results from LAVA, REST API gives different results than
XML-RPC. The issue is described in LAVA ticket:
https://git.lavasoftware.org/lava/lava/-/issues/449
This patch addresses the problem with 'unit' vs 'units'.


chaws:
LGTM

PR #898, opened 2020-11-05
core: add PluginScratch object

mwasilew:
The intention is to save big pieces of string used by plugin tasks into
database. This will prevent crashing message queue brokers due to OOM
conditions.


chaws:
LGTM, I also think it's a good idea to write a cron job to remove "old" plugin data, let's say a month old?

mwasilew:
These objects should be removed by plugin after they use the data. This way I will be able to tell when the plugin postprocessing is done and notifications can be sent. So there should be no permanent scratch objects in the database.

chaws:
Makes sense, I thought of it because I was afraid there would be no way to run a function when a task in celery chord fail, but apparently that's easy to handle https://stackoverflow.com/questions/29355613/run-a-chord-callback-even-if-the-main-tasks-fail

mwasilew:
I'll figure something out. IMHO the sequence should be as follows:
 - task clears after itself (deletes scratch object)
 - in case of crash, scratch is removed when handling exception
 - in case of OOM, there is an error callback in chord
I hope this is good enough to prevent residual scratch objects.

PR #899, opened 2020-11-06
core: commands, rewrite command to fix tradefed

amrohassaan:
Rewrite the command to fix the corrupt data for the tradefed issue to
make it faster.

### DIFF
--- a/squad/core/management/commands/fix_squadplugin_data.py
+++ b/squad/core/management/commands/fix_squadplugin_data.py
@@ -1,9 +1,10 @@
 import logging
 
-from django.db.models import Q, F
+from django.db.models import Q
 from django.core.management.base import BaseCommand
 
 from squad.core.models import Test, SuiteMetadata
+from squad.core.utils import split_list
 
 
 logger = logging.getLogger()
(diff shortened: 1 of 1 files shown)


amrohassaan:
@chaws please don't merge yet until i report back on the status of this locally.
I had the loop in this command running on my shell for almost 6 hours now. I will recheck its accuracy in the morning once it concludes.

chaws:
6 hours is ok compared to almost a week running in prod :)

squad/core/management/commands/fix_squadplugin_data.py
```diff
@@ -21,31 +21,33 @@ def handle(self, *args, **options):
...
-        condition = Q(suite__slug__icontains=android_suites[0])
-        condition |= Q(suite__slug__icontains=android_suites[1])
-        logger.info("fetching %s tests" % (batch_size if batch_size else 'all'))
-        tests = Test.objects.filter(condition).exclude(metadata__suite=F('suite__slug')).prefetch_related('suite', 'metadata').defer('log')
+        condition = Q(suite__icontains=android_suites[0])
+        condition |= Q(suite__icontains=android_suites[1])
+        logger.info("fetching %s metadata for cts and vts tests" % (batch_size if batch_size else 'all'))
+        cts_and_vts_metadata = SuiteMetadata.objects.filter(condition, kind='test')
         if batch_size:
             tests = tests[:batch_size]
```
chaws, line 29: I think `batch_size` is no longer required here

squad/core/management/commands/fix_squadplugin_data.py
```diff
@@ -21,31 +21,33 @@ def handle(self, *args, **options):
...
-        deleted_metadata_count = 0
-        for metadata in buggy_metadata:
-            test_count = Test.objects.filter(metadata=metadata).count()
-            if test_count == 0:
-                metadata.delete()
-                deleted_metadata_count += 1
-
-        print('Deleted %d buggy metadata objects' % deleted_metadata_count)
+        for s in cts_and_vts_metadata: 
+             print("Processing SuiteMetedata with suite %s" % s.suite, flush=True)
```
chaws, line 33: I think it's nice to also print `s.name` to print test name as well.

squad/core/management/commands/fix_squadplugin_data.py
```diff
@@ -21,31 +21,33 @@ def handle(self, *args, **options):
...
+             print("Found Test with for above Suitemetadata. Test.suite is %s " % first_test.suite, flush=True) 
+             correct_suite = first_test.suite.slug 
+             if correct_suite != s.suite:  
+                 correct_metadata = SuiteMetadata.objects.filter(kind='test', suite=correct_suite).first() 
+                 if correct_metadata:
+                     print("Found existing SuiteMetadata with correct suite %s. Will use it to update all affected tests" % correct_metadata, flush=True) 
+                     print("correct_metadata: %s" % correct_metadata.id, flush=True) 
+                     print("Updating tests", flush=True)             
+                     all_affected_tests= Test.objects.filter(metadata=s) 
+                     all_affected_tests.update(metadata=correct_metadata)
```
chaws, line 48: `update` returns the number of updated objects, it would be nice to print how many objects got updated as well

amrohassaan:
@chaws, i left this running. Looks good so far. I will report back on the status tomorrow morning then we can merge.

amrohassaan:
@chaws This script/command has fixed all suitemetadata where 'cts' or 'vts' exists in the 'suite' in SuiteMetadata, however not all 'suite' has 'vts' or 'cts', for example the following were not discovered by the script. This will muddy things a little bit if we wanna use this method of starting from the suitemetadata to have the command run faster. I don't think it is safe to have it also look for any entries in suitemetadata where suite like 'arm%'. So what do you think? should we leave the already running command as is? it will take longer but it will be more accurate since it starts from the test itself?
   id   |                  suite                  | kind |              name               | description | instructions_to_reproduce 
--------+-----------------------------------------+------+---------------------------------+-------------+---------------------------
 247353 | arm64-v8a.bpf_module_test/BpfRaceTest   | test | BpfRaceTest.testRaceWithBarrier |             | 
 247357 | armeabi-v7a.bpf_module_test/BpfRaceTest | test | BpfRaceTest.testRaceWithBarrier |             |

mwasilew:
@amrohassaan there will be more like this. I can probably prepare a full list.

amrohassaan:
> @amrohassaan there will be more like this. I can probably prepare a full list.

this will be very helpful then we can hardcode them and be done once and for all.

PR #900, opened 2020-11-06
core: delay notification if there are PluginScratch-es

mwasilew:
When plugin is still running on the build, but the ProjectStatus for
this build is marked as finished, delay sending notification until
timeout or until the plugin finishes (whichever comes first)


mwasilew:
With current plugins this should be noop change. I tested locally and it works.

chaws:
LGTM

Just for the record: plugins using PluginScratch should determine the best moment to call `UpdateProjectStatus` and send notification.

PR #901, opened 2020-11-06
core: migrate_attachments: reduce number of threads

chaws:
Each thread might spawn multiple threads to make S3 uploads, thus causing OOM errors and `[WARNING] Connection pool is full, discarding connection` from urllib3.

Having 2 threads seems to be the sweet spot (tried with 4 and 8, but pod always gets killed)

### DIFF
--- a/squad/core/management/commands/migrate_attachments.py
+++ b/squad/core/management/commands/migrate_attachments.py
@@ -1,5 +1,4 @@
 import math
-import multiprocessing
 import threading
 
 from django.core.files.base import ContentFile
@@ -10,7 +9,7 @@ from squad.core.models import Attachment, TestRun
 from squad.core.utils import split_list
 
 
-STEP = 1000
+STEP = 100
 
 
 class TestRunFileExporterThread(threading.Thread):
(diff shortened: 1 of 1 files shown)


squad/core/management/commands/migrate_attachments.py
```diff
@@ -78,7 +77,7 @@ def handle(self, *args, **options):
             print('Nothing to do!')
             return
 
-        num_threads = multiprocessing.cpu_count() * 2
+        num_threads = 2
```
mwasilew, line 80: maybe this should be a parameter you can define in runtime. Default to 2.

chaws: Good idea! I've parameterized it

squad/core/management/commands/migrate_attachments.py
```diff
@@ -65,20 +64,38 @@ def add_arguments(self, parser):
...
     def handle(self, *args, **options):
         show_progress = options['show_progress']
+        num_threads = options['num_threads']
+        groups = options['groups']
 
         null_storage_files = Q(tests_file_storage__isnull=True) | Q(metrics_file_storage__isnull=True) | Q(log_file_storage__isnull=True)
-        testrun_ids = TestRun.objects.filter(null_storage_files).order_by('id').values_list('id', flat=True)
+        queryset = TestRun.objects.filter(null_storage_files)
+
+        if groups != 'all':
```
mwasilew, line 89: 'all' is a valid group name. Maybe default should be empty list? You could use nargs=+ to specify that you're after a list.

mwasilew (changes requested):

squad/core/management/commands/migrate_attachments.py
chaws: Oh, I didn't think of it, let me fix it

chaws: Done, I've just make `default=''`

PR #902, opened 2020-11-10
api: rest: hide storage fields

chaws:
These aren't useful for the enduser and they'll be abstracted by their non-storage siblings.


mwasilew:
travis is too slow today. Merging as it won't cause any issues and needs to be rolled out.

PR #904, opened 2020-11-13
ci: add started_at and ended_at fields to TestJob

mwasilew:
Currently only LAVA backend implements filling in started_at and
ended_at fields. They can be empty for other backends.


mwasilew:
@stevanradakovic, @mrchapp please check if this is helpful.

stevanradakovic:
Sure is, thanks Milosz.

test/ci/backend/test_lava.py
```diff
@@ -300,7 +300,26 @@
     'status': 'Complete',
     'id': 1234,
     'definition': yaml.dump(JOB_DEFINITION),
-    'multinode_definition': ''
+    'multinode_definition': '',
+    'start_time': '2020-11-13T10:23:50.019407Z',
+    'end_time': '2020-11-13T10:23:41.265697Z'
```
mrchapp, line 305: Ends before starting?

mrchapp (commented):
+1 Looks good to me, but someone else must approve.

test/ci/backend/test_lava.py
mwasilew: who knows, maybe it travels faster than light? :)

chaws: `job name: back to the future`

chaws:
LGTM! I'm amazed to see this number of missing target in testjob tests. Tests are failing, i guess it's a flake8 nit-pick

mwasilew:
When you don't save the object to DB, there is no enforcement on the 'target'. So it used to work for tests. But not the object is saved early, so properties have to be adjusted. I'll take a look at the failure.

PR #908, opened 2020-11-19
ci: models: clear TestJob.falure on successful

chaws:
TestJobs that received TemporaryFetchIssue and are successfuly fetched afterwards still keep the failure in them. This patch makes sure to clear any failure on successful fetches.

### DIFF
--- a/squad/ci/models.py
+++ b/squad/ci/models.py
@@ -93,6 +93,7 @@ class Backend(models.Model):
 
             test_job.fetched = True
             test_job.fetched_at = timezone.now()
+            test_job.failure = None
             test_job.save()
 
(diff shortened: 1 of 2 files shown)


mwasilew:
I have a feeling we fixed it already before. I wonder why this came back. Thanks for the patch!

chaws:
@mwasilew I had the same feeling, but that was related to TestJob submission (https://github.com/Linaro/squad/pull/650). I wonder if keeping a history of TestJob failure is something we want in the future

mwasilew:
@chaws history could be kept in the django LogEntry objects. IMHO it's a low priority feature.

Issue #907, opened 2020-11-19
Delete storage files as TestRun and Attachment objects get deleted


chaws:
I've learned that When a testrun gets deleted, its storage files aren't, this might become a hassle in the long run when having lots of orphan files.

mwasilew:
is this a "feature" of boto3 or django FileField?

chaws:
Apparently it's Django's "feature": https://stackoverflow.com/questions/5372934/how-do-i-get-django-admin-to-delete-files-when-i-remove-an-object-from-the-datab

I'll add this signal in the same PR I'm working on

PR #913, opened 2020-11-24
Prioritize storage fields

chaws:
This PR makes use of storage fields rather than db ones and also deletes files when model objects get deleted

Also, it fixes https://github.com/Linaro/squad/issues/907

NOTE: I got a few `File not found` errors locally due to missing files (I ended up deleting my `storage` folder because I needed the space). In production environments this **should not** happen

NOTE2: there's also a need for squadplugins, which I'll make tomorrow


squad/core/models.py
```diff
@@ -691,14 +741,42 @@ def __str__(self):
         return self.job_id and ('#%s' % self.job_id) or ('(%s)' % self.id)
 
 
+@receiver(pre_delete, sender=TestRun)
```
amrohassaan, line 744: @chaws this is to delete the files once copying concludes I assume? I will read with fresher eyes tomorrow morning.

chaws: That gets executed when a `TestRun` object is being deleted, i.e. somewhere in `testrun.delete()`

mwasilew:
This looks good. Let's wait for the tests to complete. I'll try it locally

PR #916, opened 2020-11-25
api, rest: enable filtering specific fields

### LINKED ISSUE #903: Make possible to pick api object fields
Currently when retrieving objects from the API, squad returns the full attribute list of objects, it would be nice to select which fields to fetch, in case just a few is needed.

The idea is based on this: https://stackoverflow.com/a/23674297

It'll help squad-client fetch only wanted fields https://github.com/Linaro/squad-client/issues/72

amrohassaan:
Fixes #903. Enable filtering specific fields for queried objects.

### DIFF
--- a/test/api/test_rest.py
+++ b/test/api/test_rest.py
@@ -798,6 +798,15 @@ class RestApiTest(APITestCase):
         data = self.hit('/api/backends/')
         self.assertEqual('foobar', data['results'][0]['name'])
 
+    def test_backends_id_field_only(self):
+        data = self.hit('/api/backends/?fields=id')
+        self.assertNotIn('implementation_type', data['results'][0].keys())
+
+    def test_backends_only_four_fields(self):
+        data = self.hit('/api/backends/?fields=name,implementation_type,poll_interval,max_fetch_attempts')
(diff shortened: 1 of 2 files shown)


chaws:
Awesome, could you write some tests for it? just as a poc

amrohassaan:
> Awesome, could you write some tests for it? just as a poc

Ack @chaws

mwasilew:
Looks OK. Please also add a paragraph in the docs on how to use it. It's obvious from the code, but not everyone reads the code :P

amrohassaan:
@chaws tests added. @mwasilew ack

test/api/test_rest.py
```diff
@@ -810,6 +814,11 @@ def test_groups(self):
         data = self.hit('/api/groups/')
         self.assertEqual('mygroup', data['results'][0]['slug'])
 
+    def test_groups_slug_field_only(self):
+        data = self.hit('/api/groups/?fields=slug')
```
mwasilew, line 818: how about a test with more than 1 field?

amrohassaan: @mwasilew wouldn't harm at all :P

amrohassaan: @mwasilew add the test requested.

chaws:
LGTM

PR #920, opened 2020-11-27
Introduce new notification strategy ON_ERROR.

### LINKED ISSUE #753: [feature request] New notification type: "on failures/infra errors"
Based on the LITE usecases, I'd like to request a new notification type, in addition to existing notification types supported by SQUAD ("all", "changes", "regression"). Tentative name would be "failures". The desired behavior (subject to discussion of fine details is):

1. Definition: "test build" is "failed" if: a) at least one of its test runs has "Incomplete" status; b) among "Complete" test runs, if there's at least one test case with "failed" status.
2. For as a long as as a "test build" has "failed" status, a notification should go out.

Implications:
1. There's no need to track status of "failed" vs "not failed" across test builds, then assess changes of that status across builds, then that change to decide whether to send notification.
1.1. In particular, a case like: there was 20 successful test case in previous build, then 10 in the current build - should not lead to notifications (TODO: discuss/confirm this point). This handling is however relies on such handling that successful results should be complete. (This in turn relies on assumption that if LAVA by some reason cannot run a test job to the completion, that it will set its status as Incomplete.)
2. In regard to p.1, it of course still would be useful to get comparison results in the notification emails - for as long as the email will be sent out at all, per the specified criteria.

stevanradakovic:
This one will send notification to the recipient if any of the jobs
in the build have status "Incomplete".

Fixes #753

### DIFF
--- a/squad/core/notification.py
+++ b/squad/core/notification.py
@@ -3,6 +3,7 @@ from squad.mail import Message
 from django.conf import settings
 import django.template
 import logging
+import yaml
 from django.template.loader import render_to_string
 from re import sub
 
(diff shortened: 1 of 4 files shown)


squad/core/notification.py
```diff
@@ -75,7 +75,9 @@ def recipients(self):
                 if not self.previous_build or \
                    len(self.comparison.regressions) == 0:
                     continue
-
+            elif subscription.notification_strategy == Subscription.NOTIFY_ON_ERROR:
+                if len(self.build.test_jobs.filter(job_status="Incomplete")) == 0:
```
mwasilew, line 79: This is LAVA specific. I propose to use a setting from project_settings to decide on 'error' situation.

mwasilew (changes requested):
Looks OK, but I would prefer the error status to be configurable and not LAVA specific.

chaws:
@mwasilew The error status will be taken from project settings. I'm merging this now

PR #921, opened 2020-11-30
ci: lava: clear TestJob.failure in backend implementation

chaws:
Cleaning TestJob.failure outside of Backend.get_implementation() can hide backend-specific errors.


amrohassaan:
LGTM.

PR #922, opened 2020-12-01
core: test: add Build and Environment reference to Test

chaws:
Adding these two references should take up some extra space in the database, but the trade-offs are OK.

These are going to help improve performance on TestComparison, generating test list among other big advantages when dealing with Build data.

I'm getting an average of 1M tests being updated per minute running 12 threads locally, given that prod has around 1B, this should finish running in 16 hours, but reality it's always different, so I'd say a couple of days.


squad/core/models.py
```diff
@@ -839,6 +839,8 @@ class Test(models.Model):
 
     __test__ = False
 
+    build = models.ForeignKey(Build, related_name='tests', on_delete=models.CASCADE, null=True)
```
amrohassaan, line 842: are we planning to change the null to false later?

chaws: yes! It should be mandatory once we got all filled up

Issue #919, opened 2020-11-26
IndexError: list index out of range on submit


chaws:
I got this from sentry:

```python
IndexError: list index out of range
  File "squad/ci/tasks.py", line 37, in submit
    test_job.backend.submit(test_job)
  File "squad/ci/models.py", line 149, in submit
    test_job.job_id = job_id_list[0]
```

This was due a 408 error `POST https://lavalab.nxp.com/api/v0.2/jobs/ [408] reason: Request Timeout`

And because of that, the job_id list might be an empty list https://github.com/Linaro/squad/blob/1.28.2/squad/ci/backend/lava.py#L362 and here for resubmit https://github.com/Linaro/squad/blob/1.28.2/squad/ci/backend/lava.py#L349

I'll work up a patch for this tomorrow

chaws:
@amrohassaan 

This seems to be a "different" type of timeout. I just found that there's already a check for timeout on job submit: https://github.com/Linaro/squad/blob/master/test/ci/backend/test_lava.py#L436

but I don't know why the `408` timeout didn't raise an exception.

mwasilew:
I briefly checked requests code and it seems they don't raise exception on 408 response. This means we need to handle 408 response on our side.

PR #928, opened 2020-12-04
ci: backend: lava: handle HTTP 408 as TemporarySubmissionIssue

chaws:
Fix https://github.com/Linaro/squad/issues/919

For some reason, requests library doesn't raise 408 Timeout as exception, which causes submit to return empty list of ids and
causing IndexError.

This patch explicitly checks if requests returns 408 and handle that as TemporarySubmissionIssue


mwasilew:
LGTM +1

PR #933, opened 2020-12-08
core, commands: 4th attempt to make faster fix

amrohassaan:
A pattern has been identified for all buggy suitemetadata introduced
by the tradefed bug. The suite start with either 'armeabi*' or 'arm64*'.
Using this the command has been rewritten to execute faster. This should
be the best we have in hand right now.

### DIFF
--- a/squad/core/management/commands/fix_squadplugin_data.py
+++ b/squad/core/management/commands/fix_squadplugin_data.py
@@ -1,52 +1,130 @@
 import logging
+import math
+import threading
 
-from django.db.models import Q
 from django.core.management.base import BaseCommand
-
-from squad.core.models import Test, SuiteMetadata
+from django.db.models import OuterRef, Subquery, F, Q
+from django.db import IntegrityError
+from squad.core.models import SuiteMetadata, Test
 from squad.core.utils import split_list
 
 
+correct_metadata = SuiteMetadata.objects.filter(kind='test', suite=OuterRef('suite__slug'), name=OuterRef('metadata__name'))
+
+first_test = Test.objects.order_by().filter(metadata=OuterRef('pk')).annotate(correct_metadata=Subquery(correct_metadata.values('id')[:1]))
+
+correct_slug = Test.objects.order_by().filter(metadata=OuterRef('pk')).annotate(suite_slug=F('suite__slug'))
+
+annotations = {
+    'correct_metadata_id': Subquery(first_test.values('correct_metadata')[:1]),
+    'correct_suite_slug': Subquery(correct_slug.values('suite_slug')[:1])
+}
+
+buggy_ones = Q(suite__startswith='armeabi') | Q(suite__startswith='arm64')
+
+
 logger = logging.getLogger()
+STEP = 1000
+
+
+class SuiteMetadataFixThread(threading.Thread):
+    def __init__(self, thread_id, suitemetadata_ids, show_progress=False):
+        threading.Thread.__init__(self)
+        self.thread_id = thread_id
+        self.suitemetadata_ids = suitemetadata_ids
+        self.show_progress = show_progress
+
+    def run(self):
+        count = len(self.suitemetadata_ids)
+        logger.info('[thread-%s] processing %d suitemetadata' % (self.thread_id, count))
+        count_updates = 0
+        orphan_metadata = []
+        for offset in range(0, count, STEP):
+            ids = self.suitemetadata_ids[offset:offset + STEP]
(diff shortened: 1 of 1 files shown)


squad/core/management/commands/fix_squadplugin_data.py
```diff
@@ -1,52 +1,32 @@
...
-        android_suites = ['cts', 'vts']
-        condition = Q(suite__icontains=android_suites[0])
-        condition |= Q(suite__icontains=android_suites[1])
-        cts_and_vts_metadata = SuiteMetadata.objects.filter(condition, kind='test').order_by('-id')
-        buggy_metadata = []
-        for s in cts_and_vts_metadata:
-            print("Processing SuiteMetedata with suite %s and test %s" % (s.suite, s.name), flush=True)
+        buggy_patterns = [r'^armeabi', r'^arm64']
+        condition = Q(suite__regex=buggy_patterns[0])
+        condition |= Q(suite__regex=buggy_patterns[1])
```
chaws, line 14: I think you can get away with that without using regex: 
```python
condition = Q(suite__startswith='armeabi') | Q(suite__startswith='arm64')
```

squad/core/management/commands/fix_squadplugin_data.py
```diff
@@ -1,52 +1,32 @@
...
+        buggy_metadata = SuiteMetadata.objects.filter(condition, kind='test')
+        total = len(buggy_metadata)
```
chaws, line 16: `buggy_metadata` is a queryset, so calling `len()` on it will run the entire query just to get a count and run it again when processing the records. You can get a count and make a single query in just one go:

```python3
buggy_metadata = [s for s in SuiteMetadata.objects.filter(condition, kind='test').all()]
```

squad/core/management/commands/fix_squadplugin_data.py
```diff
@@ -1,52 +1,32 @@
...
+            try:
+                s.suite = suite_slug
+                s.save()
+            except IntegrityError:
+                existing_metadata = SuiteMetadata.objects.filter(name=s.name, suite=suite_slug, kind='test').first()
+                fixed_tests_count = Test.objects.filter(metadata=s).update(metadata=existing_metadata)
+                s.delete()
+            finally:
+                counter += fixed_tests_count if fixed_tests_count > 0 else 1
+                print('\r Updated %d out of %d' % (counter, total), end='')
```
chaws, line 31: I think it's missing `flush=True` in this print

squad/core/management/commands/fix_squadplugin_data.py
```diff
@@ -1,52 +1,32 @@
...
-        for s in cts_and_vts_metadata:
-            print("Processing SuiteMetedata with suite %s and test %s" % (s.suite, s.name), flush=True)
+        buggy_patterns = [r'^armeabi', r'^arm64']
+        condition = Q(suite__regex=buggy_patterns[0])
+        condition |= Q(suite__regex=buggy_patterns[1])
+        buggy_metadata = SuiteMetadata.objects.filter(condition, kind='test')
+        total = len(buggy_metadata)
+        counter = 0
+        for s in buggy_metadata:
             first_test = Test.objects.filter(metadata=s).first()
```
chaws, line 19: I think you can avoid `first_test = Test.objects.filter(metadata=s).first()` by annotating it to the initial SuiteMetadata query. Something like that:

```python3
from django.db.models import OuterRef, Subquery, F
first_test = Test.objects.filter(metadata=OuterRef('pk')).annotate(suite_slug=F('suite__slug'))
buggy_metadata = SuiteMetadata.objects.annotate(first_test_slug=Subquery(first_test.values('suite_slug')[:1]))
```

squad/core/management/commands/fix_squadplugin_data.py
```diff
@@ -1,52 +1,32 @@
...
-        for chunk in split_list(buggy_metadata, 1000):
-            SuiteMetadata.objects.filter(id__in=chunk).delete()
-            print("Deleted 1000 bad metadata. Remaining: %s" % len(buggy_metadata))
-        print("Cleaning All bad metadata concluded")
+            suite_slug = first_test.suite.slug
+            fixed_tests_count = 0
+            try:
+                s.suite = suite_slug
+                s.save()
+            except IntegrityError:
```
chaws, line 25: Is there a way to know which SuiteMetadata is this that triggers this exception? If there was, then you could use that and avoid the extra query to get `existing_metadata` and use it directly when update tests.

PR #935, opened 2020-12-14
api: rest: order tests by id instead of build_id

chaws:
Turns out that using build_id to order Test using CursorPagination is not a good idea, since there could be a situation where the "next" page is exact the same as the current one, thus giving an endless list of tests for clients like squad-client.


mwasilew:
+1
Does this make the API much slower?

chaws:
Argh, I was fine for the builds I tested. But it got super slow for builds with +1M tests... let me see what can be done

chaws:
@mwasilew I've ran more scenarios. With 1M tests, regular ordering was super slow to get the first page of results but fine after page 2 onwards. Desc ordering seems quick in any scenario.

PR #936, opened 2020-12-14
frontend,tests: replace $ with / in short testname

amrohassaan:
Found by coincidence in some of android tests, they fail to display
history due to test name having '$' not replaced back with '/'


chaws:
We discusses about that, but I gave it another look and we can't simply replace '$' by '/' because some tests might contain '$' character intentionally. For example: https://qa-reports.linaro.org/api/tests/1017497524/. For more examples check https://qa-reports.linaro.org/api/builds/52483/tests/?name=$

I think it's feasible to achieve the same by using urlencode to build the url then use decode to get suite and test names, what do you think?

amrohassaan:
@chaws , yes encode/decode url is the ultimate fix. I think you have an issue for that. Let's then close this PR and do the right fix.

PR #937, opened 2020-12-15
Handle multiple test results.

stevanradakovic:
Multiple test results can occur during job resubmission.
Handle them correctly in various pages accross the system.
API and notifications update will be part of the later patch.

### DIFF
--- a/squad/core/queries.py
+++ b/squad/core/queries.py
@@ -1,4 +1,5 @@
 import datetime
+from collections import Counter
 from itertools import groupby
 from functools import reduce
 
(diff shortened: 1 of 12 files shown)


squad/core/queries.py
```diff
@@ -195,3 +196,30 @@ def get_dynamic_summary(project, environments, metrics, date_start, date_end):
...
+
+def test_confidence(test):
+    status_priority = ['pass', 'fail', 'xfail', 'skip']
+
+    def most_common(lst):
+        data = Counter(lst)
+        max_count = max(data.values())
+        return {value: count for value, count in data.items() if count == max_count}
+
+    duplicates = models.Test.objects.filter(suite=test.suite, metadata=test.metadata, test_run=test.test_run, environment=test.environment, build=test.build)
```
chaws, line 209: By default all queries to Test will have `order_by('metadata__name')`. I think you can avoid that by `order_by()`

squad/frontend/views.py
```diff
@@ -476,7 +476,28 @@ def test_run_suite_test_details(request, group_slug, project_slug, build_version
     test_name = test_name.replace("$", "/")
     suite_slug = suite_slug.replace("$", "/")
     metadata = get_object_or_404(SuiteMetadata, kind='test', suite=suite_slug, name=test_name)
-    test = get_object_or_404(context['test_run'].tests, suite=context['suite'], metadata=metadata)
+    tests = Test.objects.filter(suite=context['suite'], metadata=metadata)
```
chaws, line 479: I think it's missing `context['test_run'].tests` no?

chaws:
Question:

In when looking for regressions, let's say you're comparing a test with confidence score of 70% against the same one with 100%, is this considered to be a regression?

@mwasilew @stevanradakovic thoughts?

mwasilew:
I would say it's not a regression as long as the result is the same. So pass (70%) and pass (100%) are still 'passed'. The problem starts with 50% (1 pass, 1 fail) confidence. Maybe anything below or equal 50% should be considered 'fail'?

squad/frontend/views.py
stevanradakovic: Not 100% sure. Does it go into the same test_run if it comes from a resubmitted test job?

stevanradakovic:
> I would say it's not a regression as long as the result is the same. So pass (70%) and pass (100%) are still 'passed'. The problem starts with 50% (1 pass, 1 fail) confidence. Maybe anything below or equal 50% should be considered 'fail'?

Makes sense to me. Should I update the regression logic also?

squad/frontend/views.py
mwasilew: no, each test job creates a new test run

stevanradakovic: then it's ok, this should not include it in the filtering

chaws: Hum, then you'd need to filter by `testrun.build` [and maybe by `testrun.environment`?], otherwise you'll be getting tests from the same suite for all builds/envs.

stevanradakovic:
Ready now.

chaws:
LGTM

PR #938, opened 2020-12-15
Enhance test field selection

chaws:
Since @amrohassaan enabled field selection, the tests endpoint gained  a great speed improvement.

Getting 1000 tests used to take ~1s, then if only name/status selected, it takes ~500ms.

The solution was almost good enough, it still made queries to DB to fetch all the fields and just after it get removed before rendered to json.

This PR squeezes out a few extra ms by checking which fields are being requested to reduce DB prefetched and data loading.

Getting 1000 tests with only name/status takes now ~100ms.

I don't think doing the same for all endpoints is necessary, given that tests might be the endpoint with most access.


amrohassaan:
LGTM

PR #942, opened 2020-12-16
Fix a typo in the api docs

jscook2345:


chaws:
Won't wait for checks since this is a typo fix

chaws:
Thanks @jscook2345 !

PR #946, opened 2020-12-18
Use lighter version of TestComparison

chaws:
Use `regressions_and_fixes_only=True` for comparing bulds from the same projects using the API.


jscook2345:
LGTM!

chaws:
Thanks for reviewing @jscook2345 ! Merging

PR #948, opened 2020-12-18
api: rest: enhance builds nested endpoint in projects endpoint

chaws:
This will avoid prefetching testruns when loading project's builds.


amrohassaan:
LGMT

PR #951, opened 2021-01-06
Build callbacks

chaws:
This PR adds a generic Callback class that is currently associated with Build, but this can be extended to any class thanks to Django's generic model relation.

Things worth noting:
1. Callbacks can be created via `POST /api/createbuild/` during build creation or via ` POST /api/builds/<id>/callbacks` on an existing builds using the same parameters
  1.1. There can be only one callback per build per url per event, duplicates aren't accepted
  1.2. The only mandatory field is the URL
  1.3. The callback auth headers can be given at the time of its creation or
  1.4. The callback headers can be defined by default in project settings using `CALLBACK_HEADERS` variable. They should be formatted as key => value 
2. Callbacks can be listed via `GET /api/builds/<id>/callbacks`
3. Currently the only supported event is `ON_BUILD_FINISHED`, which is triggered in `UpdateProjectStatus` in case `projectstatus.build.finished` is `True`. It's easy to add up more events


chaws:
It's failing because of pytest missing package when running dev-docker. Will fix this on Monday

amrohassaan:
LGTM!

Issue #829, opened 2020-09-03
multi-login option


chaws:
Maybe we could have an option to allow users login via github/gitlab account? 
Not a high priority for now

mwasilew:
There are a few open questions here:
 - should we combine user accounts using different auth methods? For example if I have Linaro account and github account, should they be combined so I could authenticated with either provider?
 - do we even want to allow multiple auth providers at the same time?
 - how to manage access to projects for users with different auth providers (this is mainly for qa-reports.l.o)
 - what to use as username?

chaws:
> There are a few open questions here:
> 
> * should we combine user accounts using different auth methods? For example if I have Linaro account and github account, should they be combined so I could authenticated with either provider?
> * do we even want to allow multiple auth providers at the same time?

That's what I had in mind. Perhaps a double login screen with ldap on the left and github/gitlab oauth login

> * how to manage access to projects for users with different auth providers (this is mainly for qa-reports.l.o)
> * what to use as username?

I guess non-linaro users could have email as username, this would eliminate chances of having usernames provided by e.g. github to collide with a pre-existing ones from Linaro. Regarding managing projects, I guess it would be just like managing regular users.

We'd have to implement restricting backends to specific users/groups maybe.

On another hand, do we want to store data from non-linaro folks?

roxell:
there's a new pipeline for building and testing kernel's on gitlab, where anyone should be able to fork that and use it.

What they need to do to get going are, forking their kernel tree to their gitlab "space" ask for a TuxBuild token. From this point they can do build testing when something changes in the kernel repository they mirrored.
TuxBuild is a powerful tool to build Linux kernel quickly and the user have full control what to build and how to configure their kernel.
Next step in the pipeline is to run tests on QEMU's or real HW.
That is done by using SQUAD and to view the test results too.
Right now a user that haven't a linaro.org accont, he has to create a ticket to ask for an account, then when he gets a linaro account he can login to SQUAD and setup that part of the pipeline.
If we had a way to setup the account via gitlab Oauth that would make it easier.

chaws:
@mwasilew I've looked some options to implement this feature, saw some packages that implement it but the one that looks more promising/stable is this one https://django-allauth.readthedocs.io/en/latest/. I'll work on a prototype tomorrow.

PR #953, opened 2021-01-13
socialaccount: add Github and Gitlab social account login

chaws:
Fixes https://github.com/Linaro/squad/issues/829

Answering @mwasilew questions:

> should we combine user accounts using different auth methods? For example if I have Linaro account and github account, should they be combined so I could authenticated with either provider?

That could be done, but I'm going with **no**. I don't think a user being logged in via Github/Gitlab should be authenticated as any Linaro user. It feels wrong, dunno why.

> do we even want to allow multiple auth providers at the same time?

That's done! Users can authenticate via Github or Gitlab

> how to manage access to projects for users with different auth providers (this is mainly for qa-reports.l.o)

Users will be granted default `squad` permissions, allowing them to create their own user space and with that they can submit jobs, generate their own auth-token. If further permissions are required, so Group managers should do accordingly.

> what to use as username?

I opted for using email, as this will be used to determine if user already exists or not. If a user uses the same email for both Gitlab and Github accounts, he/she can use both methods to log to the same SQUAD user. When such users create their own user namespace, the symbol `@` will be removed from it, as it's not a match for `group_slug` regex pattern.

That's how the login page will look like:
Screenshot from 2021-01-12 23-50-45

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,6 +12,7 @@ RUN apt-get update -q=2 && \
         python3-dateutil \
         python3-dev \
         python3-django \
+        python3-django-allauth \
         python3-django-auth-ldap \
         python3-django-cors-headers \
         python3-django-celery-results \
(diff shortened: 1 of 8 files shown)


amrohassaan:
LGTM

PR #957, opened 2021-01-21
Rollback debian bullseye

chaws:
I upgraded our debian base docker image to bullseye in order to get newer packages, but mainly to satisfy django-allauth's requirement of having Django 2.2. Doing so proved to cause some compatibility issues mainly with boto and ldap connections.

I reverted that commit and started using -backports instead.


chaws:
Tests passed, i'm merging this now to fix ldap in staging

PR #959, opened 2021-01-28
core: notification: mark as notified even if no recipients available

chaws:
This prevents gerrit notifications from being sent on every new coming testjob


amrohassaan:
LGTM

PR #971, opened 2021-03-12
[Don't merge it yet] core: notification: use fast TestComparison

chaws:
Passing `regressions_and_fixes_only=True` to TestComparison will force it to use a fast way to compute regressions and fixes
thus improving performance of overall Notifications

I'll run a bigger round of tests before we can merge this.


amrohassaan:
LGTM :+1:

Issue #973, opened 2021-04-06
Put callback status in build view


jscook2345:
It would be nice if the build view in squad showed the status of all callbacks since we can only currently see it by digging through the squad api.

Example build view: https://qa-reports.linaro.org/~justin.cook/mirror-linux-next-master-sanity/build/next-20210401/

Example failing callback: https://qa-reports.linaro.org/api/builds/63128/callbacks/

PR #978, opened 2021-04-07
frontend: build_callbacks: add callbacks to the Build view

chaws:
Fixes https://github.com/Linaro/squad/issues/973

And it'll look like this:

Screenshot from 2021-04-07 17-59-19

### DIFF
--- a/squad/frontend/templates/squad/build-nav.jinja2
+++ b/squad/frontend/templates/squad/build-nav.jinja2
@@ -66,6 +66,13 @@
         </a>
     </li>
     {% endif %}
+    {% if project.writable_by(user) and build.callbacks %} <!-- REMEMBER TO ADD .count() after callbacks -->
+    <li role="presentation" {% if url_name == 'build_callbacks' %}class="active"{% endif %}>
+        <a href="{{build_section_url(build, 'build_callbacks')}}">
+            {{ _('Callbacks') }}
+        </a>
+    </li>
+    {% endif %}
     {% if project.writable_by(user) %}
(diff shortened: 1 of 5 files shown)


jscook2345:
Looks good to me, with a few notes.

1. I think `jinja` automatically takes care of escaping variables for you. If not, or if you have a setting disabling that, some of the variables may need to be escaped.

2. It looks like their is a travis failure having to due with a locked table.

3. What does this look like with zero callbacks?

4. What does this look like with more than one callback?

Thanks for your work on this!

chaws:
> 1. I think jinja automatically takes care of escaping variables for you. If not, or if you have a setting disabling that, some of the variables may need to be escaped.

It's configured for auto-escape template strings: https://github.com/Linaro/squad/blob/master/squad/jinja2.py#L41. I had that concern when presenting the `Response content`, which should contain JSON only, but who knows! I think it's protected against that. Thanks for pointing it out tho.

> 2. It looks like their is a travis failure having to due with a locked table.

Dunno why that happens, it's not every time. For some reason I can't restart the job, probably because travis is migrating to *.com, I might have to enable github actions for squad. Bottom line is this error is "OK", since it breaks in sqlite while in postgres (production) it's green.

> 3. What does this look like with zero callbacks?

No "Callbacks" tab.

> 4. What does this look like with more than one callback?

See the bullet in front of "POST / ...", there's a loop iterating over all `build.callbakcs` and each callback will be an `<li>`, just like that first one you see.

chaws:
@jscook2345 Just fixed the issue that was breaking travis.

jscook2345:
LGTM! Merge away! 🥇

PR #979, opened 2021-04-08
frontend: project_settings: add important-metadata-keys

chaws:
Fixes https://github.com/Linaro/squad/issues/970


chaws:
This will look like this (important metadata keys added):
Screenshot from 2021-04-08 09-43-49

All tests have passed, I'm merging it

Issue #966, opened 2021-02-23
build next-20210212 baseline seems incorrect


jscook2345:
https://qa-reports.linaro.org/api/builds/58859/

This build (`next-20210212`) listed it's baseline build as (`next-20210112`). This seems incorrect to me since there were several builds between the two.

chaws:
Yeah, this looks weird indeed. Unfortunately it's a bit harder to fix this problem and here's why:

1. https://github.com/Linaro/squad/blob/master/squad/core/models.py#L1159, `previous_build` is retrieved when a build comparison is requested, usually when it's finished.
1.1. The problem here is that the builds before `next-20210212` and after `next-20210112` might not have been finished by the time `previous_build` of `next-20210212` was queried, thus making the baseline point to `next-20210112`
2. I say it's tricky because it's been a long-running issue in SQUAD: determining if a build is finished
2.1. There are several reasons why a build is not in `finished` state:
* not all test-jobs are completed
* some test jobs failed and it wasn't replaced by a working one
* the number of expected testjobs per environment hasn't been reached (this is more rare, but still a reason)

chaws:
@jscook2345 did you see that pattern again recently? 

I'm about to implement something that I think will fix behaves like that. I think it's the similar to what happened here: https://github.com/Linaro/squad/issues/975

jscook2345:
I have not seen it, but I haven't been actively looking for it. Thanks for the update!

Issue #975, opened 2021-04-07
Reference build "stuck", getting constant email notifications against it


pfalcon:
For https://qa-reports.linaro.org/lite/zephyr-micropython/ , I'm getting email notifications every day with the following summary: `lite/zephyr-micropython: 609 tests, 0 failed, 381 passed,  228 skipped` (same every day) and following change higlight:

![Screenshot at 2021-04-07 12-09-18](https://user-images.githubusercontent.com/500451/113841182-223b4700-979a-11eb-9903-4b5312035e60.png)

As can be seen, `#1514` build was selected as a comparison reference, which happened to have failed. Following builds succeeded, but as they get compared with `#1514`, email notification is sent with differences.

But the issue is that the reference build itself should be updated to a newer build, and that doesn't happen.

chaws:
I've subscribed myself to the notifications so I can get it the next time.

I'm looking into it now

chaws:
Squad will send you a report of a build when it's finished or after a certain timeout, which is 6 hours for lite/zephyr-micropython.

In the report/email, squad will compare the current build against the most recent finished build. A build is finished when all its Jobs are done in LAVA. From what I checked, the most recent one is #1514.

All builds after #1514 have not received any answer from LAVA wrt disco-l475-iot1 jobs. It seems that this device is not available in LAVA: https://validation.linaro.org/scheduler/device_type/disco-l475-iot1

pfalcon:
@chaws: Thanks for investigating! Ack on disco-l475-iot1 being down. I have 2 questions:

> Squad will send you a report of a build when it's finished or after a certain timeout, which is 6 hours for lite/zephyr-micropython.
> In the report/email, squad will compare the current build against the most recent finished build.

1. Why "after a certain timeout (which is 6 hours here)" rule doesn't apply to determination of "finished build". I'd say, that would be intuitive behavior I'd expect: if I set timeout to 6 hours, I'd like it to be "finished" in 6 hours, repsonsed received, or not. In other words, is there any issue to fix on SQUAD side, or we just learn to manually update SQUAD's project config for such situation?
2. How to update SQUAD's project config to workaround this issue. I.e., we'd like to mark disco-l475-iot1 as "don't consider"/"don't expect results". (The devices have hardware problems, we expecte them to be recovered, but ETA is not known currently, i.e. can be a few months).

Thanks!

chaws:
[...]

PR #980, opened 2021-04-08
core: models: add flag to allow forcing builds to finish

chaws:
Fixes https://github.com/Linaro/squad/issues/966 and parts of https://github.com/Linaro/squad/issues/975

The flag `force_finishing_builds_on_timeout` was added to Project model to signal whether or not builds should be marked as finished when Notification timeout is reached.


chaws:
The build is failing due to pull limit. I'm merging this and see if it still breaks in master

PR #981, opened 2021-04-08
Use `expected_testruns == -1` to disable and environment for receiving new submissions

chaws:
This patch let users disable environments in projects, thus avoiding test job submissions to such environments.

It's a WIP because it needs some extra discussion.


chaws:
All checks are passing, merging

Issue #988, opened 2021-05-18
TestJobs get stuck when backends timeout


chaws:
Today we had an issue where squad got the queue stuck because lava.nxp is out, so squad raises an exception due to time out and the job is retried over and over, the attempts counter is never updated!

PR #989, opened 2021-05-18
Raise TemporaryFetchIssue on fetch timeout

chaws:
Fix https://github.com/Linaro/squad/issues/988

By not handling Timeouts, TestJob.fetch_attempt was never increased so the job would be retried over and over and over.

Now with correct exception handling, Backend.fetch() will correctly increase TestJob.fetch_attempts whenever there's a Lab downtime.

### DIFF
--- a/squad/ci/backend/lava.py
+++ b/squad/ci/backend/lava.py
@@ -144,6 +144,12 @@ class Backend(BaseBackend):
                 raise FetchIssue(self.url_remove_token(str(fault)))
         except ssl.SSLError as fault:
             raise FetchIssue(self.url_remove_token(str(fault)))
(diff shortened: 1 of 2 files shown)


jscook2345:
LGTM, but travis is not happy 😠 !

chaws:
Thanks for checking this out!

Yeah, travis has been a bit stingy recently, I've restarted jobs

chaws:
I've temporarily disabled testing on real LAVA https://github.com/Linaro/squad/commit/8072874ee9c12ad904a0ba44256f27a741430f66, this is just an extra testing environment. Virtual LAVA tests are still good enough.

chaws:
Tests are all green now, merging

PR #991, opened 2021-06-08
ci: backend: lava: handle null error_msg

chaws:
Sometimes LAVA might not populate all the metadata as expected. This patch makes sure not to break when suck scenario is reached.

@stevanradakovic @ivoire do you know if this value has changed in recent LAVA releases? I spot dozens of thousands of errors like that starting last week

### DIFF
--- a/squad/ci/backend/lava.py
+++ b/squad/ci/backend/lava.py
@@ -671,7 +671,8 @@ class Backend(BaseBackend):
                 self.log_debug("'%s' is not a valid regex" % message_re)
 
         for regex in infra_messages_re_list:
-            if regex.search(metadata['error_msg']) is not None and \
+            error_msg = metadata.get('error_msg')
+            if error_msg and regex.search(error_msg) is not None and \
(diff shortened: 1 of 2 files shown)


chaws:
Found an error job: https://lkft.validation.linaro.org/scheduler/job/2849191

> [2849191] Result metadata is too long. {case: job, definition: lava, error_msg: "Kernel panic - not syncing: Attempted to
    kill init! exitcode=0x0000000b\r\r\n[    1.099235] CPU: 1 PID: 1 Comm: init Not
    tainted 5.13.0-rc4-next-20210604 #1\r\r\n[

chaws:
I'm merging this as it seems to be a specific problem

ivoire:
The metadata field is limited to 4096 characters. Prior to 2021.03, if the content was larger than 4096 characters, lava would just drop it. This is usually happening with kernel crashes.

Since https://git.lavasoftware.org/lava/lava/-/merge_requests/1446 the content is partially preserved.

PR #993, opened 2021-06-11
core: tasks: add support for patterns in known issue test_name

chaws:
This will allow known issues maintainers to provide pattern as test_name so that it can match multiple suites, or multiple tests
patterns.

For example, a known issue with test_name equals to `ltp-*/foo` will match ALL incoming new tests `ltp-syscalls/foo` or `ltp-timers/foo`. 

@nareshkamboju @roxell thoughts?

### DIFF
--- a/squad/core/tasks/__init__.py
+++ b/squad/core/tasks/__init__.py
@@ -1,8 +1,10 @@
 from django.core.exceptions import MultipleObjectsReturned
 from django.utils import timezone
 from collections import defaultdict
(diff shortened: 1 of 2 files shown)


nareshkamboju:
@chaws  Thanks for working on this.
Looks good to me.
+1.

chaws:
It seems to be working fine: https://qa-reports.linaro.org/api/tests/1695910348/

PR #994, opened 2021-06-28
Move to github actions

chaws:
This PR moves SQUAD's CI from Travis to Github actions

I moved all testing I could find in .travis.yml, but testing on real LAVA is still pending since about a couple of months ago.


amrohassaan:
@chaws LGTM.

Issue #997, opened 2021-07-08
Disable "resubmit" button of jobs that already got resubmitted


chaws:
YongQin finds it confusing to have the "resubmit" option for jobs that already got resubmitted, specially if multiple re-submissions are needed.

liuyq:
Hi, @chaws,
like for job1, either it's resubmitted by SQUAD automatically or resubmitted manually from webpage or api,  we will get a new job 2, in this time, we would like to have the "resubmit" button for job 1 disabled.
and more further, if the job 2 gets resubmitted and get a new job of job 3 for some reason, then the resubmit for job2 should be disabled too, and only job 3 could be resubmitted.

And in the output of api call, like here: https://qa-reports.linaro.org/api/testjobs/5337877/, 
would like to have some some flag to indicate if the job is a parent job, so that we would know if the result for these jobs should be ignored easily.

liuyq:
remember @mrchapp has some concerns on this "disabling the resubmit" feature?

PR #998, opened 2021-07-12
Disable resubmit button

chaws:
Fix https://github.com/Linaro/squad/issues/997

This patch will disable the "resubmit" option of jobs that already got resubmitted. Things will look like this:

Screenshot from 2021-07-12 09-08-54

The patch also hides the "cancel" button on Jobs that are fetched already (it only makes sense to cancel jobs that aren't yet finished)

### DIFF
--- a/squad/frontend/templates/squad/testjobs.jinja2
+++ b/squad/frontend/templates/squad/testjobs.jinja2
@@ -111,16 +111,18 @@
                         </a>
                     </div>
                     {% endif %}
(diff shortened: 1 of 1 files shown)


jscook2345:
LGTM, but I would try re-running some of those failed checks.

liuyq:
@chaws thanks for the work!
From the image here, Just thought of that, maybe hiding them is better than disable?

PR #1000, opened 2021-09-24
Fetch job as they are watched

chaws:
This patch adds a fetch task immediately as a job is triggered for watching, thus causing watched jobs to be fetched right away.


chaws:
Now it's tested, will merge once tests are green

jscook2345 (approved):

chaws:
I fixed the issue with importlib-metadata, merging this now

PR #1006, opened 2021-10-30
Enable metrics regression

chaws:
This is a huge PR, I wish I made it smaller, but it was hard to split it, so what I did was to split the work into meaningful commits. It:
1. Removes Metric.name column from the database, as it's already present in SuiteMetadata (this takes up lots of DB space!)
2. Adds Project reference to MetricThreshold and allows `environment=None` so that SQUAD can support thresholds across all environments in a project
3. Supports regressions and fixes in MetricComparison (and it should be quite fast doing so :)
3.1. Now one can get that by using `by=metrics` in `/api/projects/<id>/compare_builds/` endpoint
3.2. The result of regressions and fixes are also stored in ProjectStatus and can be accessed via `/api/builds/<id>/status/` endpoint


chaws:
Tests are all green now. I'm merging this and roll it out to staging

squad/core/comparison.py
```diff
@@ -54,6 +54,49 @@ def __init__(self, *builds):
...
+                'values': {
+                    'baseline_id': baseline.id,
+                    'target_id': target.id
+                }
+            }
+
+            # If builds belong to different projects, environment comparison should
+            # use slug
+            if self.__same_projects__():
+                self.base_sql['where'].append('target.environment_id = baseline.environment_id')
```
jscook2345, line 92: `from` seems like it would be empty here. Would that be a problem?

squad/core/comparison.py
```diff
@@ -54,6 +54,49 @@ def __init__(self, *builds):
...
+            else:
+                self.base_sql['from'].append('core_environment baseline_environment')
```
jscook2345, line 94: are spaces separating tables OK here?

jscook2345: Or should it be two appends?

squad/core/templates/squad/notification/diff.html.jinja2
```diff
@@ -81,6 +81,7 @@
     <h2>{{threshold_metric[0].name}}</h2>
     <ul>
       <li>Environment: {{threshold_metric[0].environment.slug}}</li>
+      <li>Metric name: {{threshold_metric[1].full_name}}</li>
```
jscook2345, line 84: Why do you use both `0` and `1` indexes for the same data section? Seems odd. Maybe a comment?

squad/frontend/static/squad/controllers/metricThreshold.js
```diff
@@ -25,17 +24,28 @@ function getKeyByValue(object, value) {
 }
 
 function MetricThresholdController($scope, Threshold) {
+    $scope.openNewThresholdModal = function() {
+        $scope.currentThreshold = new Threshold()
+	$scope.currentThreshold.environment = 'all'
```
jscook2345, line 29: missing leading spaces

squad/frontend/static/squad/controllers/metricThreshold.js
```diff
@@ -25,17 +24,28 @@ function getKeyByValue(object, value) {
...
+    }
 
     $scope.initMetricThresholds = function() {
         // Initialize threshold for modal dialog.
         $scope.currentThreshold = new Threshold()
+	$scope.currentThreshold.environment = 'all'
```
jscook2345, line 35: missing leading spaces

squad/frontend/static/squad/controllers/metricThreshold.js
```diff
@@ -47,8 +57,16 @@ function MetricThresholdController($scope, Threshold) {
         var threshold_index = $scope.currentThreshold.index
```
jscook2345, line 57: inconsistent leading spaces in this section

test/core/test_metric_comparison.py
```diff
@@ -134,3 +157,79 @@ def test_multiple_values_same_metric(self):
...
+            'suite_a/improved-metric-lower-better': (2, 1, True),
+            'suite_a/regressing-metric-higher-better': (2, 1, False),
+            'suite_a/regressing-metric-lower-better': (1, 2, False),
+            'suite_a/stable-metric': (1, 1, None),
+            'suite_a/thresholdless-metric': (1, 2, None),
+            'suite_a/valueness-threshold-metric': (1, 2, None),
+        }
+
+        for metric_name in test_cases.keys():
+            build_a_result = test_cases[metric_name][0]
```
jscook2345, line 176: I think this could be unpacked like this:

```
build_a_result, build_b_result, expected = test_cases[metric_name]
```

But the current code may be more verbose.

test/core/test_metric_comparison.py
```diff
@@ -134,3 +157,79 @@ def test_multiple_values_same_metric(self):
...
+            self.receive_test_run(self.project, self.build_b.version, self.environment_a.slug, {
+                metric_name: build_b_result,
+            })
+
+            comparison = MetricComparison(self.build_a, self.build_b, regressions_and_fixes_only=True)
+            if expected is True:
+                self.assertIn(metric_name, comparison.fixes[self.environment_a.slug])
+            elif expected is False:
+                self.assertIn(metric_name, comparison.regressions[self.environment_a.slug])
+            else:
```
jscook2345, line 195: I might add a comment saying `# None` here. I got confused for a minute :)

jscook2345 (commented):
I added a few comments, but my guess is they are not super scary.

Some of the javascript changes look like the indenting is incorrect, but maybe that's just github?

Looks great, thanks for all your work @chaws !

squad/core/comparison.py
chaws: That is expected. MetricComparison will fill that up with `core_metric` whereas TestComparison will do the same with `core_test`. The comparison approach is 99% the same for both test and metrics, except that each type need to use their own table names. Good observation tho!

squad/core/comparison.py
chaws: That's SQL standard
```sql
SELECT * FROM my_table AS alias;

-- The "AS" can be omitted here
SELECT * FROM my_table alias;
```

squad/core/models.py
```diff
@@ -1100,13 +1099,53 @@ def __str__(self):
...
     is_higher_better = models.BooleanField(default=False)
 
+    def _check_duplicates(self):
+        """
+        We have to make sure of the following
+        - a project-wide threshold CANNOT colide to an environment-specific one
+          We need to check if there's any environment that matches the thresold
+
+        - an environment-specific threshold CANNOT colide to a project-wide one
+          We meed tp check if there's already a project-wide threshold
```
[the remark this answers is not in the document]
chaws: Nice catch! Fixed it

squad/core/models.py
```diff
@@ -1100,13 +1099,53 @@ def __str__(self):
...
+        """
+        project_wide = self.environment is None
+        existingThresholds = MetricThreshold.objects.filter(
+            name=self.name,
+            value=self.value,
+            is_higher_better=self.is_higher_better,
+            project=self.project,
+            environment__isnull=not project_wide)
```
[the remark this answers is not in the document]
chaws: Yeah, it looks weird to have spaces like that, but it feels so smooth reading it as regular text :p

squad/core/templates/squad/notification/diff.html.jinja2
chaws: That's actually a tuple and the object gets built here: https://github.com/Linaro/squad/blob/1.43.1/squad/core/models.py#L1342

I scratched the back of my head exactly like you did and I don't like it how it looks, but I didn't want to refactor that atm.

squad/frontend/static/squad/controllers/metricThreshold.js
chaws: Fixed

squad/frontend/static/squad/controllers/metricThreshold.js
chaws: Fixed

squad/frontend/static/squad/controllers/metricThreshold.js
chaws: Fixed this one too, there was actually a lot more lines like this one. Nice catch

test/core/test_metric_comparison.py
chaws: :P

chaws:
Thanks for your review, @jscook2345!
I'm applying the changes in a new PR I'm making to add a few extra tweaks to get the metric regressions useful to you guys.

PR #1007, opened 2021-11-04
api: Allow to update TestJob definition via API

mwasilew:
This patch enables updating TestJob.definition field using REST API.
Before that it wasn't possible. 

Signed-off-by: Milosz Wasilewski <milosz.wasilewski@foundries.io>

### DIFF
--- a/squad/api/rest.py
+++ b/squad/api/rest.py
@@ -1441,6 +1441,16 @@ class TestJobSerializer(DynamicFieldsModelSerializer, serializers.HyperlinkedMod
         fields = '__all__'
 
 
+class TestJobUpdateSerializer(serializers.HyperlinkedModelSerializer):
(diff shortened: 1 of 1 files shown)


squad/api/rest.py
```diff
@@ -1441,6 +1441,13 @@ class Meta:
         fields = '__all__'
 
 
+class TestJobUpdateSerializer(serializers.HyperlinkedModelSerializer):
```
chaws, line 1444: Could you please propagate read-only fields:
```python
id = serializers.IntegerField(read_only=True)
external_url = serializers.CharField(source='url', read_only=True)
```

mwasilew: Yeah, makes sense. This makes the API consistent.

chaws:
LGTM! Thanks!

PR #1010, opened 2021-11-25
View testjobs progress

chaws:
This patch allows SQUAD users to follow up which testjobs for a given build are completed, failed, running or still queued.

Below is a snapshot of how it looks like:

Screenshot from 2021-11-25 16-13-34

And below is another snapshot of the same view, this time for each environment:
Screenshot from 2021-11-25 16-13-55

I have also added an interval function in the browser that keeps updating the progress bar every MINUTE until the progress is 100%.

### DIFF
--- a/squad/api/rest.py
+++ b/squad/api/rest.py
@@ -849,6 +849,12 @@ class BuildViewSet(NestedViewSetMixin, ModelViewSet):
 
         List of test jobs in the build (if any)
 
+     * `api/builds/<id>/testjobs_summary` GET
+
+        List of test jobs in the build summarized by job_status.
+        If `per_environment` is specified in the query string, the summary
+        will be split by environments.
+
      * `api/builds/<id>/cancel` POST
 
         Cancel all test jobs of the build (if any)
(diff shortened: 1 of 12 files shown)


mrchapp:
WOW, this looks fantastic!

nareshkamboju:
@chaws  This is super useful for me !

jscook2345:
Code and demo LGTM. If @mrchapp and @nareshkamboju approve, I think it's safe to merge this.

chaws:
Thanks for review, @jscook2345 ! Please don't merge it yet as I need to apply feedback from @mrchapp and @nareshkamboju . The backend code won't change much, but some of JS will.

chaws:
Hey guys, I finished the update and it's looking like this:

Screenshot from 2021-12-03 15-49-07

I exaggerated the dragonboard env so it can stand out like a sore thumb. But this will keep updating on the fly and the bar width should change based on the number of testjobs as well.

If you like this now, I can merge it and we give it a go. Feedbacks are always welcome :)

mrchapp (approved):

mrchapp:
I'm all for it. Let's take it for a test drive and see how it feels. Thanks!

chaws:
Ack! I'll just add a bit more tests to keep coverage above 80%, then I'll merge and roll it out to qa-reports.

PR #1021, opened 2022-01-02
all: add a page to view the metrics for a build

jscook2345:

### DIFF
--- a/squad/frontend/metrics.py
+++ b/squad/frontend/metrics.py
@@ -1,0 +1,38 @@
+from django.shortcuts import render
+
+from squad.http import auth
+from squad.core.models import Environment, Metric
+from squad.frontend.views import get_build
+
+
+@auth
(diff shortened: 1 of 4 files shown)


jscook2345:
127 0 0 1_8000_lkft_linux-next-oe_build_next-20181005_metrics_

chaws:
LGTM. 

@jscook2345 Thanks for the patch!

Just two things worth mentioning for future patches:
1. Try to use file name in the commit message. It doesn't always work, but in the case of this commit you could've done `frontend: metrics: add a page to view the metrics for a build`
2. Use related relationship helpers: Example
```python
environments = Environment.objects.filter(project=project).order_by("name")
metrics = Metric.objects.filter(build=build)
```
You could've used the related fields
```python
environments = project.environments.order_by("name")
metrics = build.metrics.all()
```
It's a bit cleaner

PR #1023, opened 2022-01-17
Lava api errors

mwasilew:
This is a series of patches that deals with errors resulting from using LAVA's REST API.

### DIFF
--- a/squad/ci/backend/lava.py
+++ b/squad/ci/backend/lava.py
@@ -337,7 +337,7 @@ class Backend(BaseBackend):
                 return False
         else:
             response = requests.post(
-                urljoin(self.api_url_base, "jobs/%s/cancel" % (job_id)),
+                urljoin(self.api_url_base, "jobs/%s/cancel/" % (job_id)),
                 headers=self.authentication,
                 timeout=self.settings.get(timeout_variable_name, DEFAULT_TIMEOUT)
             )
@@ -357,7 +357,7 @@ class Backend(BaseBackend):
         if self.use_xml_rpc:
             return self.proxy.scheduler.resubmit_job(job_id)
         response = requests.post(
-            urljoin(self.api_url_base, "jobs/%s/resubmit" % (job_id)),
+            urljoin(self.api_url_base, "jobs/%s/resubmit/" % (job_id)),
             headers=self.authentication,
             timeout=self.settings.get(timeout_variable_name, DEFAULT_TIMEOUT)
         )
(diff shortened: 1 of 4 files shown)


test/ci/backend/test_lava.py
```diff
@@ -1257,3 +1258,21 @@ def test_resubmit_job_lava_error(self, __resubmit__):
...
+        testjob = TestJob(
+            definition=test_definition,
+            backend=self.backend,
+            target=self.project,
+            submitted=True,
+            job_status='Canceled',
+            job_id="1236"
+        )
+        with requests_mock.Mocker() as m:
+            m.post("http://example.com/api/v0.2/jobs/1236/resubmit",
```
chaws, line 1275: I saw this error: `requests_mock.exceptions.NoMockAddress: No mock address: POST http://example.com/api/v0.2/jobs/1236/resubmit/`

Maybe `requests_mock` is adding the extra slash there? And it seems the requested URL and the mocked one need to be exactly the same: https://stackoverflow.com/questions/62981932/requests-mock-nomockaddress-exception

squad/ci/backend/lava.py
```diff
@@ -357,7 +357,7 @@ def __resubmit__(self, job_id):
         if self.use_xml_rpc:
             return self.proxy.scheduler.resubmit_job(job_id)
         response = requests.post(
-            urljoin(self.api_url_base, "jobs/%s/resubmit" % (job_id)),
+            urljoin(self.api_url_base, "jobs/%s/resubmit/" % (job_id)),
```
chaws, line 360: While you're at it, could you please also add the trailing slash here https://github.com/Linaro/squad/blob/b2c17c7c333b37a4a95d33c9ce75fbdc96c98a42/squad/ci/backend/lava.py#L377

test/ci/backend/test_lava.py
mwasilew: it's my mistake. I forgot to update the address when adding a trailing slash into the code.

squad/ci/backend/lava.py
mwasilew: I guess I can. This one isn't affected by redirect as it's GET not POST.

test/ci/backend/test_lava.py
```diff
@@ -1243,3 +1244,35 @@ def test_resubmit_canceled_job(self, __resubmit__):
...
+            definition=test_definition,
+            backend=self.backend,
+            target=self.project,
+            submitted=True,
+            job_status='Canceled',
+            job_id="1236"
+        )
+        with requests_mock.Mocker() as m:
+            m.post("http://example.com/api/v0.2/jobs/1236/resubmit/",
+                    status_code=405,
```
chaws, line 1276: I think flake8 is complaining about the indentation here. My guess is that it's expecting 1 tab, but there's 2.

chaws:
LGTM, thanks for the patch @mwasilew !

PR #1024, opened 2022-01-26
[WIP]: Adding TuxSuite support for SQUAD

chaws:
This is a first draft of TuxSuite support for SQUAD.

The idea is to have squad-client [send watchjob requests](https://github.com/Linaro/squad-client/pull/122) to SQUAD using the following notation for job_id:
```
* BUILD:linaro@anders#1yPYGaOEPNwr2pCqBgONY43zORq
* TEST:arm@bob#1yPYGaOEPNwr2pCqBgONY43zORp
```

As soon as SQUAD receives these jobs, the TuxSuite backend will know how to fetch its results using TuxSuite api. Therefore, metadata, metrics, logs and tests are going to be processed accordingly.

For jobs from TuxBuild results:
* There'll be created:
  * 1 test: `build/gcc-9-defconfig` (example)
  * 2 metrics: `build/gcc-9-defconfig-duration` and `build/gcc-9-defconfig-warnings`
 
For jobs from TuxTest results:
* There'll be created as many suites as there are in the `/results` endpoint. The `lava` suite will be ignored.
  * There'll be created as many tests as there are in each suite 

For now it seems that the log lines in TuxRun are not matching the actual log file. I already raised a ticket in TuxRun pointing that out.

@jscook2345 I think this kinda covers what you proposed in https://github.com/Linaro/squad-client/pull/143. I actually reused most of your stuff to create build test results. Perhaps I can adopt your ideas in this current patch to SQUAD and ideally we could deprecate sending individual tuxbuild results via squad-client.


chaws:
@jscook2345 almost forgot! I liked your idea of retrieving each config from the url in `kconfig`, I even went a step further and grepped only lines that start with and actual `CONFIG`, then sorted that out and finally got a hash out of that. I just thought it's as concise as it can be. Maybe that's not needed at all.

ivoire:
@chaws do you also support plans?

chaws:
@ivoire not yet.

PR #1027, opened 2022-02-09
frontend: failures: compute a confidence score

jscook2345:
It is difficult to tell if a test failure is a true regression or a
flakey test. To try and address this, we create a new view of test
failures for a build that displays a confidence score.

A confidence score is the ratio of the number of times the test has
passed over some number number of times the test has been run.

### DIFF
--- a/squad/core/models.py
+++ b/squad/core/models.py
@@ -563,6 +563,32 @@ class Build(models.Model):
             return False
         return True
 
+    __failures__ = None
+
+    @property
+    def failures(self):
+        if self.__failures__ is None:
+            failures = self.tests.filter(result=False).exclude(has_known_issues=True).prefetch_related("metadata", "environment")
+
+            limit = self.project.build_confidence_count
+            builds = self.project.builds.filter(
+                id__lt=self.id,
+            ).order_by("-id").only("id")[:limit]
+
+            history = Test.objects.filter(
+                metadata__in=[f.metadata for f in failures],
+                environment__in=[f.environment for f in failures],
+                build_id__in=list(builds.values_list("id", flat=True)),
+            ).order_by("-build_id").prefetch_related("metadata", "environment").defer("log")
+
+            for f in failures:
+                f_history = [t for t in history if t.metadata == f.metadata and t.environment == f.environment]
+                f.set_confidence(self.project.build_confidence_threshold, f_history)
+
+            self.__failures__ = failures
+
+        return self.__failures__
+
     @property
     def finished(self):
         """
@@ -978,6 +1004,35 @@ class Test(models.Model):
         else:
             suite = self.metadata.suite
         return join_name(suite, self.name)
+
+    class Confidence(object):
+        def __init__(self, threshold, tests):
+            self.threshold = threshold
+            self.tests = tests
+
+        @property
+        def count(self):
+            return len(self.tests)
+
+        @property
+        def passes(self):
+            return sum(1 for t in self.tests if t.status == "pass")
+
+        @property
+        def score(self):
+            return 100 * (self.passes / self.count)
+
+    __confidence__ = None
+
+    def set_confidence(self, threshold, tests):
+        self.__confidence__ = Test.Confidence(
+            threshold=threshold,
+            tests=tests,
+        )
+
+    @property
+    def confidence(self):
+        return self.__confidence__
 
     class History(object):
         def __init__(self, since, count, last_different):
--- a/squad/frontend/failures.py
+++ b/squad/frontend/failures.py
@@ -1,0 +1,37 @@
+from django.shortcuts import render
+
+from squad.http import auth
+from squad.frontend.views import get_build
+
+
+@auth
+def failures(request, group_slug, project_slug, build_version):
+    project = request.project
+    build = get_build(project, build_version)
+    failures = build.failures
+    environments = project.environments.order_by("slug")
+
+    search = request.GET.get('search', '')
+
+    if search:
+        failures = failures.filter(metadata__name__icontains=search)
+
+    unique_failures = sorted(set([t.full_name for t in failures]))
+
+    rows = {}
+    for t in build.failures:
+        if t.environment.slug not in rows:
+            rows[t.environment.slug] = {}
+
+        rows[t.environment.slug][t.full_name] = t
+
+    context = {
+        "project": project,
+        "build": build,
+        "environments": environments,
+        "ufailures": unique_failures,
+        "rows": rows,
+        "search": search,
+    }
+
+    return render(request, 'squad/failures.jinja2', context)
--- a/squad/frontend/templates/squad/failures.jinja2
+++ b/squad/frontend/templates/squad/failures.jinja2
@@ -1,0 +1,82 @@
+{% extends "squad/base.jinja2" %}
+
+{% block content %}
+
+<div ng-app='Build'>
+{% include "squad/build-nav.jinja2" %}
+</div>
+
+<h2>{{ _('All test failures') }}</h2>
+
+<div>
+    <div class='row row-bordered'>
+        <div class='col-md-12 col-sm-12 filter'>
+          <a id="searchLink"><button type='button' class='btn btn-primary fa fa-search'></button></a>
+          <input name='search' id='search' type='text' placeholder='{{ _('Filter results ...') }}' value='{{search}}' />
+        </div>
+    </div>
+
+    <table class='test-results'>
+        <thead>
+            <th>{{ _('Test') }}</th>
+            {% for e in environments %}
+            <th>{{ e.slug }}</th>
+            {% endfor %}
+        </thead>
+        {% if not ufailures %}
+        <tr>
+            <td colspan="{{environments|length|add(1)}}" class="alert alert-warning">
+                <em>{{ _('This build has no failures yet.') }}</em>
+            </td>
+        </tr>
+        {% else %}
+            {% for u in ufailures %}
+              <tr>
+                <td>{{ u }}</td>
+                {% for e in environments %}
+                {% if e.slug in rows and u in rows[e.slug]: %}
+                  <td class="{{ rows[e.slug][u].status }}">
+                    {{ rows[e.slug][u].confidence.score }}
+                    <span data-toggle="tooltip" title="{{ _('Click to display confidence info') }}">
+                      <button class="fa fa-info-circle popover-regressions-fixes"></button>
+                      <span title="{{ _('Confidence') }}" class="hidden">
+                        Pass: {{ rows[e.slug][u].confidence.passes }}
+                        Count: {{ rows[e.slug][u].confidence.count }}
+                        Threshold: {{ rows[e.slug][u].confidence.threshold }}
+                      </span>
+                    </span>
+                  </td>
+                {% else %}
+                  <td>&nbsp;</td>
+                {% endif %}
+                {% endfor %}
+              <tr>
+            {% endfor %}
+        {% endif %}
+    </table>
+
+</div>
+
+{% endblock %}
+
+{% block javascript %}
+<script type="module" src='{{static("squad/build.js")}}'></script>
+<script type="text/javascript" src='{{static("squad/table.js")}}'></script>
+<script type="text/javascript">
+$('[data-toggle="tooltip"]').tooltip();
+function loadSearchURL(pageParam, search) {
+  searchURL = pageParam + "&search=" + search;
+  window.location = searchURL;
+}
+$("#search").keypress(function(e) {
+  if(e.which == 13) {
+    window.location = "?page=1&search=" + $("#search").val();
+    return false;
+  }
+});
+$("#searchLink").click(function(event) {
+  window.location = "?page=1&search=" + $("#search").val();
+  return false;
+});
+</script>
+{% endblock %}
--- a/squad/frontend/templates/squad/build-nav.jinja2
+++ b/squad/frontend/templates/squad/build-nav.jinja2
@@ -56,6 +56,11 @@
             {{ _('All test results') }}
         </a>
     </li>
+    <li role="presentation" {% if url_name == 'failures' %}class="active"{% endif %}>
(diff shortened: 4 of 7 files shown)


jscook2345:
127 0 0 1_8000_lkft_linux-stable-rc-4 14-oe_build_v4 14 74_failures_ (1)

jscook2345:
I am concerned about the number of sql queries and the length of time it is taking. If you have some suggestions there that'd be great. Thanks!

squad/frontend/failures.py
```diff
@@ -0,0 +1,48 @@
...
+        tests = build.tests.filter(
+            result=False,
+            has_known_issues=False,
+            metadata__name__icontains=search,
+        ).prefetch_related('environment')
+    else:
+        tests = build.tests.filter(
+            result=False,
+            has_known_issues=False,
+        ).prefetch_related('environment')
```
terceiro, line 25: you can avoid this duplication by chaining filters, e.g.

```python
tests = build.tests.filter(
            result=False,
            has_known_issues=False,
).prefetch_related('environment')
if search:
    tests = tests.filter(metadata__name_icontains=search)
```

squad/core/models.py
```diff
@@ -979,6 +979,42 @@ def full_name(self):
             suite = self.metadata.suite
         return join_name(suite, self.name)
 
+    class Confidence(object):
+        def __init__(self, passes, count, threshold):
+            self.passes = passes
+            self.count = count
+            self.threshold = threshold
```
terceiro, line 986: `threshold` does not seem to be used for anything.

squad/core/models.py
```diff
@@ -979,6 +979,42 @@ def full_name(self):
...
+            return self.__confidence__
+
+        date = self.test_run.build.datetime
+        limit = self.build.project.build_confidence_count
+        tests = Test.objects.filter(
+            suite=self.suite,
+            metadata__name=self.name,
+            test_run__build__datetime__lt=date,
+            test_run__environment=self.test_run.environment,
+        ).order_by("-test_run__build__datetime")[:limit]
```
terceiro, line 1006: making this confidence score like this will indeed cause 1 extra query for each (test,environment) pair in the build. Given how large the tests table is, this will probably cause performance issues. What if you think of another design:

- call your `Confidence` class `FailureWithConfidence` instead
- have a single method in `Build`, called `failures_with_confidence`, that fetches all relevant `Test` objects in a single query, and then create the `FailureWithConfidence` objects interactively.

This might require you to change the design of `FailureWithConfidence` (prev. `Confidence`) slightly, like this:

```python
class FailureWithConfidence(object):
    def __init__(self, passes=0, count=0):
        self.passes = passes
        self.count = count

    def add_pass(self):
        self.passes += 1
        self.count += 1

    def add_failure(self):
        self.count += 1
```

(thread the above as pseudocode, it's untested)

The implementation of `Build.failures_with_confidence` could then look like the body of your view: iterate over the `Test` objects, collect the failure objects in a 2-dimensional array, creating new ones the first time it seems that (test name, environment) pair, and calling either `add_pass` or `add_failure`. You might create Failure objects for all tests, and the later discard the ones where `passes == count`.

squad/frontend/failures.py
```diff
@@ -0,0 +1,48 @@
...
+    visited = set()
+    ufailures = [visited.add(t.full_name) or t for t in tests if t.full_name not in visited]
+    ufailures.sort(key=lambda uf: uf.full_name)
+
+    rows = {}
+    for t in tests:
+        if t.environment.slug not in rows:
+            rows[t.environment.slug] = {}
+
+        rows[t.environment.slug][t.full_name] = t
```
terceiro, line 36: as noted above, this logic here could be moved into the model as `Build.failures_with_confidence`

squad/frontend/templates/squad/failures.jinja2
```diff
@@ -0,0 +1,80 @@
...
+                <td>{{ u.full_name }}</td>
+                {% for e in environments %}
+                {% if e.slug in rows and u.full_name in rows[e.slug]: %}
+                  <td class="{{ rows[e.slug][u.full_name].status }}">
+                    {{ rows[e.slug][u.full_name].status }}
+                    <span data-toggle="tooltip" title="{{ _('Click to display confidence info') }}">
+                      <button class="fa fa-info-circle popover-regressions-fixes"></button>
+                      <span title="{{ _('Confidence') }}" class="hidden">
+                        Score: {{ rows[e.slug][u.full_name].confidence.score  }}
+                      </span>
```
terceiro, line 44: it's a bit funny for you to go through all this trouble to then hide the confidence score in a tooltip. I know it's tricky given the volume of information in this table, but maybe it would make sense to use some color coding, e.g. a low confidence score having very light red background, and a high confidence score having a strong red background, while keeping?

you can save some space by not displaying the "fail 🛈" string in the cell body: since this page only lists failures, each cell is either empty meaning "not a failure", or non-empty, meaning a failure. You could even display the confidence score directly in the cell contents.

terceiro (changes requested):

squad/frontend/templates/squad/failures.jinja2
mrchapp: Very true. Some things to consider that are planned that might explain a few concerns.

Right now it's about failures and how confident we're about them being actual failures, but the plan is to use this same method to discover expected failures that are not failures anymore. The problem we have today is that the list of xfails is passed on from generation to generation (of stable branches, that is), so they are rarely reviewed. We'd like to find out when that's changed.

The other is that we want to reuse this confidence index (for failures, for passes) through an API. I see these changes are in the frontend/ directory, but am hoping that this code can be reused.

I'm not sure about the hidden score, though. Justin and I discussed this a bit and I wanted to have the cell information simplified to just the score (not even the label), and have any extra information provided via a tooltip, but the score definitely needs to be shown on the cell.

mrchapp:
Another thing related to performance: Charles mentioned that this operation could be DB intensive, and therefore it'd be better to store the confidence index in the DB when results are received (only once, only when data is entered). Justin says that in a subsequent conversation with Charles he mentioned this was not needed after all. (That's as far as my understanding goes.)

terceiro:
> Another thing related to performance: Charles mentioned that this operation could be DB intensive, and therefore it'd be better to store the confidence index in the DB when results are received (only once, only when data is entered). Justin says that in a subsequent conversation with Charles he mentioned this was not needed after all. (That's as far as my understanding goes.)

yes I think it can be calculated dynamically, provided that you optimize how the data is obtained in the db. If we can get the data in a single query then it's just fine. If we need one query for each test object then we are in trouble because those are many.

squad/frontend/failures.py
jscook2345: Thanks. Fixed.

squad/core/models.py
jscook2345: Fixed, it's now printed in the tooltip, with more use intended later on.

squad/frontend/templates/squad/failures.jinja2
jscook2345: Fixed.

I thought about using a color scheme, but @mrchapp was not thrilled with the idea. For now, just display the score and have the other information available via tooltip.

squad/frontend/failures.py
jscook2345: Question: Is the performance of the patch OK on its own? Or do you think we really do need to change things up?

I'm not sure I follow 100%, but I created a patch that maybe follows along with this. See below.

If the patch is not what you had in mind, do you really think we need to pull all the tests (including the history) at once instead of over multiple queries?

```
... in squad/core/models.py, class Build ...

+  566     __failures__ = None
+  567
+  568     @property
+  569     def failures(self):
+  570         if self.__failures__:
+  571             return self.__failures__
+  572
+  573         failing = self.tests.filter(
+  574             result=False,
+  575             has_known_issues=False,
+  576         )
+  577
+  578         for t in failing:
+  579             limit = self.project.build_confidence_count
+  580             history = Test.objects.filter(
+  581                 metadata=t.metadata,
+  582                 environment=t.environment,
+  583                 build_id__lt=self.id,
+  584             ).order_by("-build_id")[:limit]
+  585
+  586             // t.set_confidence(history)
+  587
+  588         self.__failures = failing
+  589         return self.__failures__
```

squad/core/models.py
jscook2345: Question: Is the performance of the patch OK on its own? Or do you think we really do need to change things up?

I'm not sure I follow 100%, but I created a patch that maybe follows along with this. See below.

If the patch is not what you had in mind, do you really think we need to pull all the tests (including the history) at once instead of over multiple queries?

```
... in squad/core/models.py, class Build ...

+  566     __failures__ = None
+  567
+  568     @property
+  569     def failures(self):
+  570         if self.__failures__:
+  571             return self.__failures__
+  572
+  573         failing = self.tests.filter(
+  574             result=False,
+  575             has_known_issues=False,
+  576         )
+  577
+  578         for t in failing:
+  579             limit = self.project.build_confidence_count
+  580             history = Test.objects.filter(
+  581                 metadata=t.metadata,
+  582                 environment=t.environment,
+  583                 build_id__lt=self.id,
+  584             ).order_by("-build_id")[:limit]
+  585
+  586             // t.set_confidence(history)
+  587
+  588         self.__failures = failing
+  589         return self.__failures__
```

terceiro: multiple queries can be ok when you know how many of them you need. in this case, we can easily have hundreds or thousands of them. Think a particularly broken AOSP build where 1% of its 1M tests fail. Then you have 1000 queries.

In this case you can solve it with 2 queries: 1 for "which tests fail on the current build", and 1 for "get the history of this set of tests for the latest N builds", and then you can do the confidence calculation on the python side. The 2nd query is similar to the one you have inside the loop, but with `metadata__in=[t.metadata for t in failing]` instead of `metadata=t.metadata`. Make sure to augment the first query with `prefetch_related("metadata")` as well to avoid have the same 100 queries but against the metadata table.

jscook2345:
@terceiro @chaws I think this is ready for a re-review now. I got the number of queries down based on feedback.

One open issue: How to handle the search part that was normally passed to the query set? Just filter the array of failures?

Would you prefer also I name it `failures_with_confidence` instead of `failures`? That would be fine as well.

Thanks again / ahead of time for your time here.

squad/core/models.py
```diff
@@ -563,6 +563,31 @@ def has_extra_metadata(self):
...
+    @property
+    def failures(self):
+        if self.__failures__ is None:
+            failures = []
+            limit = self.project.build_confidence_count
+
+            failing = self.tests.filter(result=False).exclude(has_known_issues=True).prefetch_related("metadata", "environment")
+
+            for f in failing:
+                failures.append(f)
```
terceiro, line 577: at the end, `failing` and `failures` are the same list as you append every item of the former to the later. you only need one of them.

squad/core/models.py
```diff
@@ -563,6 +563,31 @@ def has_extra_metadata(self):
...
+
+                history = Test.objects.filter(
+                    metadata=f.metadata,
+                    environment=f.environment,
+                    build_id__lt=self.id,
+                ).order_by("-build_id")[:limit]
```
terceiro, line 583: there is still one such query for each failing test. you want something like this, replacing the entire loop:

```python
# get all results for the currently failing tests in the last N builds, in a
# single query.
history = Test.objects.filter(
    metadata__in=[f.metadata for f in failing],
    environment__in=[f.environment for f in failing],
    build_id__lt=self.id,
).order_by("-build_id")[:limit]

# for each failing test, filter it's history locally, without going again to
# the database
for f in failing:
    f_history = [
        _t
        for _t in history
        if _t.metadata_id == f.metadata__in and _t.environment_id == f.environment_id
    ]
    f.set_confidence(self.project.build_confidence_threshold, f_history)
```

terceiro (changes requested):

squad/core/models.py
jscook2345:
```
).order_by("-build_id")[:limit]
```

If we set the limit here, we only get 20 tests in total. I need a history for each failing test for N builds. If I remove the limit, that would get all tests matching that.

Let me push something else to try here.

terceiro: yes, you are right. instead of limiting here you can do a query to get the ids the the last `limit` builds, and filter on those builds_ids instead.

terceiro (approved):

terceiro:
Hi, LGTM now at least from a performance standpoint. This will now get all the needed data with only 3 queries.

chaws:
Thanks for the patch @jscook2345 ! It looks good. Here are my comments below:

1. For some reason `boot/dragonboard-410c` in my dev env has a count greater than the other ones:
Screenshot from 2022-03-03 10-08-30

2. In line 581 you have 
```python
build_id__in=builds.values_list("id", flat=True)
```
And that is not yet a list of build ids, so Django uses it as a subquery. Whereas if you wrapped that around with `list()`, the query would be evaluated and you'd be dealing with an actual list of build ids. I did this locally and was able to speed up that query from ~200ms to ~20ms, around 10x!

3. If you don't plan to use test log, it's a nice idea using `defer("log")` in the queries to squeeze out a few extra ms.

This is nice-to-have: 
1. It would be nice if the user could click on the test name and be redirected to the test history page. Ex: https://qa-reports.linaro.org/lkft/linux-stable-rc-linux-5.10.y/tests/kselftest-kvm/kvm.hyperv_clock

jscook2345:
I pushed the performance fixes you suggested. Thanks!

> 1. https://qa-reports.linaro.org/lkft/linux-stable-rc-linux-5.10.y/tests/kselftest-kvm/kvm.hyperv_clock

Can we make this an issue for now and I can work on it in another PR that I can work on soon?

chaws:
Thanks, @jscook2345 ! np with the link, it's just a nice-to-have. I'll also send the patch that makes release of squad on Github action, and will let you roll out this one if you want?

mrchapp:
Hurrah!

PR #1032, opened 2022-03-11
core: models: return a zero confidence score if there is no history

jscook2345:
If this is the first time a test ran and it failed, there is no test
history to generate a confidence score from.

Return a score of zero for this case.


chaws:
Thanks, I'll merge and roll it out as soon as it finishes testing

PR #1033, opened 2022-03-18
frontend: failures: add pagination to the failures view

jscook2345:
Add a limit to the number of failures you can view per page (25) to
avoid performance problems when there are a large number of failures.


jscook2345:
@chaws and I discussed some changes to this where we will do some of the work in memory to get what we need. I will rework to that and force push over the top of this when I'm done.

jscook2345:
Please see the following replacement PR: https://github.com/Linaro/squad/pull/1035

PR #1034, opened 2022-03-22
Force fetch

chaws:
Authenticated users can now manually request a fech for a given job:

Screenshot from 2022-03-22 11-04-37

When clicking on "manual fetch", squad will enqueue the job again for fetching.


jscook2345 (approved):
LGTM! Nice work! You 🪨

PR #1035, opened 2022-03-22
Add pagination to failures view

jscook2345:
Use the FailuresWithConfidence class to control database access,
pagination, and searching for the failures view.

This class provides these features to improve performance when gathering
failing tests and their confidence score.

### DIFF
--- a/squad/core/failures.py
+++ b/squad/core/failures.py
@@ -1,0 +1,69 @@
+from django.core.paginator import Paginator
+from squad.core.models import Test
+
+
+class FailuresWithConfidence(object):
+    def __init__(self, project, build, per_page=25, page=1, search=''):
+        self.project = project
+        self.build = build
+        self.per_page = per_page
+        self.page = page
+        self.search = search
+
+    __paginator__ = None
+
+    @property
+    def paginator(self):
+        if self.__paginator__ is not None:
+            return self.__paginator__
+
+        names = self.build.tests.filter(
+            result=False,
+        ).exclude(
+            has_known_issues=True,
+        ).only(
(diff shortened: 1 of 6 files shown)


squad/frontend/templates/squad/failures.jinja2
```diff
@@ -23,26 +27,27 @@
             <th>{{ e.slug }}</th>
             {% endfor %}
         </thead>
-        {% if not ufailures %}
+        {% if fc.failures()|length == 0 %}
```
chaws, line 30: `fc.failures()` is the heaviest query and you are already calling it on line 21 in squad/frontend/failures.py

jscook2345: I changed this to `fc.failures().exists()` instead. Thanks for catching this.

chaws: Does `.exists()` re-use the same query? I think it'll still re-run the query + `LIMIT 1`

squad/core/failures.py
```diff
@@ -0,0 +1,70 @@
...
+    def number(self):
+        return self.page
+
+    def failures(self):
+        page = self.paginator.page(self.page)
+
+        failures = self.build.tests.filter(
+            result=False,
+            suite__slug__in=[t.suite__slug for t in page.object_list],
+            metadata__name__in=[t.metadata__name for t in page.object_list],
```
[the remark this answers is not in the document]
jscook2345: I implemented this as suggested.

chaws:
LGTM

PR #1036, opened 2022-03-24
frontend: metrics: prefetch metric's metadata

chaws:
The "All Metrics" page was way too slow, it was due to a missing prefetch on SuiteMetadata to retrieve the metric's name.


jscook2345:
Looks good. My lack of Django experience is showing ;)

jscook2345 (approved):

PR #1040, opened 2022-05-18
plugins: linux-log-parser: split boot log

chaws:
This will split linux log parser tests into 2 different suites: `linux-boot-log-parser` and `linux-test-log-parser`. The first suite will contain matches of kernel failures during boot whereas the second one will contain matches after boot.

* linux-boot-log-parser:
Screenshot from 2022-05-18 14-32-17

* linux-test-log-parser:
Screenshot from 2022-05-18 14-31-07


chaws:
Anders suggested "linux-boot-log-parser" -> "boot-log-parser" as suite name, the same for test.

chaws:
All checks are green

PR #1043, opened 2022-05-23
Set names for builds and test jobs

chaws:
Set TestJob.name for TuxSuite builds and tests


chaws:
All checks are green. The only failing test is due to a migration that will be fixed in future PRs.

PR #1046, opened 2022-05-27
core: failures: query past N tests instead of builds

chaws:
SQUAD was comparing the history with the past N builds, and because not necessarily a build has the test in it, this could cause variations on the "count" in the UI.

Now it'll query the past N number of tests x environment that ensure that the count number is always equals to N.


roxell:
LGTM, thank you @chaws

chaws:
Thanks for checking it out @roxell !

PR #1048, opened 2022-05-30
ci: tuxsuite: handle failed jobs

chaws:
The TuxSuite backend was ignoring all test jobs that had failed results. There can be two reasons why a test job fails: a) the build failed or b) the tests in the job actually failed.

SQUAD will now handle the second scenario as a regular job, except that the failure will be saved the testjob.


roxell:
LGTM.

PR #1065, opened 2022-08-04
Relicense under the GPLv3

terceiro:


terceiro:
@chaws PTAL

chaws:
@terceiro LGTM

PR #1068, opened 2022-08-16
frontend: fix reverse url for annotations UI

mwasilew:
When submitting annotation from UI it requires reverse API URL as a
parameter. Previous implementation only send Build object ID. It
resulted in a HTTP 400 response. This patch fixes the issue and provides
full reverse URL for the Build object API.

Signed-off-by: Milosz Wasilewski <milosz.wasilewski@foundries.io>


chaws:
Thanks @mwasilew ! Worked like a charm

PR #1069, opened 2022-08-16
core: frontend: add is_release flag to Build object

mwasilew:
This patch series adds is_release flag to Build object and
allows to display it in the build list and build details page.
Minor changes to REST API were also made.


chaws:
LGTM

Screenshot from 2022-08-16 14-47-25

It looks nice and the tag icon is pretty intuitive IMHO. 

Checks are red due to a missconfig I left it there. I'll fix it as soon as I merge this PR

Thanks @mwasilew !

PR #1070, opened 2022-08-18
frontend: build: add redirect URL to get API URL

terceiro:
When data is submitted via the API, there is currently no structured way
to obtain the internal build ID so that it can be used for API requests.
Add an URL at /<project>/<build>/build/<version>/api/ that will redirect
to the correct API endpoint for the given build, i.e. /api/builds/<id>/.


chaws:
LGTM

Just for future references, there's already a link in the build view ("api view of this build") pointing to the same api endpoint

PR #1071, opened 2022-08-23
api: add_test_run: signal invalid request on validation errors

terceiro:
Currentlty, an invalid date will cause an unhandled exception.


chaws:
LGTM!

PR #1080, opened 2022-10-10
plugins: linux_log_parser: compute contents to generate unique naming

chaws:
This will take in any match of a given log-parser result and generate a SHA256 sum to append to the name of the regex.

The intention is to have the possibility to compare same warnings/panics/etc against any other instance of the same fault in other builds.

Signed-off-by: Charles Oliveira <charles.oliveira@linaro.org>


roxell:
LGTM

PR #1082, opened 2022-10-21
Extend badges

chaws:
This allows generating badges with suite and environment as filter


chaws:
All checks are green, I'll merge this now and use the new badges to create a status table for TRS.

PR #1095, opened 2023-04-26
ci: lava: add websocket listener

chaws:
LAVA has recently enabled websocket in favor of ZMQ. Newer instances running on the cloud no longer have ZMQ working, so now SQUAD should first attempt connecting to websocket and fallback to ZMQ if the first doesn't work.


chaws:
Tests are green, there are warnings about code coverage and deprecation of some functions. I'm merging it now

PR #1096, opened 2023-04-28
plugins: Fix URL composing in github plugin

mwasilew:
Github will reject URLs with double / in the middle, like: https://api.github.com//some/api/path
Removing the redundant / fixes the problem. This patch makes sure there is no additional / even if the URL in the database ends with /.


chaws:
Tests are failing due to missing coverage, but it's been like that for a while now. I'll fix that once I have a bit of bandwidth for it.

PR #1097, opened 2023-06-08
plugins: Fix URL returned by github plugin to the UI

mwasilew:
Github API has changed since the plugin was created. The new URL fixes HTTP 404 status. It's still not ideal as it leads to the JSON object instead of the nice web UI.
https://docs.github.com/en/rest/commits/commits?apiVersion=2022-11-28#get-a-commit


mwasilew:
I added a test to cover `get_url` but it still can't go over 79% coverage.

chaws:
@mwasilew thanks for the fix, I'm yet to find the time to get the coverage above 80%, I'll work on a squad feature this week, will look into increase testing coverage then.

PR #1098, opened 2023-06-14
api: rest: add new endpoint 'statuses'

roxell:
This will be needed in order to to pull data for a specific board/boards and suites.
Example: How many LTP tests did we do for boards db410c and db845c during 2022?


chaws:
Done here: https://github.com/Linaro/squad/commit/f41e320a6ef6e4f44e4a240f73accb3734d8fd4d

PR #1099, opened 2023-06-22
Allauth only configured

mwasilew:


chaws:
LGTM, thanks for the patch @mwasilew !

PR #1100, opened 2023-06-27
frontend: fix bug with socialaccount discovery

mwasilew:
Templatetag for socialaccount discovery had a bug that crashed /login view. This patch fixes the bug and adds unit tests for this templatetag.


chaws:
Thanks for the fix :)

Also thank you for increasing the coverage as well

PR #1102, opened 2023-09-20
Change important_metadata to follow order from "Important metadata keys"

katieworton:
Change the important_metadata field so the items are displayed in the order from the "Important metadata keys" project setting rather than alphabetically sorting the list.


chaws:
Tests are passing locally but something popped up on github. I'm merging this now and will deal with GH failures later

PR #1108, opened 2023-10-30
api/rest.py: Add URLs to access attachments

katieworton:
WIP draft of support for accessing attachments. Adds an AttachmentViewSet so a hyperlink can be created.

Work in progress as I have not yet figured out how to combine the url and the filename to create a clickable link directly to the file.

### DIFF
--- a/squad/api/rest.py
+++ b/squad/api/rest.py
@@ -1,4 +1,5 @@
 import json
+import urllib.parse
 import yaml
 
 from django.db.models import Q, F, Value as V, CharField, Prefetch
@@ -1310,9 +1311,19 @@ class StatusViewSet(NestedViewSetMixin, ModelViewSet):
 
 class AttachmentSerializer(serializers.ModelSerializer):
 
+    download_url = serializers.SerializerMethodField()
+
     class Meta:
         model = Attachment
-        fields = ('filename', 'mimetype', 'length')
+        fields = ('download_url', 'filename', 'mimetype', 'length')
+
+    def get_download_url(self, attachment):
+        request = self.context.get('request')
+        if request is None:
+            return None
+        base_url = rest_reverse('testrun-detail', args=[attachment.test_run.pk], request=request)
+        filename_url_encoded = urllib.parse.quote(attachment.filename, safe='')
+        return f'{base_url}attachments/?filename={filename_url_encoded}'
 
 
 class TestRunSerializer(DynamicFieldsModelSerializer, serializers.HyperlinkedModelSerializer):
--- a/test/frontend/test_basics.py
+++ b/test/frontend/test_basics.py
@@ -249,6 +249,16 @@ class FrontendTest(TestCase):
         self.assertEqual('text/plain', response['Content-Type'])
         self.assertEqual(b'text file', response.content)
 
+    def test_attachment_download_url(self):
+        data = bytes('text file', 'utf-8')
+        filename = 'foo.txt'
+        attachment = self.test_run.attachments.create(filename=filename, length=len(data), mimetype="text/plain")
+        attachment.save_file(filename, data)
+        # NOTE: /api/testruns/%s/attachments?filename=foo.txt redirects to /api/testruns/%s/attachments/?filename=foo.txt
+        response = self.hit('/api/testruns/%s/attachments/?filename=foo.txt' % (self.test_run.id))
+        self.assertEqual('text/plain', response['Content-Type'])
+        self.assertEqual(b'text file', response.content)
+
     def test_log(self):
         response = self.hit('/mygroup/myproject/build/1.0/testrun/%s/suite/%s/test/%s/log' % (self.test_run.id, self.suite.slug, self.test.name))
         self.assertEqual('text/plain', response['Content-Type'])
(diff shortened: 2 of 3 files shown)


katieworton:
Noticed there were some issues with the previously pushed version of this code, so have changed approach. As a temporary solution, I have added an extra field which builds the url for accessing attachments. This successfully fulfils the requirement of adding a downloadable link, but doesn't look like the cleanest solution when compared to how other file downloads are handled in the code. I will explore more to see if there is a cleaner approach.

squad/api/rest.py
```diff
@@ -1310,9 +1310,17 @@ class StatusViewSet(NestedViewSetMixin, ModelViewSet):
...
     class Meta:
         model = Attachment
-        fields = ('filename', 'mimetype', 'length')
+        fields = ('download_url', 'filename', 'mimetype', 'length')
+
+    def get_download_url(self, attachment):
+        request = self.context.get('request')
+        if request is None:
+            return None
+        return request.build_absolute_uri(f"/{attachment.test_run.project}/build/{attachment.test_run.build.version}/attachments/testrun/{attachment.test_run.pk}/{attachment.filename}")
```
chaws, line 1323: I thought about that as well, but the problem with using file name as part of the url could be troublesome as it can contain funny characters.

The way I implemented, one can see all attachments for a testrun using `/api/testruns/123/`, then download individual attachments by constructing the URL `/api/testruns/123/attachments?filename=the-file-name`. In this case, the file name can contain whatever characters and they'll all be url-encoded.

Did you want links in each attachment so you wouldn't have to generate urls manually?

katieworton: Yes - I didn't actually realise it was possible to download the attachments this way (if you can point me towards the code that makes this possible that would be great).

I think it would be easiest to work with if there were links directly to the download URLs of the files, as it seems quite unintuitive/error-prone to have to manually construct the URL.

Regarding the issue with funny characters - wouldn't this be an issue regardless of whether we have a download URL stored in SQUAD? (aka if someone puts a space in a filename, it is going to be a problem when doing the "...?filename=..." method, too?)

chaws: I think you could do that with something like this:
```python
import urllib.parse

...

    def get_download_url(self, attachment):
        request = self.context.get('request')
        if request is None:
            return None
        base_url = rest_reverse('testrun-detail', pk=attachment.test_run.pk, request=request)
        filename_url_encoded = urllib.parse.quote(attachment.filename, safe='')
        return f'{base_url}/attachments?filename=filename_url_encded'
```

How does it look?

katieworton: Looks good to me - looks similar to another approach I put together, actually (but I was just missing the fact the attachments could be downloaded with "?filename=...")

This was my other approach if you were interested:
```
    def get_download_url(self, attachment):
        request = self.context.get('request')
        group = str(attachment.test_run.project).split('/')[0]
        project = str(attachment.test_run.project).split('/')[1]
        args = (group, 
                  project, 
                  attachment.test_run.build.version, 
                  attachment.test_run.pk, 
                  attachment.filename)
        relative_url = reverse('build_attachments', args=args)
        return request.build_absolute_uri(relative_url)
```

I will give your suggested method a try :)

I guess this part of the code can also be removed:
```
        request = self.context.get('request')
        if request is None:
            return None
```
I didn't realised at first that using `rest_reverse` would remove the need for the `self.context.get('request')`, but it seems obvious now :)

katieworton: actually just spotted that request is needed for rest_reverse, it's just that in the other places where it is used, request is passed in as an arg to the function - had just been wondering why the `request = self.context.get('request')` line was not used elsewhere in the file.

katieworton:
@chaws I've added a test case to check the download url works, updated the previous attachments test case to include the download url & I also generated and also performed and added the migrations command (not actually sure if that's come out correctly) - let me know if this looks ok or if any other changes would be required.

squad/api/rest.py
```diff
@@ -1310,9 +1311,19 @@ class StatusViewSet(NestedViewSetMixin, ModelViewSet):
 
 class AttachmentSerializer(serializers.ModelSerializer):
 
+    download_url = serializers.SerializerMethodField()
+
     class Meta:
         model = Attachment
-        fields = ('filename', 'mimetype', 'length')
+        fields = ('download_url', 'filename', 'mimetype', 'length')
```
roxell, line 1318: do we really need the 'filename' field here, that is in the download_url now or?

test/frontend/test_basics.py
```diff
@@ -249,6 +249,16 @@ def test_attachment(self):
         self.assertEqual('text/plain', response['Content-Type'])
         self.assertEqual(b'text file', response.content)
 
+    def test_attachment_download_url(self):
+        data = bytes('text file', 'utf-8')
+        filename = 'foo.txt'
+        attachment = self.test_run.attachments.create(filename=filename, length=len(data), mimetype="text/plain")
+        attachment.save_file(filename, data)
+        # NOTE: /api/testruns/%s/attachments?filename=foo.txt redirects to /api/testruns/%s/attachments/?filename=foo.txt
```
chaws, line 257: Good catch!

squad/api/rest.py
chaws: Well, it wouldn't hurt I think it's fine IMHO

chaws:
LGTM! I'll merge it and make 1.76.1 release so you can use it in your squad-client patch

chaws (approved):
LGTM

PR #1113, opened 2023-11-22
Add user preference for failures only

katieworton:
Add support for a user preference for the "failures only" field in the build results screen.
When the user is logged in, they can set a user preference in their profile for this "failures only" field. The default for this field is for the "failures only" tickbox to be checked.


katieworton:
@chaws Let me know what you think of this patch. Since our last chat, I've added unit tests and support for when the user isn't logged in.

chaws:
Sweet! LGTM, thanks Katie

chaws (approved):
LGTM

PR #1116, opened 2023-12-11
ci/backend/tuxsuite.py: Support for fetching test metadata file

katieworton:
Add support for fetching the TuxSuite test metadata file and adding its data to the SQUAD metadata file for the given test.

### DIFF
--- a/squad/ci/backend/tuxsuite.py
+++ b/squad/ci/backend/tuxsuite.py
@@ -287,6 +287,17 @@ class Backend(BaseBackend):
 
         return status, completed, metadata, tests, metrics, logs
 
+    def update_metadata_from_file(self, results, metadata):
+        if "download_url" in results:
+            download_url = results["download_url"]
+            try:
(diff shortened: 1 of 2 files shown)


katieworton:
Note: As far as I could tell, download_url does not appear to be available in the results for tests (only builds). For this reason, I perform a get request on the job_url to get the download_url from the job API data.

katieworton:
> > Note: As far as I could see, download_url does not appear to be available in the results for tests (only builds). For this reason, I perform a get request on the job_url to get the download_url from the job API data.
> 
> `download_url` isn't in current sample tests in SQUAD because that was recently added by Tuxsuite. You can do it like that:
> 
> ```
>         job_data = requests.get(job_url).json() # Remove this line
> 
>         # Add extra metadata from metadata file if it exists
>         self.update_metadata_from_file(job_data=results, metadata=metadata)
> ```

Perfect, thank you! I've now updated the patch to read `download_url` from the test results 😄

chaws (approved):
LGTM

PR #1124, opened 2024-01-16
Add required debian packages for local install

swvanbuuren:
This commit adds a recommendation to install additional Debian packages to be able to run SQUAD locally.

The installation of the required python packages using pip might fail, if these debian packages aren't installed. Unfortunately, the resulting error messages do not directly state that these debian packages are missing. Hence, a hint in the documentation is helpful to new users.


chaws (approved):
LGTM, I think I've been through that too. Thanks for adding this note

PR #1126, opened 2024-02-09
plugins: linux_log_parser: check-kernel-exception: add a new cut here

roxell:
The linux log parser doesn't catch all 'cut here' kernel-exceptions. Today it catches cut here looking like:
'------------[ cut here ]------------'
Add another multiline that can catch:
'--- cut here ---' too.


roxell:
https://tuxapi.tuxsuite.com/v1/groups/linaro/projects/lkft/tests/2bMlvywpGC6LN8eJfRQXQ4cTlEK/logs?format=html

PR #1129, opened 2024-03-07
doc: Update authorization token header in SQUAD docs

katieworton:
Change the authorization token from "Auth-Token:" to "Authorization: token" in SQUAD's documentation, to align with Django Rest Framework's Authorization token format.

While both formats are currently supported by SQUAD, Django only treats requests with the "Authorization: token" as authenticated, and will treat requests made with "Auth-Token" as unauthenticated requests. This can sometimes lead to incorrect behavior when using "Auth-Token". For this reason documentation should be updated to change "Authorization: token" to the preferred header for authentication.

### DIFF
--- a/doc/ci.rst
+++ b/doc/ci.rst
@@ -61,7 +61,7 @@ Example (with test job definition as POST parameter)::
 
     $ DEFINITION="$(cat /path/to/definition.txt)"
     $ curl \
-        --header "Auth-Token: $SQUAD_TOKEN" \
+        --header "Authorization: token $SQUAD_TOKEN" \
         --form backend=lava \
         --form definition="$DEFINITION" \
         https://squad.example.com/api/submitjob/my-group/my-project/x.y.z/my-ci-env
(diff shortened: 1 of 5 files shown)


katieworton:
@chaws From my exploration, I found the "Authentication: token" should always work correctly, while "Auth-Token" only works correctly in some cases - so I think it would be best if we update the documentation to point users towards "Authentication: token". SQUAD Client also use "Authentication: token" rather than "Auth-Token".

Let me know what you think of these proposed documentation updates :)

bhcopeland (approved):
LGTM

katieworton:
Thanks for this approval, Ben :+1: Just to note, I don't have the power to merge on this repo, so someone with this access will have to perform the merge.

chaws:
Just for future record: the REST API uses `Authorization` header, whereas the first API endpoints that were in squad used `Auth-Token`. Still both of these headers are still supported in non-REST API endpoints: https://github.com/Linaro/squad/commit/2f52508251e265d45f6d0fe0906751eca8874974

PR #1138, opened 2024-05-31
Do not look oneliners twice

chaws:
The plugin used to look oneliner regexes within multiliner matches, meaning more instances of tests being generated.

Now, the plugin does not look within multiliner matches.


chaws:
Antonio and Katie had looked over this PR and were ok with it. Merging it now.

